PR #1, opened 2025-01-13
feat: Initial draft of encoding module

dianacarvalho1:
It builds (with a bunch of warnings but no errors!)

TODO:
- General:
  - [ ] what should be public/private?
  - [ ] error handling?
- Models:
  - [x] Simplify Order
  - [x] Handle (un)wrap ETH
  - [x] How to support encoding only the 1 swap? (add_router_calldata)
  - [x] Handle permit2
- SwapEncoder:
  - [ ] Handle addresses gracefully 
- StrategyEncoder:
  - [x] Encode swap executors address and selector? 
- UX:
  - [ ] How do we expect the users to interact with the module?? 
    - [ ] Obvious swap with our router
    - [ ] Swap directly on the swap executor
    - [ ] Cowswap? 😬 
    - [ ] Allow unchecked?


src/encoding/approvals_manager.rs
```diff
@@ -0,0 +1,51 @@
...
+        &self,
+        token: Bytes,
+        spender_address: Address,
+        router_address: Address,
+    ) -> bool {
+        // should be something like
+        // let allowance = self
+        //     .client
+        //     .call(token, "allowance(address,address)(uint256)", (router_address, spender_address))
+        //     .await;
```
dianacarvalho1: it's not exactly like this, but we can figure it out when we implement

src/encoding/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod approvals_manager;
```
dianacarvalho1: later we need to decide what should be public or not

src/encoding/models.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// Address of the receiver.
+    pub receiver: Bytes,
+    /// List of swaps to fulfill the order.
+    pub swaps: Vec<Swap>,
+    /// Whether to include router calldata (true) or just swap data (false).
+    add_router_calldata: bool,
+
+    pub slippage: f64,
+    pub min_checked_amount: Option<BigUint>,
+}
```
dianacarvalho1: the ones that are not pub are currently not used anywhere 🤔 I'm thinking if we actually need them tbh

src/encoding/models.rs
```diff
@@ -0,0 +1,73 @@
+use alloy_primitives::Address;
+use lazy_static::lazy_static;
+use num_bigint::BigUint;
+use tycho_core::{dto::ProtocolComponent, Bytes};
+
+pub struct Solution {
+    pub orders: Vec<Order>,
+    // if not set, then the Propeller Router will be used
+    pub router_address: Option<Address>,
```
dianacarvalho1: I think here we should add another attribute to wrap or unwrap ETH..

src/encoding/models.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// Amount of the checked token.
+    checked_amount: BigUint,
+    /// Address of the sender.
+    sender: Bytes,
+    /// Address of the receiver.
+    pub receiver: Bytes,
+    /// List of swaps to fulfill the order.
+    pub swaps: Vec<Swap>,
+    /// Whether to include router calldata (true) or just swap data (false).
+    add_router_calldata: bool,
```
dianacarvalho1: I think this attribute should be moved up into the higher `Solution`

src/encoding/models.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub receiver: Address,
+    pub exact_out: bool,
+    pub router_address: Address,
+}
+
+lazy_static! {
+    pub static ref ROUTER_ADDRESS: Address = Address::from_slice(
+        &hex::decode("f847a638E44186F3287ee9F8cAF73FF4d4B80784")
+            .expect("Invalid string for router address"),
+    );
```
dianacarvalho1: aah this is unnecessary. will remove it

src/encoding/router_encoder.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    fn get_strategy(&self, order: &Order) -> &dyn StrategyEncoder {
+        if order.swaps.len() == 1 {
+            &SingleSwapStrategyEncoder {}
+        } else if order.swaps.iter().all(|s| s.split == 0.0) {
+            &SequentialExactInStrategyEncoder {}
+        } else {
+            &SlipSwapStrategyEncoder {}
+        }
```
dianacarvalho1: I collapsed having the dictionary and then this if conditions.. doing everything here now

src/encoding/strategy_encoder.rs
```diff
@@ -0,0 +1,130 @@
...
+        &self,
+        protocol_data: Vec<u8>,
+        protocol_system: String,
+        // Token indices, split, and token inclusion are only used for split swaps
+        token_in: u16,
+        token_out: u16,
+        split: u16, // not sure what should be the type of this :/
+    ) -> Vec<u8> {
+        let executor_address = get_swap_executor_address(&protocol_system);
+        let args = (executor_address, token_in, token_out, split, protocol_data);
```
dianacarvalho1, line 31: I think we should go further here actually 👀 
what do you think of having both the executor_address and the selector?
this way we could:
- select an executor with any address (a normal swap executor or a master swap executor)
- if the selector is empty -> use the standard swap one
- if the selector is set -> it's a special function in a master executor

src/encoding/strategy_encoder.rs
```diff
@@ -0,0 +1,130 @@
...
+                router_address,
+            };
+            let protocol_data = swap_encoder.encode_swap(swap.clone(), encoding_context)?;
+            let swap_data = self.encode_protocol_header(protocol_data, protocol_system, 0, 0, 0);
+            swaps.push(swap_data);
+        }
+        let (selector, action_type) = if order.exact_out {
+            (
+                "sequentialExactOut(uint256, uint256, bytes[])",
+                ActionType::SequentialExactOut,
```
dianacarvalho1: maybe we could collapse these into a hashmap? that if we know one we know the other one? 🤔 not sure if necessary

src/encoding/swap_encoder.rs
```diff
@@ -0,0 +1,86 @@
...
+            .unwrap();
+        let approval_needed = runtime.block_on(async {
+            token_approvals_manager
+                .approval_needed(
+                    swap.token_in.clone(),
+                    encoding_context.router_address,
+                    self.vault_address.unwrap(),
+                )
+                .await
+        });
```
dianacarvalho1: I feel icky about this but I'm not sure how we could do this differently 😕

src/encoding/swap_encoder.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+pub fn get_swap_executor_address(protocol_system: &str) -> Address {
+    match protocol_system {
+        "uniswap_v2" => Address::from_str("0x5C2F5a71f67c01775180ADc06909288B4C329308")
+            .expect("Invalid address"),
+        "vm:balancer_v2" => Address::from_str("0x543778987b293C7E8Cf0722BB2e935ba6f4068D4")
+            .expect("Invalid address"),
+        _ => panic!("Unknown protocol system: {}", protocol_system),
+    }
```
dianacarvalho1: not sure what is the best way to handle this.. should the addresses be attributes of the SwapEncoders? Should we read them from env? or config?

src/encoding/utils.rs
dianacarvalho1, line 1: all the methods in this module are courtesy of tycho-simulation 😂 (except the ple_encode)

src/encoding/router_encoder.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn new() -> Self {
+        let router_address = Address::from_str(
+            &env::var("ROUTER_ADDRESS").expect("Missing ROUTER_ADDRESS in environment"),
+        )
+        .expect("Invalid ROUTER_ADDRESS");
+        Self { router_address }
+    }
+
+    pub fn encode_router_calldata(&self, solution: Solution) -> Result<Vec<u8>, Error> {
+        let mut calldata_list: Vec<Vec<u8>> = Vec::new();
```
dianacarvalho1, line 25: I think we should have very thorough input validations before everything else 🫠

src/encoding/strategy_encoder.rs
tamaralipows: Why?

src/encoding/models.rs
tamaralipows: Hmmmm I can see that the sender is not necessary since we just use `msg.sender` in our router.
About the checked token and amount - why aren't we using the `checked_amount`? Didn't fully understand from the code below (I can check more later)

src/encoding/models.rs
tamaralipows: Yes - exactly what I was thinking

src/encoding/models.rs
tamaralipows: Wouldn't it also make more sense to put the `swaps` at the higher Solution? They are not part of the order, after all

src/encoding/models.rs
tamaralipows: Ahhhh. I just now see below a min_checked_amount.
I think that should be removed (keep in mind it would be the max amount for ExactOut)
And then checked_amount is public

src/encoding/router_encoder.rs
tamaralipows: i like it, thank you

src/encoding/strategy_encoder.rs
dianacarvalho1: fLeXibiLitY 🤸🏼

src/encoding/models.rs
dianacarvalho1: why not??? aren't the specific to a certain order? 🤔

tamaralipows: they are part of a solution, not the order? the order itself doesn't tell you which pools to swap through

src/encoding/mod.rs
kayibal: Ideally as little as possible should be public.

src/encoding/approvals_manager.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+use alloy_primitives::{Address, U256};
+use dotenv::dotenv;
+use tycho_core::Bytes;
+
+pub struct TokenApprovalsManager {
+    client: Arc<RootProvider<BoxTransport>>,
+}
+
+impl TokenApprovalsManager {
```
kayibal: Should this be a trait instead?

src/encoding/models.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// True if the order is an exact output order.
+    pub exact_out: bool,
+    /// The token being sold (exact in) or bought (exact out).
+    pub given_token: Bytes,
+    /// Amount of the given token.
+    pub given_amount: BigUint,
+    /// The token being bought (exact in) or sold (exact out).
+    checked_token: Bytes,
+    /// Amount of the checked token.
+    checked_amount: BigUint,
```
kayibal: Why are these not pub?

src/encoding/models.rs
```diff
@@ -0,0 +1,68 @@
+use alloy_primitives::Address;
+use num_bigint::BigUint;
+use tycho_core::{dto::ProtocolComponent, Bytes};
+
+pub struct Solution {
+    pub orders: Vec<Order>,
+    // if not set, then the Propeller Router will be used
+    pub router_address: Option<Address>,
```
kayibal: This seems a bit out of place here to be honest... Like solution and address seems like they should be on different levels of abstraction no? One holds high level information - the set of swaps I orders I want to execute and then a very lowlevel thing the Address. I would also recommend you to not use alloy throughout public interfaces since this will make going cross chain a lot harder.

src/encoding/models.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub token_in: Bytes,
+    /// Token being output from the pool.
+    pub token_out: Bytes,
+    /// Fraction of the amount to be swapped in this operation.
+    pub split: f64,
+}
+
+// maybe this struct is useful - keeping it here for now (maybe we could collapse this with another
+// struct)
+pub struct EncodingContext {
```
kayibal, line 74: The address attribute that seemed out of place in Solution struct above seems correct here. I would still prefer if it was Bytes.

src/encoding/permit2.rs
```diff
@@ -0,0 +1,44 @@
+use alloy_primitives::{Address, U256};
+use num_bigint::BigUint;
+use std::str::FromStr;
+use tycho_core::Bytes;
+
+pub struct PermitRequest {
```
kayibal: this seems to be an EVM specific thing so I guess alloy types are ok here. But we probably need some interface around it in this case (ApprovalsManager maybe?).

src/encoding/router_encoder.rs
```diff
@@ -0,0 +1,74 @@
...
+struct RouterEncoder {
+    router_address: Address,
+}
+
+impl RouterEncoder {
+    pub fn new() -> Self {
+        let router_address = Address::from_str(
+            &env::var("ROUTER_ADDRESS").expect("Missing ROUTER_ADDRESS in environment"),
+        )
+        .expect("Invalid ROUTER_ADDRESS");
```
kayibal: I guess this is temporary?

src/encoding/router_encoder.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    fn get_strategy(&self, order: &Order) -> &dyn StrategyEncoder {
+        if order.swaps.len() == 1 {
+            &SingleSwapStrategyEncoder {}
+        } else if order.swaps.iter().all(|s| s.split == 0.0) {
+            &SequentialExactInStrategyEncoder {}
+        } else {
+            &SlipSwapStrategyEncoder {}
+        }
+    }
```
kayibal: Would be nice if the user could extend this to their liking - e.g. via a default trait implementation

src/encoding/utils.rs
```diff
@@ -0,0 +1,56 @@
...
+use num_bigint::BigUint;
+use tycho_core::Bytes;
+
+/// Safely converts a `Bytes` object to an `Address` object.
+///
+/// Checks the length of the `Bytes` before attempting to convert, and returns a `SimulationError`
+/// if not 20 bytes long.
+pub fn bytes_to_address(address: &Bytes) -> Result<Address, Error> {
+    if address.len() == 20 {
+        Ok(Address::from_slice(address))
```
kayibal, line 13: Maybe there is a `try_from_slice` method?

src/encoding/swap_encoder.rs
tamaralipows: Hmmmmm this is actually a good question to ask - where will we be storing the executor addresses in general? In some doc? Perhaps we should keep them in AWS secrets and pull them from there? Or some other online DB? I am trying to avoid human error here - what if we forget to switch one executor in some place (either encoder or router).

src/encoding/swap_encoder.rs
tamaralipows: The fact that we are using `block_on`? Why are we keeping the tokens manager async only to block on later? Is it because of the RPC call?

src/encoding/strategy_encoder.rs
tamaralipows: Sorry, I meant a more specific question:

> if the selector is set -> it's a special function in a master executor

What is the master executor? Why are we planning to have this? When would the selector ever be different than the "swap"? Do we expect one executor to have two different entrypoints? I didn't imagine any of that

src/encoding/strategy_encoder.rs
tamaralipows: This is the `SequentialExactInStrategyEncoder` so it will always be exact in :D no need for this

src/encoding/models.rs
dianacarvalho1: why? I'm really confused here. I always assumed that the swaps were to fulfil a certain order

src/encoding/strategy_encoder.rs
dianacarvalho1: no no. but remember the discussions of having a master executor? one contract with multiple methods like: swap_uniswap_v2, swap_uniswap_v3, etc etc. This would leave the design open for this possibility if we want in the future (it should be more gas efficient)

src/encoding/swap_encoder.rs
dianacarvalho1: yes.. the RPC call will be async 😕 I didn't want to convert everything to be async because of this..

src/encoding/swap_encoder.rs
dianacarvalho1: huummm.. two things:
1. We would have this list in our encoding side and then approved executors on the router side. If then in the router side, we use an address that is not approved, then fail fail fail.
2. Remember that this will have to be run by users locally.. they won't have access to our AWS 😬 maybe it has to be just a config file?? doing env vars it's too overwhelming for this (we expect to have many many)

src/encoding/models.rs
dianacarvalho1: so far they haven't been used anywhere 😬 it's a wip

src/encoding/models.rs
dianacarvalho1: > would also recommend you to not use alloy throughout public interfaces since this will make going cross chain a lot harder.

true true true

I'm going to refactor these structs then 🤔

src/encoding/router_encoder.rs
dianacarvalho1: why?

src/encoding/permit2.rs
dianacarvalho1: huuuuummmmm 🤔

src/encoding/models.rs
dianacarvalho1: I'm confused here actually.. 🤔 
shouldn't we support both? (talking for exact in)
- we should have an expected buy_amount that then we apply slippage in
- and a absolute min buy amount that the user sets

src/encoding/models.rs
dianacarvalho1: but this struct is only used deeper within our methods as a helper struct.. it isn't exposed to the user 😕

src/encoding/utils.rs
dianacarvalho1: no :/

src/encoding/permit2.rs
dianacarvalho1: can you please see my latest commit [here](https://github.com/propeller-heads/tycho-execution/pull/1/commits/d3a3522a66cd10efd5bba2920040e0cf79440790) please?

src/encoding/approvals_manager.rs
dianacarvalho1: I made a trait for this in the latest commit [here](https://github.com/propeller-heads/tycho-execution/pull/1/commits/d3a3522a66cd10efd5bba2920040e0cf79440790)

src/encoding/models.rs
dianacarvalho1: Refactored them [here](https://github.com/propeller-heads/tycho-execution/pull/1/commits/9f6150b093df7a3e7fe94eed22ffbada180a4fb3)

src/encoding/models.rs
kayibal: It's public here? What do you mean it is not exposed to the user?

src/encoding/strategy_encoder.rs
kayibal: I like having the selector flexible

src/encoding/swap_encoder.rs
kayibal: I think it should be a config file. And I think you should use the builder pattern here instead of panicking (if this is only PoC code it's all good)

src/encoding/swap_encoder.rs
kayibal: I think you will have to make this async

src/encoding/router_encoder.rs
kayibal: I don't think we should read this from env at struct instantiation. It may happen that a downstream app starts up fine, then finds a swap and then panicks only here once it tries to encode the swap. That is a terrible user experience.

src/encoding/approvals/interface.rs
```diff
@@ -0,0 +1,13 @@
...
+use tycho_core::Bytes;
+
+pub struct Approval {
+    pub spender: Bytes,
+    pub owner: Bytes,
+    pub token: Bytes,
+    pub amount: BigUint,
+}
+
+pub trait ApprovalsManager {
```
kayibal: As you mentioned in the commit message this is now an ambiguous trait. I think this trait should focus on user approvals. Idk if we need a trait for the router to manage approvals to pools afaik this will always be simple unlimited approvals and nothing fancy with permit2 or similar.

src/encoding/swap_encoder.rs
```diff
@@ -0,0 +1,85 @@
...
+struct UniswapV2SwapEncoder {}
+
+impl SwapEncoder for UniswapV2SwapEncoder {
+    fn encode_swap(&self, swap: Swap, encoding_context: EncodingContext) -> Result<Vec<u8>, Error> {
+        todo!()
+    }
+}
+
+struct BalancerV2SwapEncoder {
+    vault_address: Bytes,
```
kayibal: It's not necessary for the implementation to use Bytes internally. You could have kept Address here I would actually prefer it.

src/encoding/models.rs
dianacarvalho1: it's created inside the `StrategyEncoder` only. to the user only the Solution, Order and Swap will be exposed

src/encoding/swap_encoder.rs
tamaralipows: ok agreed! (forgot about users running this locally.. brainfart last night sorry)

src/encoding/strategy_encoder.rs
```diff
@@ -103,37 +107,47 @@ impl StrategyEncoder for SequentialExactInStrategyEncoder {
...
+            "sequentialExactOut(uint256, uint256, bytes[])"
+        } else {
+            "sequentialExactIn(uint256, uint256, bytes[])"
         }
     }
 }
 
 pub struct SlipSwapStrategyEncoder {}
 
 impl StrategyEncoder for SlipSwapStrategyEncoder {
```
tamaralipows: SlipSwap hehehe

src/encoding/strategy_encoder.rs
```diff
@@ -103,37 +107,47 @@ impl StrategyEncoder for SequentialExactInStrategyEncoder {
...
+    fn encode_strategy(&self, order: Order) -> Result<Vec<u8>, Error> {
         todo!()
     }
+    fn action_type(&self, _exact_out: bool) -> ActionType {
+        ActionType::SplitIn
+    }
+
+    fn selector(&self, _exact_out: bool) -> &str {
+        "splitExactIn(uint256, address, uint256, bytes[])"
+    }
```
tamaralipows, line 154: Note for future us - should we panic if someone wants split exact out?

src/encoding/models.rs
tamaralipows: correct but then the `expected_buy_amount` should not be called `checked_amount` because this is confusing and historically has been the amount that we **check**.

I say we have `buy_amount` and `check_amount`
One is for applying slippage, one is the user-specified amount to check in the contract

src/encoding/strategy_encoder.rs
dianacarvalho1: maybe at the validations stage?

src/encoding/strategy_encoder.rs
tamaralipows: hmm ok ok, I'm open to keeping it flexible then! shouldn't be too much work anyway

src/encoding/strategy_encoder.rs
tamaralipows: ok!

src/encoding/models.rs
tamaralipows: agreed on slack to simplify these classes to `Solution` and `vec<Solution>`

src/encoding/swap_encoder.rs
dianacarvalho1: @kayibal how would you use the builder pattern here? 🤔 if it's a config file, we would just load the file into a hashmap and then retrieve from it no? I feel like I'm missing something

dianacarvalho1: @kayibal played a bit with it and came up with [this](https://github.com/propeller-heads/tycho-execution/pull/1/commits/44df8e48c3e47be269540e9ab61b71921f96a26c). Let me know if this was more or less what you had in mind!

src/encoding/swap_encoder/builder.rs
```diff
@@ -0,0 +1,39 @@
...
+        self
+    }
+
+    pub fn build(self) -> Result<Box<dyn SwapEncoder>, String> {
+        let executor_address = self.executor_address.ok_or_else(|| {
+            format!(
+                "Executor address must be provided for protocol: {}",
+                self.protocol_system
+            )
+        })?;
```
louise-poole: If no executor address was provided you could check the config file, no? Otherwise it shouldn't be made optional here if you're going to just error if it's missing.

src/encoding/swap_encoder/swap_encoder.rs
```diff
@@ -8,23 +8,37 @@ use std::str::FromStr;
...
 
-struct UniswapV2SwapEncoder {}
+pub struct UniswapV2SwapEncoder {
+    executor_address: Address,
+}
 
+impl UniswapV2SwapEncoder {
+    pub fn new(executor_address: Address) -> Self {
+        Self { executor_address }
+    }
```
louise-poole: We expect all SwapEncoders to implement this `new(executor_address)` so should have this in the trait definition

src/encoding/swap_encoder/builder.rs
dianacarvalho1: but I already loaded it from the config file to begin with 😕

louise-poole: What I'm saying is if you want to load it from the config file outside of the builder, then you need to make the executor address non-optional inside the builder. Doesn't make sense to have it optional but panic when it's missing... It needs to be passed like the protocol system is passed when you create the builder.

dianacarvalho1: but then I don't need the builder pattern anymore, do I? 😂

dianacarvalho1: only for the choice of SwapEncoder, ok ok. I accept

tamaralipows (approved):
A good starting point! Any remaining open questions/concerns will be addressed in future PRs (it's easier than attempting to tackle all of them now with no implementation knowledge).

PR #3, opened 2025-01-16
feat: Add foundry environment and CI

dianacarvalho1:
Got all of this with a 
````
forge init --no-git
````


tamaralipows (approved):

tamaralipows:
I'm a bit surprised to be honest :o that so many files are needed. But I guess the other projects are like this too?

dianacarvalho1:
> I'm a bit surprised to be honest :o that so many files are needed. But I guess the other projects are like this too?

yes, defibot-contracts was like that as well

PR #8, opened 2025-01-17
feat: Add EncodingError

dianacarvalho1:
Change method signatures to expect it and raise it where it makes sense

### DIFF
--- a/src/encoding/errors.rs
+++ b/src/encoding/errors.rs
@@ -9,6 +9,9 @@ use thiserror::Error;
 /// Variants:
 /// - `InvalidInput`: Indicates that the encoding has failed due to bad input parameters.
 /// - `FatalError`: There is problem with the application setup.
+/// - `RecoverableError`: Indicates that the encoding has failed with a recoverable error. Retrying
+///   at a later time may succeed. It may have failed due to a temporary issue, such as a network
+///   problem.
 #[derive(Error, Debug)]
 #[allow(dead_code)]
 pub enum EncodingError {
@@ -16,6 +19,8 @@ pub enum EncodingError {
     InvalidInput(String),
     #[error("Fatal error: {0}")]
     FatalError(String),
+    #[error("Recoverable error: {0}")]
+    RecoverableError(String),
 }
 
 impl From<io::Error> for EncodingError {
--- a/src/encoding/errors.rs
+++ b/src/encoding/errors.rs
@@ -1,0 +1,31 @@
+use std::io;
+
+use thiserror::Error;
+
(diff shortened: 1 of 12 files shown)


src/encoding/errors.rs
```diff
@@ -0,0 +1,30 @@
...
+
+/// Represents the outer-level, user-facing errors of the tycho-execution encoding package.
+///
+/// `EncodingError` encompasses all possible errors that can occur in the package,
+/// wrapping lower-level errors in a user-friendly way for easier handling and display.
+/// Variants:
+/// - `InvalidInput`: Indicates that the encoding has failed due to bad input parameters.
+/// - `FatalError`: There is problem with the application setup.
+#[derive(Error, Debug)]
+pub enum EncodingError {
```
dianacarvalho1, line 17: @tamaralipows do you think we need something more complex? 🤔 I'm thinking this is ok at this point tbh

tamaralipows: I'm thinking about our previous project, where we had Fatal, Recoverable with different input, and Recoverable if you try again. I think there we have the first two, but we are missing the third case - which I can see happening due to for example network issues with the RPC

tamaralipows (commented):
Just one comment replying your question, the rest looks good!

src/encoding/errors.rs
dianacarvalho1: like a temporary fatal? 😂

tamaralipows: temporary **recoverable**

tamaralipows: you can call it retry error or something (i dont remember how we called in it simulation - have a look there? the user is likely already familiar with it by this point)

dianacarvalho1: it's a recoverable.. but the thing is: this is like something you call once and that's it.. you don't run it forever and ever no? I think if it's an RPC failure we could just say that in the fatal message

tamaralipows: i don't think it's safe to make assumptions how the client will be using this - it may be just as automated as simulation. i dont think its a fatal error - fatal describes moreso bugs in our code that the user cant overcome...

tamaralipows: feels weird that rpc errors in simulation are recoverale but not in execution... when the user may be using them chained anyway

dianacarvalho1: huuummmm ok. I will add a recoverable error then!

tamaralipows (approved):
Thank you! I like it!

PR #9, opened 2025-01-17
feat: Simplify StrategyEncoders and RouterEncoder

dianacarvalho1:
- No more SingleSwapStrategyEncoder or SequentialStrategyEncoder 
- No more batch routing (so no need for batch permits as well). If multiple solutions are found, return multiple Transactions


tamaralipows (approved):
tenk yu

PR #11, opened 2025-01-17
feat: TychoRouter skeleton

tamaralipows:


foundry/src/TychoRouter.sol
```diff
@@ -1,6 +1,79 @@
...
+    }
+
+    /**
+     * @dev Executes a swap graph with internal splits token amount
+     *  splits, checking that the user gets more than minUserAmount of buyToken.
+     *
+     *  Assumes the swaps in swaps_ already contain any required token
+     *  addresses.
+     */
+    function splitExactIn(
```
dianacarvalho1: I think we could call it only swap since it's the only one 🤷🏼‍♀️

tamaralipows:
I've made a bit of a git mess so I'm closing this and creating a new branch!

PR #12, opened 2025-01-20
feat: Implement ProtocolApprovalsManager

dianacarvalho1:
I decided to make ProtocolApprovalsManager public method sync not to propagate asynceness early on onto the rest of the code. If later we decide that it is better to have this async, it should be easier to make that change than the opposite one.

### DIFF
--- a/src/encoding/evm/approvals/protocol_approvals_manager.rs
+++ b/src/encoding/evm/approvals/protocol_approvals_manager.rs
@@ -48,9 +48,10 @@ impl ProtocolApprovalsManager {
 
                 Ok(allowance.is_zero())
             }
-            Err(err) => {
-                Err(EncodingError::RecoverableError(format!("Call failed with error: {:?}", err)))
-            }
+            Err(err) => Err(EncodingError::RecoverableError(format!(
+                "Allowance call failed with error: {:?}",
+                err
+            ))),
         }
     }
 }
--- a/src/encoding/evm/approvals/protocol_approvals_manager.rs
+++ b/src/encoding/evm/approvals/protocol_approvals_manager.rs
@@ -1,49 +1,98 @@
 use std::{env, sync::Arc};
 
 use alloy::{
-    providers::{ProviderBuilder, RootProvider},
+    providers::{Provider, ProviderBuilder, RootProvider},
+    rpc::types::{TransactionInput, TransactionRequest},
     transports::BoxTransport,
 };
-use alloy_primitives::Address;
+use alloy_primitives::{Address, Bytes, TxKind, U256};
+use alloy_sol_types::SolValue;
 use dotenv::dotenv;
+use tokio::runtime::Runtime;
+
+use crate::encoding::{errors::EncodingError, evm::utils::encode_input};
 
 #[allow(dead_code)]
 pub struct ProtocolApprovalsManager {
     client: Arc<RootProvider<BoxTransport>>,
+    runtime: Runtime,
 }
 impl ProtocolApprovalsManager {
     pub fn new() -> Self {
-        Self { client: get_client() }
+        let runtime = Runtime::new().expect("Failed to create runtime");
+        let client = runtime.block_on(get_client());
+        Self { client, runtime }
     }
-    pub async fn approval_needed(
+    pub fn approval_needed(
         &self,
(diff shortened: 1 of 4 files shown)


src/encoding/evm/approvals/protocol_approvals_manager.rs
```diff
@@ -1,49 +1,98 @@
...
 
 #[allow(dead_code)]
 pub struct ProtocolApprovalsManager {
     client: Arc<RootProvider<BoxTransport>>,
+    runtime: Runtime,
 }
 impl ProtocolApprovalsManager {
     pub fn new() -> Self {
-        Self { client: get_client() }
+        let runtime = Runtime::new().expect("Failed to create runtime");
```
tamaralipows, line 22: No using expect in production code I guess? We should make this a `FatalError`

src/encoding/evm/approvals/protocol_approvals_manager.rs
```diff
@@ -1,49 +1,98 @@
...
+        match output {
+            Ok(response) => {
+                let allowance: U256 = U256::abi_decode(&response, true).map_err(|_| {
+                    EncodingError::FatalError("Failed to decode response for allowance".to_string())
+                })?;
+
+                Ok(allowance.is_zero())
+            }
+            Err(err) => {
+                Err(EncodingError::RecoverableError(format!("Call failed with error: {:?}", err)))
```
tamaralipows: Maybe specify that it was an RPC call to get approvals?

src/encoding/evm/approvals/protocol_approvals_manager.rs
dianacarvalho1: why not? I thought it was the unwrap that we shouldn't use in production code

tamaralipows: Please see the reasoning here - https://github.com/propeller-heads/tycho-simulation/pull/12#discussion_r1808498053

This has been our norm throughout indexer and simulation so it would be nice to keep consistency with execution too I think?

The user in this case may choose to ignore any protocol with dynamic approvals if our manager is broken...
Also, what is the point of having `FatalError` if we use expects?

dianacarvalho1: huuumm I dunno if I agree 😕 (also I've been using expect in simulation in certain situations still 😬 )
In this case, to raise a FatalError I would have to change the signature of the `new()` method to be a Result. I feel this is an overkill 😕 I think that this particular example should be an expect. Creating a runtime is not something the user can influence or has any power over it. If it fails to create the run time, then yeah mega disasters and it should crash and burn

dianacarvalho1: we are actually doing the exact same thing [here](https://github.com/propeller-heads/tycho-simulation/blame/1168a204bd76e02f1670f47914da531f46c49f33/src/evm/simulation.rs#L194)

tamaralipows: > Creating a runtime is not something the user can influence or has any power over it.

Most of our fatal errors are also on our side -> user can't control it...

I think we need to then have a clear boundary of when to use FatalError and when to use expect because right now I don't see the boundary at all

tamaralipows: The errors that I mentioned in the other thread were also big disasters that the user cannot control:

- malformed ERC20.abi file
- malformed directory structure

dianacarvalho1: I would be ok with those being expects as well tbh. At least in my understanding of rust error handling (which is not a lot 🫠 ) I feel that we can use `expect` when it is something like hardcoded almost, for example, parsing a constant and something that should ever ever ever happen. No point in giving the user the chance to handle it gracefully if it is just a crazy disaster. 
I think in the simulation it was different because those errors were specific of vm pools and the user could choose to ignore those pools but tbh if that was raised, the code would be malfunctioning completely, it wouldn't be like: ups connection is temporarily down or the input is wrong. it would be really a: there is a bug.

tamaralipows (approved):

PR #13, opened 2025-01-20
feat: TychoRouter Skeleton

tamaralipows:
It was trickier than I thought to figure out from scratch how the submodules, imports, remapping, and Slither works. I think I finally have a pretty good starting point, though!

For this, two submodules were added:
- openzeppelin for access control
- permit2 for approvals

Note: decided not to include all helper methods in this skeleton - just main swap method. The helper methods are properly documented in the jira task so this shouldn't be a big issue.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "foundry/lib/openzeppelin-contracts"]
+	path = foundry/lib/openzeppelin-contracts
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "foundry/lib/openzeppelin-contracts"]
```
dianacarvalho1, line 1: ooohhhhh this looks tough to have figured out 😬

dianacarvalho1 (approved):
Veery niceeeee! 🙏🏼 Thank you so much @tamaralipows !

PR #15, opened 2025-01-21
feat: Implement Permit2 encoding

dianacarvalho1:
I used these tutorials as guidelines:
- [Permit2](https://blog.uniswap.org/permit2-integration-guide)
- [Sign permit](https://github.com/alloy-rs/examples/blob/54b1c57a90cc055d541c02feee7f5759fe51edf6/examples/wallets/examples/sign_permit_hash.rs) using alloy

Done:
- Retrieve allowance data from an RPC
- Encode approvals:
  - Create PermitSingle (and PermitDetails)
  - Use alloy eip712 to get signing hash on typed data
  - Sign data
- Add chrono and alloy eip712 and signer-local to Cargo.toml


src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,42 +1,230 @@
...
-        // get allowance data (if it exists) and the nonce
-        //  returns permitAmount, expiration, nonce
-        todo!()
+        user: &Bytes,
+        router_address: &Bytes,
+        token: &Bytes,
+    ) -> Result<Allowance, EncodingError> {
+        let args =
+            (bytes_to_address(user)?, bytes_to_address(token)?, bytes_to_address(router_address)?);
+        let data = encode_input("allowance(address,address,address)", args.abi_encode());
```
dianacarvalho1, line 84: this method calls `allowance` on the [permit2 contract](https://etherscan.io/address/0x000000000022D473030F116dDEE9F6B43aC78BA3#readContract) where the output is (amount, expiration, nonce). In all my attempts (with different values) I ever only received (0,0,0). I'm not sure in which cases this would not be zero 🤔 I'm thinking that it would be non zero if there was already a valid permit (within the deadline for this token, owner and user)? Like that we had called already in our router the `permit2.permit(msg.sender, permitSingle, signature);` 
Do you think we should address the non zero case? what would even happen if we made a new signature for the same token, owner and user? 🤔

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,42 +1,240 @@
...
-        Self {
-            address: Bytes::from_str("0x000000000022D473030F116dDEE9F6B43aC78BA3")
-                .expect("Permit2 address not valid"),
-        }
+    pub fn new(signer: PrivateKeySigner, chain_id: ChainId) -> Result<Self, EncodingError> {
+        let runtime = Runtime::new()
+            .map_err(|_| EncodingError::FatalError("Failed to create runtime".to_string()))?;
+        let client = runtime.block_on(get_client());
+        Ok(Self {
+            address: Address::from_str("0x000000000022D473030F116dDEE9F6B43aC78BA3")
```
tamaralipows, line 67: Since this is constant should we look into the hex! macro that Alan mentioned here? https://propellerswap.slack.com/archives/GB59UTQ2K/p1737394960264169?thread_ts=1737392914.620059&cid=GB59UTQ2K

src/encoding/evm/approvals/permit2.rs
tamaralipows: I think this is exactly what it means - could you test it where permit2.permit is already called? It's probably valid to overwrite the current allowance if it's called again - i.e. perhaps the expiration may be set to a later time before the current allowance has expired, or the amount may be increased...

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,42 +1,240 @@
...
+        let current_time = Utc::now()
+            .naive_utc()
+            .and_utc()
+            .timestamp() as u64;
+
+        let mut encoded_approvals = Vec::new();
+
+        for approval in approvals {
+            let (_, _, nonce) =
+                self.get_allowance_data(&approval.owner, &approval.spender, &approval.token)?;
```
tamaralipows: This fn name is a bit vague - I'd say it should be called `get_existing_allowance` or something like this to be clear this is the previous allowance not the current one we are trying to set?

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,42 +1,240 @@
...
+            let expiration = current_time + PERMIT_EXPIRATION;
+            let sig_deadline = current_time + PERMIT_SIG_EXPIRATION;
+
+            let details = PermitDetails {
+                token: bytes_to_address(&approval.token)?,
+                amount: U160::from(biguint_to_u256(&approval.amount)),
+                expiration: U48::from(expiration),
+                nonce,
```
tamaralipows, line 132: I just now understand that this `nonce` doesn't represent the user's own wallet's `nonce` (like we thought in a previous conversation where I was concerned about that making no sense), but rather an internal `nonce` of the permit2 contract that increases every time a signature is used.

> The nonces used to protect against replay attacks of signatures are similar to standard incrementing nonces. However, we pack nonces with an allowed amount, and an allowed duration. This means that nonces are incremented per owner, per token, and per spender. Which further implies that you could sign two different permits at the same time with the same nonces and they won’t cancel each other out so long as the token or spender differ.

Source https://docs.uniswap.org/contracts/permit2/reference/allowance-transfer#nonce-schema

So it looks like it does make sense to expect a non-zero case - as this nonce will increase for the same token/user pair.

tamaralipows (commented):
I think it looks pretty good - can you just make sure my replies about your doubts make sense to you?

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: meh.. I think Alan mentioned hex! because there we were using hex::decode in the first place

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: > could you test it where permit2.permit is already called?

not really 😕 I would have to actually call the contract for real on a fork no? not sure how hard that would be in a test.

dianacarvalho1: actually this is a great idea. I'm trying to do it but the tx fails 🫠 however I do think that it is better to loose some time to figure this out now (we would have to figure this out at some point anyway)

src/encoding/evm/approvals/permit2.rs
tamaralipows: I thought it was because this is a bad pattern to possibly return an error here for a constant?

src/encoding/evm/approvals/permit2.rs
tamaralipows: agreed!

src/encoding/evm/approvals/permit2.rs
tamaralipows: Not sure if this is hard to do - I assumed not?

dianacarvalho1: huuuummmm I thought he was complaining about the pattern of making constants in the first place. this is not a constant. This is an attribute of `permit2`

tamaralipows: it's hardcoded here... isn't that the definition of a constant haha

tamaralipows: I won't push though if you think it's fine

dianacarvalho1: aaah I thought he meant literally using the `const` or the `lazy_static`

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+            pending_tx.get_receipt().await.unwrap()
+        });
+        assert!(result.status(), "Permit transaction failed");
+
+        // Assert that the allowance was set correctly in the permit2 contract
+        let (allowance_amount, _, nonce) = permit2
+            .get_existing_allowance(&anvil_account, &spender, &token)
+            .unwrap();
+        assert_eq!(allowance_amount, U160::from(biguint_to_u256(&amount)));
+        assert_eq!(nonce, U48::from(1));
```
dianacarvalho1, line 351: so.. with this test I could verify that the nonce actually increases as we call `permit`. Now.. do we want to account for this? 🤔 and that if the permit is already given for the amount and time we are looking for we should skip the permit and not pass any approval encoded data at all? 
I also checked that it is fine to call `permit` over and over again for the same token. I'm leaning to not address this for now.

tamaralipows: I don't think permit would be called twice with the exact same time... on the off-chance that ever happens, I think it's fine to just call it twice. Thanks for checking though! I think as long as we pass the proper nonce to the approval method it should be ok!

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+    address: Address,
+    client: Arc<RootProvider<BoxTransport>>,
+    runtime: Runtime,
+    signer: PrivateKeySigner,
+    chain_id: ChainId,
+}
+
+/// Type alias for representing allowance data as a tuple of (amount, expiration, nonce). Used for
+/// decoding
+type Allowance = (U160, U48, U48); // (amount, expiration, nonce)
```
tamaralipows: Double comment? lol

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+/// Expiration period for permits, set to 30 days (in seconds).
```
tamaralipows, line 38: Is this the norm? It's a quite long time, no?

tamaralipows: So the permit expiration is how long the permit lasts after we approve using the signature, right?

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+                let allowance: Allowance =
+                    Allowance::abi_decode(&response, true).map_err(|_| {
+                        EncodingError::FatalError(
+                            "Failed to decode response for permit2 allowance".to_string(),
+                        )
+                    })?;
+                Ok(allowance)
+            }
+            Err(err) => Err(EncodingError::RecoverableError(format!(
+                "Call to permit 2 allowance method failed with error: {:?}",
```
tamaralipows: Sorry nitpicking
```suggestion
                "Call to permit2 allowance method failed with error: {:?}",
```

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+                self.get_existing_allowance(&approval.owner, &approval.spender, &approval.token)?;
+            let expiration = U48::from(current_time + PERMIT_EXPIRATION);
+            let sig_deadline = U256::from(current_time + PERMIT_SIG_EXPIRATION);
+            let amount = U160::from(biguint_to_u256(&approval.amount));
+
+            // validations
+            assert!(MAX_UINT256 > sig_deadline, "signature deadline out of range");
+            assert!(MAX_UINT48 > nonce, "nonce out of range");
+            assert!(MAX_UINT160 > amount, "amount out of range");
+            assert!(MAX_UINT48 > expiration, "expiration out of range");
```
tamaralipows: These are pretty, but they cause a panic no? panic bad

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: yes, we are passing the proper nonce!

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+            if self.nonce != other.nonce {
+                return false;
+            }
+
+            true
+        }
+    }
+
+    #[test]
+    fn test_get_allowance_data() {
```
tamaralipows:
```suggestion
    fn test_get_existing_allowance() {
```

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+
+        let result = manager
+            .get_existing_allowance(&owner, &spender, &token)
+            .unwrap();
+        assert_eq!(
+            result,
+            (Uint::<160, 3>::from(0), Uint::<48, 1>::from(0), Uint::<48, 1>::from(0))
+        );
+    }
+    #[test]
```
tamaralipows, line 224: Blank line between tests?

src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,46 +1,361 @@
...
+        assert_eq!(
+            decoded_permit_single, expected_permit_single,
+            "Decoded PermitSingle does not match expected values"
+        );
+    }
+
+    /// This test actually calls the permit method on the Permit2 contract to verify the encoded
+    /// data works. It requires an Anvil fork, so please run with the following command: anvil
+    /// --fork-url <RPC-URL> And set up the following env var as ETH_RPC_URL=127.0.0.1:8545
+    /// Use an account from anvil to fill the anvil_account and anvil_private_key variables
```
tamaralipows, line 274: Thank you soooooooo much for figuring this out! I'm sure this forking will be a useful pattern in the future, so you've made all of our lives easier.

tamaralipows (commented):
I am very very happy with this! Thanks so much! The only real concern I have is about the `assert!`, other than that there are just tiny things

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: yes, 30 days is the value used at uniswap labs.

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: oh god

dianacarvalho1: true

dianacarvalho1: I will just remove them, yolo. I added them as an extra check that the problem was not here. but yeah

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: I think I will add a story called "improvements" and add a task where we would make the permit optional

tamaralipows (approved):
Thank you!!

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: added [here](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-4169)

PR #16, opened 2025-01-21
feat: UniswapV2 Swap Encoder

tamaralipows:


src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -16,17 +18,51 @@ pub struct UniswapV2SwapEncoder {
...
-        _encoding_context: EncodingContext,
+        swap: Swap,
+        encoding_context: EncodingContext,
     ) -> Result<Vec<u8>, EncodingError> {
-        todo!()
+        let token_in_address = bytes_to_address(&swap.token_in)?;
+        let token_out_address = bytes_to_address(&swap.token_out)?;
+
+        let zero_for_one = Self::get_zero_to_one(token_in_address, token_out_address);
+        let protocol_id = Bytes::from(
```
dianacarvalho1: component_id no? not protocol_id

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -16,17 +18,51 @@ pub struct UniswapV2SwapEncoder {
...
+            decode(
+                swap.component
+                    .id
+                    .trim_start_matches("0x"),
+            )
+            .map_err(|_| {
+                EncodingError::FatalError(format!(
+                    "Failed to parse component id: {}",
```
dianacarvalho1:
```suggestion
                    "Failed to parse component id for uniswap v2: {}",
```

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -16,17 +18,51 @@ pub struct UniswapV2SwapEncoder {
...
+                    swap.component.id
+                ))
+            })?,
+        );
+
+        // Sell token address is always needed to perform manual transfer from router into the pool,
+        // since no optimizations are performed that send from one pool to the next
```
dianacarvalho1:
```suggestion
        // Token in address is always needed to perform a manual transfer from the router into the pool,
        // since no optimizations are performed that send from one pool to the next
```

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -16,17 +18,51 @@ pub struct UniswapV2SwapEncoder {
...
+        let args = (
+            token_in_address,
+            bytes_to_address(&protocol_id)?,
+            bytes_to_address(&encoding_context.receiver)?,
+            zero_for_one,
+            encoding_context.exact_out,
+        );
+
+        Ok(args.abi_encode_packed())
```
dianacarvalho1, line 65: this works????? 👀

dianacarvalho1: oooohh very nice! I didn't know about this method! 👀 actually there are a few places where we should use it instead of regular `abi_encode`

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -77,8 +113,47 @@ impl SwapEncoder for BalancerV2SwapEncoder {
...
+            token_in: Bytes::from("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"),
+            token_out: Bytes::from("0x6b175474e89094c44da98b954eedeac495271d0f"),
+            split: 0f64,
+        };
+        let encoding_context = EncodingContext {
+            receiver: Bytes::from("0x0000000000000000000000000000000000000001"),
+            exact_out: false,
+            router_address: Bytes::zero(20),
+        };
+        let encoder = super::UniswapV2SwapEncoder::new(String::from("0x"));
```
dianacarvalho1: I think this `super` is unnecessary 🤔

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I only have minor comments!

dianacarvalho1 (approved):
Thank you 🙏🏼

PR #17, opened 2025-01-22
feat: add withdraw methods

royvardhan:
Added emergency withdraw methods for `FUND_RESCUER_ROLE`. 

- Validates receiver address, otherwise reverts. 
- Checks balance for both token and native.


foundry/src/TychoRouter.sol
```diff
@@ -60,9 +69,46 @@ contract TychoRouter is AccessControl {
...
+        external
+        onlyRole(FUND_RESCUER_ROLE)
+    {
+        if (receiver == address(0)) revert TychoRouter__InvalidReceiver();
+
+        uint256 amount = address(this).balance;
+        if (amount > 0) {
+            emit Withdrawal(address(0), amount);
+            // slither-disable-next-line arbitrary-send-eth
+            (bool success,) = receiver.call{value: amount}("");
```
royvardhan: Slither is complaining about low-level calls at the `withdrawNative` function since I am using `call()`, but I think it's a good way to send eth since `transfer()` and `send()` are limited to 2300 gas. 

cc: @dianacarvalho1 @tamaralipows

kayibal: Why would it be problematic to limit the transfer here to 2300 gas, assuming that the receiver is a simple wallet that does not invoke a fallback?

kayibal (approved):

foundry/src/TychoRouter.sol
tamaralipows: Also trying to understand why the limit is bad... I'd prefer to keep slither happy

foundry/src/TychoRouter.sol
```diff
@@ -3,10 +3,17 @@ pragma solidity ^0.8.28;
 
 import "@openzeppelin/contracts/access/AccessControl.sol";
 import "@permit2/src/interfaces/IAllowanceTransfer.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+error TychoRouter__WithdrawalFailed();
+error TychoRouter__InvalidReceiver();
```
tamaralipows, line 10: Never seen errors defined like this with the double underscore... is this the norm? Asking so I can learn too.. :)

foundry/src/TychoRouter.sol
royvardhan: If the receiver is a simple wallet then send is not bad. Otherwise, call() is the recommended way.

https://www.cyfrin.io/glossary/sending-ether-transfer-send-call-solidity-code-example

foundry/src/TychoRouter.sol
royvardhan: You can declare errors without double underscores. Like `error Unauthorised()` but when you are dealing with multiple contracts the convention is to use `error ContractName__UniqueErrorName` so that you can trace them better at tests. And I think we will have multiple contracts in this repo this is why I am using the convention.

Errors are better than require statement with String in it because they only store a 4-byte selector instead of a full string.

tamaralipows: Good to know - thanks!

foundry/src/TychoRouter.sol
dianacarvalho1: I think it's fine to limit it to 2300 gas and go the safer way!

foundry/src/TychoRouter.sol
```diff
@@ -17,6 +24,8 @@ contract TychoRouter is AccessControl {
     bytes32 public constant FUND_RESCUER_ROLE =
         0x912e45d663a6f4cc1d0491d8f046e06c616f40352565ea1cdb86a0e1aaefa41b;
 
+    event Withdrawal(address indexed token, uint256 amount);
```
tamaralipows: Should we maybe include the receiver in the event too?

foundry/src/TychoRouter.sol
royvardhan: Awesome, CI is good now.

foundry/src/TychoRouter.sol
royvardhan: good catch, I forgot.

foundry/src/TychoRouter.sol
```diff
@@ -60,9 +69,46 @@ contract TychoRouter is AccessControl {
...
+    function withdrawNative(address receiver)
+        external
+        onlyRole(FUND_RESCUER_ROLE)
+    {
+        if (receiver == address(0)) revert TychoRouter__InvalidReceiver();
+
+        uint256 amount = address(this).balance;
+        if (amount > 0) {
+            emit Withdrawal(address(0), amount, receiver);
+            // slither-disable-next-line arbitrary-send-eth
```
tamaralipows, line 106: We can remove this disable now :)

tamaralipows (approved):
Just the final comment about removing the disable then I think this is good!

foundry/src/TychoRouter.sol
royvardhan: It is still checking for the destination, so I kept it: 
image

dianacarvalho1 (approved):
Thank you so much @royvardhan ! 🙏🏼 It looks good to me!

foundry/src/TychoRouter.sol
tamaralipows: ok all good then!

PR #18, opened 2025-01-22
feat: Set swap executors and verifiers

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4044)

- Moved the deployment method into a test template for organization
- Created skeletons of dispatcher contracts
- Added all possible test cases for thoroughness

### DIFF
--- a/foundry/src/CallbackVerificationDispatcher.sol
+++ b/foundry/src/CallbackVerificationDispatcher.sol
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+/**
+ * @title Dispatch callback verification to external contracts
+ * @author PropellerHeads Devs
+ * @dev Provides the ability to delegate callback verification to external
+ *  contracts. This allows dynamically adding new supported protocols
+ *  without needing to upgrade any contracts. External contracts will
+ *  be called using delegatecall so they can share state with the main
+ *  contract if needed.
+ *
+ *  Note Verifier contracts need to implement the ICallbackVerifier interface
+ */
+contract CallbackVerificationDispatcher {
+    mapping(address => bool) public callbackVerifiers;
+}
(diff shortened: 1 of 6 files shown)


foundry/src/CallbackVerificationDispatcher.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+/**
+ * @title Dispatch callback verification to external contracts
+ * @author PropellerHeads Devs
+ * @dev Provides the ability to delegate callback verification to external
+ *  contracts. This allows dynamically adding new supported protocols
+ *  without needing to upgrade any contracts. External contracts will
+ *  be called using delegatecall so they can share state with the main
```
dianacarvalho1: do we really need delegate call here? or to be figured out in future tasks?

foundry/src/SwapExecutionDispatcher.sol
```diff
@@ -0,0 +1,17 @@
...
+ * @dev Provides the ability to delegate execution of swaps to external
+ *  contracts. This allows dynamically adding new supported protocols
+ *  without needing to upgrade any contracts. External contracts will
+ *  be called using delegatecall so they can share state with the main
+ *  contract if needed.
+ *
+ *  Note Executor contracts need to implement the ISwapExecutor interface
+ */
+contract SwapExecutionDispatcher {
+    mapping(address => bool) public swapExecutors;
```
dianacarvalho1, line 16: why does it need to be a mapping? 🤔 it's just the closest to a set we can do in solidity right?

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 Looks gooood ✨  very thorough test cases!

foundry/src/SwapExecutionDispatcher.sol
tamaralipows: Yes exactly T_T

foundry/src/CallbackVerificationDispatcher.sol
tamaralipows: No we don't good catch! Will update

PR #19, opened 2025-01-22
feat: add fee methods

royvardhan:
- Added `feeReceiver` state var + setter.
- Added `fee` state var + setter. 
- Setters can be called only with `FEE_SETTER_ROLE `

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -7,7 +7,7 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
 error TychoRouter__WithdrawalFailed();
-error TychoRouter__InvalidReceiver();
+error TychoRouter__AddressZero();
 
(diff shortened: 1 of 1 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -71,14 +81,34 @@ contract TychoRouter is AccessControl {
         // TODO
     }
 
+    /**
+     * @dev Allows setting the fee receiver.
+     */
+    function setFeeReceiver(address newfeeReceiver)
+        external
+        onlyRole(DEFAULT_ADMIN_ROLE)
```
tamaralipows: I think this should also be the FEE_SETTER_ROLE

tamaralipows (approved):
Thank you! I think this looks good, just please address the one comment

dianacarvalho1 (approved):
Thank you @royvardhan 🙏🏼 looks very good!

foundry/src/TychoRouter.sol
royvardhan: Done!

PR #21, opened 2025-01-23
feat: add LibPrefixLengthEncodedByteArray

royvardhan:
The idea was to convert [old lib](https://github.com/propeller-heads/defibot-contracts/blob/main/foundry/src/libraries/PrefixLengthEncodedByteArray.sol) from assembly to solidity, which I did [here](https://github.com/propeller-heads/tycho-execution/blob/f25da218d7b40878a61f6feb09f39c7fb06433f5/foundry/src/lib/LibPrefixLengthEncodedByteArray.sol) with all tests passing but then reverted to assembly because 'what works before should not be changed' especially in smart contracts. But I have kept the tests.


kayibal:
Thanks I think keeping this in assembly will help us to be gas efficient

tamaralipows (approved):
Thanks for this! The tradeoff for keeping it in assembly is it's harder to maintain/catch bugs, but I think that in this case, because it's just more of a helper method and not something like a main executor - it should be alright to keep the assembly :) and better for gas of course

dianacarvalho1 (approved):
Thank you @royvardhan 🙏🏼 sorry for the back and forth but I'm happy with the final decision: assembly + thorough tests!

royvardhan:
Thanks! @kayibal @tamaralipows @dianacarvalho1

royvardhan:
> Thank you @royvardhan 🙏🏼 sorry for the back and forth but I'm happy with the final decision: assembly + thorough tests!

No worries, back and forth is good!

PR #20, opened 2025-01-22
feat: delegatecall to executor in SwapExecutionDispatcher

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4043)

It may be easier to review by commit (or at least review the second-last commit separately since it's a refactor).

PR #23, opened 2025-01-23
feat: UniswapV2 SwapExecutor

dianacarvalho1:
Used version from commit `7aeb4cc5` in defibot-contracts as a starting point but I simplified it a lot:
- Removed all the exact out logic
- Remove all the payment logic - we support only normal transfers
- Remove ability to support not passing the token in, we need it every time

Question: Do we really like the name "SwapExecutor" ? 😅


foundry/src/interfaces/ISwapExecutor.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: UNLICENSED
```
dianacarvalho1: stole this from @tamaralipows [PR](https://github.com/propeller-heads/tycho-execution/pull/20)

tamaralipows: I don't actually know if this is correct - I stole this from the previous repo

foundry/test/executors/UniswapV2SwapExecutor.t.sol
```diff
@@ -0,0 +1,104 @@
...
+    function testSwap() public {
+        uint256 amountIn = 10 ** 18;
+        uint256 amountOut = 1847751195973566072891;
+        bool zeroForOne = false;
+        bool exactOut = true;
+        bytes memory protocolData = abi.encodePacked(
+            WETH_ADDR, WETH_DAI_POOL, BOB, zeroForOne, exactOut
+        );
+
+        vm.startPrank(ADMIN);
```
tamaralipows: Do we need to run this as `ADMIN`? Why?

foundry/src/executors/Uniswapv2SwapExecutor.sol
```diff
@@ -0,0 +1,73 @@
...
+    using SafeERC20 for IERC20;
+
+    function swap(uint256 givenAmount, bytes calldata data)
+        external
+        returns (uint256 calculatedAmount)
+    {
+        address target;
+        address receiver;
+        bool zeroForOne;
+        bool exactOut;
```
tamaralipows: Since we are not supporting buy orders, should we still encode and send `exactOut` to the executor? We aren't using it here, so I would say no.. and when the day comes that we want to support it again, we need to redesign the executor anyway, so we might as well redesign the encoding along with it (should be easy)

tamaralipows (commented):
Looking good, just a few comments. About `SwapExecutor` - can you suggest a better name?

foundry/src/interfaces/ISwapExecutor.sol
tamaralipows: OHHH you mean the whole file haha, not the license. Ok cool

dianacarvalho1: yes yes, the whole file sorry xD

foundry/src/executors/Uniswapv2SwapExecutor.sol
dianacarvalho1: agree! will remove it!

foundry/test/executors/UniswapV2SwapExecutor.t.sol
dianacarvalho1: huumm it doesn't need to be ADMIN. it can be any address at this point.. do you think I should use other than ADMIN?

dianacarvalho1:
> can you suggest a better name?

I would make it simpler.. I feel that SwapExecutor is too many words. What about just "Executor"? We only support swaps anyway

foundry/src/executors/Uniswapv2SwapExecutor.sol
```diff
@@ -0,0 +1,70 @@
...
+
+        IUniswapV2Pair pool = IUniswapV2Pair(target);
+        if (zeroForOne) {
+            pool.swap(0, calculatedAmount, receiver, "");
+        } else {
+            pool.swap(calculatedAmount, 0, receiver, "");
+        }
+    }
+
+    function _decodeData(bytes calldata data)
```
royvardhan: I am thinking we want to do some calldata validation here? Three addresses plus a bool would be 61 bytes, so `data.length >= 61` check can be added here.

royvardhan (commented):
Looks good to me!

royvardhan:
> > can you suggest a better name?
> 
> I would make it simpler.. I feel that SwapExecutor is too many words. What about just "Executor"? We only support swaps anyway

Yes the `Executor` is a lot better and goes with `Tycho Execution` naming.

foundry/test/executors/UniswapV2SwapExecutor.t.sol
tamaralipows: Do we need to use `vm.startPrank` in this case? I've only been using that to change the caller

tamaralipows:
I like executor too! Feel free to change the name :)

foundry/test/executors/UniswapV2SwapExecutor.t.sol
dianacarvalho1: good catch! no we don't!! removing it!

tamaralipows (approved):
Yay thank you!

PR #24, opened 2025-01-24
feat: add pause/unpause methods

royvardhan:
Added OZ Pausable implementation with `pause` and `unpause` controlled by `PAUSER_ROLE` and `UNPAUSER_ROLE`


dianacarvalho1 (approved):
Veeeery nice! Thank you @royvardhan ! 🙏🏼

foundry/test/TychoRouter.t.sol
```diff
@@ -206,4 +206,29 @@ contract TychoRouterTest is TychoRouterTestSetup {
         tychoRouter.setFeeReceiver(FEE_RECEIVER);
         vm.stopPrank();
     }
+
+    function testPause() public {
+        vm.startPrank(PAUSER);
+        assertEq(tychoRouter.paused(), false);
+        tychoRouter.pause();
+        assertEq(tychoRouter.paused(), true);
+        vm.stopPrank();
```
tamaralipows, line 186: Can we just add a `// TODO` here to try to call swap once we implement it? And make sure it doesn't work? I think that's the most important part

foundry/test/TychoRouter.t.sol
```diff
@@ -206,4 +206,29 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+
+        vm.startPrank(UNPAUSER);
+        tychoRouter.unpause();
+        assertEq(tychoRouter.paused(), false);
+        vm.stopPrank();
+    }
+
+    function testPauseFailures() public {
```
tamaralipows: Maybe specify in the name that it's `testPauseWrongRole`

royvardhan: Hmmm the same test uses the correct role to `unpause` with `UNPAUSER` role to demonstrate that even the `UNPAUSER` cannot unpause if the correct is not paused.

foundry/test/TychoRouter.t.sol
royvardhan: Done!

foundry/test/TychoRouter.t.sol
tamaralipows: I think these should be two different tests in that case, with the titles clearly saying which case is being tested :)

royvardhan: Done, thanks for reviewing.

tamaralipows (approved):
Thank you!

PR #25, opened 2025-01-24
feat: Swap public method

dianacarvalho1:
Implements:
-  `swap` public method 
   - handles approvals
   - checks amount
   - wrap and unwrap ETH
   - takes fee
- `_splitSwap`
- Adds tests for **everything**


foundry/src/TychoRouter.sol
```diff
@@ -61,22 +71,85 @@ contract TychoRouter is
...
+            _wrapETH(amountIn);
+        } else {
+            permit2.permit(msg.sender, permitSingle, signature);
+            permit2.transferFrom(
+                msg.sender,
+                address(this),
+                uint160(amountIn),
+                permitSingle.details.token
+            );
+        }
```
tamaralipows: We actually need to do an explicit check that `tokenIn` is not native ETH, that's why I added it to the fn sig. Remember that for `UniswapV4` the sell token may be native ETH and we may not want to wrap it - i.e. there are pools accepting native ETH as is. 

And of course in this case, no permit2 needed

foundry/src/TychoRouter.sol
```diff
@@ -61,22 +71,85 @@ contract TychoRouter is
...
      *  splits, checking that the user gets more than minUserAmount of buyToken.
      */
     function swap(
         uint256 amountIn,
-        address tokenIn,
-        uint256 minUserAmount,
-        bool wrapEth,
-        bool unwrapEth,
+        address tokenOut,
+        uint256 checkAmountOut,
```
tamaralipows: Can we just call it `minAmountOut` for clarity? This naming here sounds like it's a `bool` not an amount. This swap will always be sell orders so no need for this vague `checkedAmount` lingo I think

foundry/src/Swap.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.28;
+
+library Swap {
```
dianacarvalho1: should I include the executor address and selector here? I'm thinking yes

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,115 @@ contract TychoRouter is
...
+        amountOut = _splitSwap(amountIn, nTokens, swaps);
+
+        if (fee > 0) {
+            uint256 feeAmount = (amountOut * fee) / 10000;
+            amountOut -= feeAmount;
+            IERC20(tokenOut).safeTransfer(feeReceiver, feeAmount);
+            if (unwrapEth == false) {
+                IERC20(tokenOut).safeTransfer(receiver, amountOut);
+            }
+        }
```
dianacarvalho1: sooo this makes me uncomfortable.. 
I was assuming that the receiver would be the user and the last pool could already send the tokens there. BUT if we want to take a fee, we need to send the funds to the router and only then take a fee and send to the user.
should we just assume that all pools will always send to the router? even if it's the last swap? 😬

tamaralipows: I don't think this needs to be the concern of the router at all no, no? I think we can make sure that this all makes sense in the encoding?

That was we can either optimize this (send to router on the final swap only when there is a fee or unwrapping)
or leave it unoptimized and simple (send to the router after every swap). The best part is we don't need to make this decision now since it won't require any redeploying of the router after, right?

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,115 @@ contract TychoRouter is
...
-        amountOut = 0;
-        // TODO
+        bytes calldata signature,
+        bytes calldata swaps
+    ) external payable whenNotPaused returns (uint256 amountOut) {
+        // For native ETH, assume funds already in our router. Else, transfer and handle approval.
+        if (wrapEth) {
+            _wrapETH(amountIn);
+        } else {
+            if (tokenIn != address(0)) {
```
tamaralipows: wouldn't an `else if` suffice to avoid nested if statements?

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,115 @@ contract TychoRouter is
...
+     *
+     * @return amountOut The total amount of the output token received by the receiver, after deducting fees if applicable.
      */
     function swap(
         uint256 amountIn,
         address tokenIn,
-        uint256 minUserAmount,
+        address tokenOut,
+        bool checkMinAmount,
+        uint256 minAmountOut,
```
tamaralipows: Just a thought, can't we just set `minAmountOut` to 0 when we don't care about checking? Or do you think it's better to leave this explicit?

foundry/test/TychoRouter.t.sol
```diff
@@ -189,4 +189,372 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        uint256 amount = 1 ether;
+        deal(WETH_ADDR, address(tychoRouter), amount);
+
+        tychoRouter.unwrapETH(amount);
+
+        assertEq(address(tychoRouter).balance, amount);
+        assertEq(IERC20(WETH_ADDR).balanceOf(address(tychoRouter)), 0);
+    }
+
+    function testSplitSwapSimple() public {
```
tamaralipows: Maybe remove split from the names where there is no split happening? I can imagine this getting confusing

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,115 @@ contract TychoRouter is
...
+            amountOut -= feeAmount;
+            IERC20(tokenOut).safeTransfer(feeReceiver, feeAmount);
+            if (unwrapEth == false) {
+                IERC20(tokenOut).safeTransfer(receiver, amountOut);
+            }
+        }
+
+        if (checkMinAmount && amountOut < minAmountOut) {
+            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
+        }
```
tamaralipows: Just a note - I think we need to mention in our docs that we check the amount before fees not after

foundry/src/Swap.sol
tamaralipows: **ahhh this is a great idea**, in that case it's not necessary to decode them inside the executor dispatcher itself - we can just pass it. This removes a lot of the inconsistencies that we struggled with last time with some things being decoded, others being part of the swap struct, and other things getting passed. And especially since every single swap will have an executor address and selector this is a great idea.

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -1,28 +1,61 @@
...
-        tychoRouter = new TychoRouter(permit2Address);
+        tychoRouter = new TychoRouterExposed(permit2Address, WETH_ADDR);
         tychoRouter.grantRole(keccak256("EXECUTOR_SETTER_ROLE"), BOB);
         tychoRouter.grantRole(keccak256("FUND_RESCUER_ROLE"), FUND_RESCUER);
         tychoRouter.grantRole(keccak256("FEE_SETTER_ROLE"), FEE_SETTER);
         tychoRouter.grantRole(keccak256("PAUSER_ROLE"), PAUSER);
         tychoRouter.grantRole(keccak256("UNPAUSER_ROLE"), UNPAUSER);
+        tychoRouter.grantRole(
+            keccak256("EXECUTOR_SETTER_ROLE"), EXECUTOR_SETTER
+        );
```
tamaralipows, line 50: Above this, we have already set bob as the executor setter. Maybe remove him then? Or document your reasoning for doing this

foundry/test/Constants.sol
```diff
@@ -9,6 +9,10 @@ contract Constants is Test {
     address FUND_RESCUER = makeAddr("fundRescuer");
     address FEE_SETTER = makeAddr("feeSetter");
     address FEE_RECEIVER = makeAddr("feeReceiver");
+    address EXECUTOR_SETTER = makeAddr("executorSetter");
+    address ALICE = 0xcd09f75E2BF2A4d11F3AB23f1389FcC1621c0cc2;
+    uint256 ALICE_PK =
+        0x123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234;
```
tamaralipows, line 15: Curious, why did you add Alice when we already have Bob

tamaralipows (commented):
Very nice track! I hope my answers to your questions make sense, let me know otherwise! Other than that, just little things / other questions

foundry/src/TychoRouter.sol
dianacarvalho1: >  I think we can make sure that this all makes sense in the encoding?

but in encoding we don't know if we need to take a fee or not.. we would need to query the router to see if the fee is set 😬

dianacarvalho1: > The best part is we don't need to make this decision now since it won't require any redeploying of the router after, right?

we do need to make this decision now. because if we assume that the last pool sends the tokens, then in this method we don't need anything else. if it doesn't, we still need to send the tokens here to the user

foundry/test/TychoRouter.t.sol
dianacarvalho1: yeah but it is testing the splitSwap method 😕

foundry/src/TychoRouter.sol
dianacarvalho1: do you think it should be the other way around?

foundry/test/Constants.sol
dianacarvalho1: because I don't know Bob's private key. And to make alice I created a random private key and then computed the address from it. We need a private key that makes sense for the permit2 approvals

foundry/test/TychoRouterTestSetup.sol
dianacarvalho1: yeah I will just use the EXECUTOR_SETTER everywhere. just to be coherent on how this will look like in production

foundry/src/Swap.sol
royvardhan: Shall this be moved into /lib and also renamed to LibSwap?

foundry/src/Swap.sol
```diff
@@ -0,0 +1,40 @@
...
+    /// Returns the InToken index into an array of tokens
+    function tokenInIndex(bytes calldata swap)
```
royvardhan: Q: Should we be checking the calldata sanity? Like `if (swap.length < MIN_SWAP_LENGTH) revert InvalidSwapLength();`

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,111 @@ contract TychoRouter is
...
-        bytes calldata swaps,
+        address receiver,
         IAllowanceTransfer.PermitSingle calldata permitSingle,
-        bytes calldata signature
-    ) external whenNotPaused returns (uint256 amountOut) {
-        amountOut = 0;
-        // TODO
+        bytes calldata signature,
+        bytes calldata swaps
+    ) external payable whenNotPaused returns (uint256 amountOut) {
```
royvardhan: Perhaps, adding a reentrancy guard would be a good idea since this is now payable?

royvardhan (commented):
Nit comments and suggestions.

foundry/src/Swap.sol
dianacarvalho1: but where would we check that? 🤔 in all of these methods? 😬 or inside the _splitSwap method?

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,115 @@ contract TychoRouter is
...
+
+        if (minAmountOut > 0 && amountOut < minAmountOut) {
+            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
+        }
+
+        if (unwrapEth) {
+            _unwrapETH(amountOut);
+            payable(receiver).transfer(amountOut);
+        }
+    }
```
royvardhan, line 165: Should emit a swap event

foundry/src/TychoRouter.sol
```diff
@@ -77,22 +87,115 @@ contract TychoRouter is
...
      */
     function swap(
         uint256 amountIn,
         address tokenIn,
-        uint256 minUserAmount,
+        address tokenOut,
+        uint256 minAmountOut,
         bool wrapEth,
         bool unwrapEth,
         uint256 nTokens,
```
royvardhan, line 125: I also think that we should do something like this:

```
uint256 actualTokenCount = _validateSwapsTokenCount(swaps);
        if (actualTokenCount != nTokens) {
            revert TychoRouter__InvalidTokenCount(nTokens, actualTokenCount);
        }
```

foundry/src/TychoRouter.sol
dianacarvalho1: can you explain why you think it is a good idea please? How could this method be abused?

foundry/src/TychoRouter.sol
dianacarvalho1: huumm do we want a swap event? I was not planning on having one tbh. emitting events is rather expensive

foundry/src/Swap.sol
royvardhan: I was thinking about the lib before reading the rest of the code below.

foundry/src/TychoRouter.sol
royvardhan: I strongly suggest emitting an event on the router swap for indexing and data presentation purposes. "Hey, our router did $1B in volume this Friday"

foundry/src/TychoRouter.sol
royvardhan: Cannot come up with an immediate attack idea, I saw payable and fallback and suggested reentrancy guard haha! Okay, what happens when swap reverts after wrapping eth?

foundry/src/TychoRouter.sol
dianacarvalho1: that is a bit tricky to do without going through the swaps 😬 I think this would be too gas expensive to be worth doing tbh

foundry/src/TychoRouter.sol
dianacarvalho1: but it will be the users that need to pay for this event 😕 we are already not doing a lot of gas optimizations, I feel uneasy in adding one more gas expensive thing just for us to know. We could index the txs later on in a different way anyway (it will be harder)

royvardhan: Okay we have an agreement then, I am resolving this. :)

foundry/src/TychoRouter.sol
royvardhan: Got it!

royvardhan (approved):
Looks good to me! Thank you @dianacarvalho1

foundry/test/TychoRouter.t.sol
tamaralipows: I thought we weren't going to call it the splitSwap method anymore - just swap. Why still call the internal one splitSwap?

foundry/src/TychoRouter.sol
tamaralipows: I don't know tbh... maybe? Otherwise the client needs to be constantly aware of our fees, keep track of if it has changed.... if there is a fee bug they can also lose money

foundry/src/TychoRouter.sol
tamaralipows: > but in encoding we don't know if we need to take a fee or not.. 

Uff okay, in this case the only option is to send to the router at the very end every time

When I said we don't need to decide now, I was strictly referring to the fact that this can all be taken care of in encoding. If not, then yeah let's assume it's always sent to the router

foundry/src/ExecutionDispatcher.sol
```diff
@@ -50,7 +50,7 @@ contract ExecutionDispatcher {
      * @dev Calls an executor, assumes swap.protocolData contains
      *  protocol-specific data required by the executor.
      */
-    // slither-disable-next-line dead-code
+    // slither-disable-next-line delegatecall-loop
```
tamaralipows, line 53: Why did slither cry about this?

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -10,6 +10,7 @@ error UniswapV2Executor__InvalidDataLength();
 contract UniswapV2Executor is IExecutor {
     using SafeERC20 for IERC20;
 
+    // slither-disable-next-line locked-ether
```
tamaralipows, line 13: Is this because it's payable and we don't have a withdraw method?

foundry/src/TychoRouter.sol
dianacarvalho1: >  I was strictly referring to the fact that this can all be taken care of in encoding

yeah, that is not true 😕 or should we encode in the calldata a bool to say if the end transfer is needed? 🤔

foundry/test/TychoRouter.t.sol
dianacarvalho1: huumm fair enough xD should it be `swap` and `_swap`? 😂

foundry/src/TychoRouter.sol
dianacarvalho1: but it would be really shitty for the user to set a check amount and then the check passes but after fees it is less than what the user wanted 😬

foundry/src/executors/UniswapV2Executor.sol
dianacarvalho1: exactly

foundry/src/ExecutionDispatcher.sol
dianacarvalho1: it gives me this
````
ExecutionDispatcher._callExecutor(address,bytes4,uint256,bytes) (src/ExecutionDispatcher.sol#53-80) has delegatecall inside a loop in a payable function: (success,result) = executor.delegatecall(abi.encodeWithSelector(selector,amount,data)) (src/ExecutionDispatcher.sol#65-67)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation/#payable-functions-using-delegatecall-inside-a-loop
````

File: foundry/src/TychoRouter.sol (hunk 4 above)
Comment (tamaralipows) [reply foundry/src/TychoRouter.sol]: ai that sounds too messy. I vote to just send to the router at each step

File: foundry/test/TychoRouter.t.sol (hunk 7 above)
Comment (tamaralipows) [reply foundry/test/TychoRouter.t.sol]: yes haha

File: foundry/src/TychoRouter.sol (hunk 8 above)
Comment (tamaralipows) [reply foundry/src/TychoRouter.sol]: yes I agree, I was saying the same point xD lets put it before

Comment (tamaralipows) [reply foundry/src/TychoRouter.sol]: BUT WAIT, then solvers will need to know about this fee when searching for their routes....

Comment (dianacarvalho1) [reply foundry/src/TychoRouter.sol]: I'm confused.. we should put it after fees no?? like this
```

        if (fee > 0) {
            uint256 feeAmount = (amountOut * fee) / 10000;
            amountOut -= feeAmount;
        }

        if (minAmountOut > 0 && amountOut < minAmountOut) {
            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
        }
```

Comment (dianacarvalho1) [reply foundry/src/TychoRouter.sol]: yeah true.. 😬 "ah let's add the fee feature, it will be super simple" they said

Comment (tamaralipows) [reply foundry/src/TychoRouter.sol]: > we should put it after fees no?? like this
 
 yes and then document it well, that their solvers need to take our router fees into consideration when finding solutions

Comment (tamaralipows) [reply foundry/src/TychoRouter.sol]: > I'm confused.. we should put it after fees no?? like this

exactly

File: foundry/test/TychoRouterTestSetup.sol [hunk 17]
```diff
@@ -1,28 +1,61 @@
...
+
+    function wrapETH(uint256 amount) external payable {
+        return _wrapETH(amount);
+    }
+
+    function unwrapETH(uint256 amount) external {
+        return _unwrapETH(amount);
+    }
+
+    function ExposedSwap(
```
tamaralipows, line 21:
```suggestion
    function exposedSwap(
```

dianacarvalho1: nooooooo too late 😂

PR #26, opened 2025-01-24
feat: add balancer v2 encoder test

royvardhan:
Added encoding test for Balancer V2 and made them synchronous.

### DIFF
--- a/src/encoding/evm/swap_encoder/encoders.rs
+++ b/src/encoding/evm/swap_encoder/encoders.rs
@@ -106,7 +106,7 @@ impl SwapEncoder for BalancerV2SwapEncoder {
             encoding_context.exact_out,
             approval_needed,
         );
-        Ok(args.abi_encode())
+        Ok(args.abi_encode_packed())
     }
 
     fn executor_address(&self) -> &str {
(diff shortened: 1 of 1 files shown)


dianacarvalho1 (approved):
Thank you 🙏🏼 seems good to me! (I think the description is out of date though)

royvardhan:
> Thank you 🙏🏼 seems good to me! (I think the description is out of date though)

Thanks for reviewing! Updated.

tamaralipows (approved):
Thank you! Looks great!

royvardhan:
Thank you @tamaralipows 🙏🏻

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -159,4 +159,47 @@ mod tests {
...
+
+        assert_eq!(
+            hex_swap,
+            String::from(concat!(
+                // token in
+                "c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+                // token out
+                "6b175474e89094c44da98b954eedeac495271d0f",
+                // pool id 
+                "307838386536413063326444443236464545623634463033396132633431323936466342336635363430",
```
dianacarvalho1, line 195: just realised we have a bug here @royvardhan ! This is not the pool ID 😬 in the encoding you need to convert the component id from String to bytes with something like
```
Bytes::from_str(&swap.component.id)?
```
(it needs to be alloy Bytes not tycho Bytes)

royvardhan: Fixed this in this pr: https://github.com/propeller-heads/tycho-execution/pull/29

PR #27, opened 2025-01-24
feat: Perform staticcall to CallbackVerifier

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4046)


foundry/src/CallbackVerificationDispatcher.sol
```diff
@@ -12,4 +17,63 @@ pragma solidity ^0.8.0;
...
+
+        if (!callbackVerifiers[verifier]) {
+            revert CallbackVerificationDispatcher__UnapprovedVerifier();
+        }
+
+        bytes4 selector = decodedSelector == bytes4(0)
+            ? ICallbackVerifier.verifyCallback.selector
+            : decodedSelector;
+
+        address sender = msg.sender;
```
dianacarvalho1, line 74: why is this necessary?

foundry/test/CallbackVerificationDispatcher.t.sol
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    function testCallVerifierNoSelector() public {
+        // For this test, we can use any callback verifier and any calldata that we
+        // know works for this verifier. We don't care about which calldata/executor,
+        // since we are only testing the functionality of the staticcall and not
+        // the inner verifier.
+        // Thus, this test case designed from scratch using previously-deployed
+        // Maverick callback verifier. Looking at the code, we can easily design
+        // passing calldata.
```
dianacarvalho1: instead of writing this again maybe we could just say that this is the same as the previous test but we modified the call data not to specify a selector

foundry/src/CallbackVerificationDispatcher.sol
```diff
@@ -12,4 +17,88 @@ pragma solidity ^0.8.0;
...
+
+        // slither-disable-next-line low-level-calls
+        (bool success, bytes memory result) = verifier.staticcall(
```
dianacarvalho1, line 77: veeery niceeee

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 I only have very minor comments! I also agree that the setting/removing logic should be in the dispatcher and not in the router

royvardhan (approved):
This looks good to me! 🙏🏻

foundry/src/CallbackVerificationDispatcher.sol
tamaralipows: We send this address into the verifier contract to ensure that the caller is a Maverick/USV3/Pancake pool and not an attacker

dianacarvalho1: ah I was asking why we needed to define a new variable for it in the first place..

PR #28, opened 2025-01-24
feat: USV3 Swap Encoder

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4067)

[copied from commit message]
- No more supply owed action type or action data - we will always transfer from our router using a regular transfer in the callback.
- The receiver will always be sent with the actual swap call, meaning that no forward received action type or data is necessary either.
- We decode the pool fee as le-bytes and not be-bytes because that's how it's stored: https://github.com/propeller-heads/tycho-protocol-sdk/blob/3c08359cf112e15c137dd5256b8dc8e9cd6c1626/substreams/ethereum-uniswap-v3/src/modules/1_map_pool_created.rs#L64

### DIFF
--- a/src/encoding/evm/swap_encoder/encoders.rs
+++ b/src/encoding/evm/swap_encoder/encoders.rs
@@ -70,6 +70,84 @@ impl SwapEncoder for UniswapV2SwapEncoder {
     }
 }
 
+pub struct UniswapV3SwapEncoder {
+    executor_address: String,
+}
+
+impl UniswapV3SwapEncoder {
+    fn get_zero_to_one(sell_token_address: Address, buy_token_address: Address) -> bool {
+        sell_token_address < buy_token_address
+    }
+}
+
+impl SwapEncoder for UniswapV3SwapEncoder {
+    fn new(executor_address: String) -> Self {
+        Self { executor_address }
+    }
+
+    fn encode_swap(
+        &self,
+        swap: Swap,
+        encoding_context: EncodingContext,
+    ) -> Result<Vec<u8>, EncodingError> {
+        let token_in_address = bytes_to_address(&swap.token_in)?;
+        let token_out_address = bytes_to_address(&swap.token_out)?;
+
+        let zero_for_one = Self::get_zero_to_one(token_in_address, token_out_address);
+        let component_id = Bytes::from(
+            decode(
+                swap.component
+                    .id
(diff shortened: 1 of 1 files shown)


src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -159,6 +239,53 @@ mod tests {
             ))
         );
     }
+    #[tokio::test]
```
tamaralipows: I see we are no longer using `tokio`. Will change this before merging.

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -70,6 +70,84 @@ impl SwapEncoder for UniswapV2SwapEncoder {
...
+
+    fn encode_swap(
+        &self,
+        swap: Swap,
+        encoding_context: EncodingContext,
+    ) -> Result<Vec<u8>, EncodingError> {
+        let token_in_address = bytes_to_address(&swap.token_in)?;
+        let token_out_address = bytes_to_address(&swap.token_out)?;
+
+        let zero_for_one = Self::get_zero_to_one(token_in_address, token_out_address);
```
dianacarvalho1: `zero_to_one` no?

src/encoding/evm/swap_encoder/encoders.rs
dianacarvalho1: yeah, it's not async. there is not need for tokio here

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -70,6 +70,84 @@ impl SwapEncoder for UniswapV2SwapEncoder {
...
+                    .id
+                    .trim_start_matches("0x"),
+            )
+            .map_err(|_| {
+                EncodingError::FatalError(format!(
+                    "Failed to parse component id for Uniswap v3: {}",
+                    swap.component.id
+                ))
+            })?,
+        );
```
dianacarvalho1: do we need all of this?? I think something like this should be enough
```
Address::from_str(&swap.component.id)?
```

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 Only have small suggestions!

royvardhan (approved):
Looks clean, thank you!

PR #29, opened 2025-01-27
feat: add balancer v2 executor, tests and fix encoder bug

royvardhan:
[Issue](https://datarevenue.atlassian.net/jira/software/projects/ENG/issues/ENG-4035?jql=project%20%3D%20%22ENG%22%20ORDER%20BY%20created%20DESC)


foundry/lib/balancer-labs/v2-interfaces/CHANGELOG.md
```diff
@@ -0,0 +1,57 @@
+# Changelog
```
tamaralipows: How did you add this lib? We have been adding it as a submodule by doing
`forge install --no-commit libname`

This way we don't commit so many files.
Sorry, this should have been documented somewhere

foundry/src/executors/BalancerV2Executor.sol
```diff
@@ -0,0 +1,91 @@
...
+        external
+        returns (uint256 calculatedAmount)
+    {
+      
+        (
+            IERC20 tokenIn,
+            IERC20 tokenOut,
+            bytes32 poolId,
+            address receiver,
+            bool exactOut,
```
tamaralipows: I just realized that we missed this when reviewing the `BalancerV2SwapEncoder`

We no longer support exact out - only sell orders (for now). So you can remove that from the encoding and also from this decoding. Then hardcode the `kind` to `IVault.SwapKind.GIVEN_IN` below (as well as the limit)

tamaralipows (commented):
Thank you! Looks good - looking forward to the tests

foundry/lib/balancer-labs/v2-interfaces/CHANGELOG.md
royvardhan: Actually, I tried using forge install but it couldn't find the lib and had to use yarn and move them to lib manually. Not something I wanted :(  .. and it does not look clean. Let me try something.

foundry/src/executors/BalancerV2Executor.sol
royvardhan: Thank you for pointing this!

foundry/lib/balancer-labs/v2-interfaces/CHANGELOG.md
royvardhan: I have now installed balancer-v2 pkg using yarn because foundry install wasn't working. Updated gitignore cc: @dianacarvalho1

dianacarvalho1: why wasn't foundry install working?

royvardhan: It wasn't able to find the lib pkg, can you try `forge install @balancer-labs/v2-interfaces --no-commit` in the repo and see if it's working for you?

tamaralipows: I think it's this
`forge install balancer/balancer-v2-monorepo --no-commit`

And then in remappings.txt add this
@balancer-labs/v2-interfaces=lib/balancer-v2-monorepo/pkg/interfaces

That's just what chatgpt told me. The first step seems to work

foundry/src/executors/BalancerV2Executor.sol
```diff
@@ -0,0 +1,83 @@
...
+            IERC20 tokenIn,
+            IERC20 tokenOut,
+            bytes32 poolId,
+            address receiver,
+            bool needsApproval
+        ) = _decodeData(data);
+
+        if (needsApproval) {
+            // slither-disable-next-line unused-return
+            tokenIn.approve(VAULT, givenAmount);
```
tamaralipows: For simplicity we are actually just setting this to the max amount every time

foundry/lib/balancer-labs/v2-interfaces/CHANGELOG.md
royvardhan: This helped, thanks a lot!

tamaralipows (commented):
Thank you! It's looking great now, especially the submodules issue of before - I just have one main issue with the approve amount!

In our encoder side - we are only checking that this amount is non-zero to say that it doesn't need an approval. Therefore, approving the max value is simply the easiest and most gas-efficient way to go about this (else we will need to re-approve way more often)

Let me know if that makes sense to you

dianacarvalho1 (commented):
Thank you @royvardhan 🙏🏼 I agree with @tamaralipows comment and the rest looks good to me!

foundry/src/executors/BalancerV2Executor.sol
royvardhan: Yes, that makes sense to me. Done!

royvardhan:
Thank you good frens for helping out today. @tamaralipows @dianacarvalho1

dianacarvalho1 (approved):
✨

tamaralipows (approved):

PR #31, opened 2025-01-28
feat: Implement generic callback

dianacarvalho1:
Since this is so simple I didn't add a test for it. The test will be the whole flow for uniswap v3 (we still miss the executor and the verifier for that though)

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -79,7 +79,22 @@ contract TychoRouter is
      *  caller is not a pool.
      */
     fallback() external {
-        // TODO execute generic callback
+        _executeGenericCallback(msg.data);
+    }
+
+    /**
+     * @dev Check if the sender is correct and executes callback actions.
+     *  @param msgData encoded data. It must includes data for the verification.
+     */
+    function _executeGenericCallback(bytes calldata msgData) internal {
+        (
+            uint256 amountOwed,
+            uint256 amountReceived,
+            address tokenOwed,
+            uint16 offset // I think we actually don't need this!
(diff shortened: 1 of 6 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -71,7 +71,22 @@ contract TychoRouter is
...
+    /**
+     * @dev Check if the sender is correct and executes callback actions.
+     *  @param msgData encoded data. It must includes data for the verification.
+     */
+    function _executeGenericCallback(bytes calldata msgData) internal {
+        (
+            uint256 amountOwed,
+            uint256 amountReceived,
+            address tokenOwed,
+            uint16 offset // I think we actually don't need this!
```
tamaralipows: Agreed. Just came across this too in my PR

tamaralipows: The sad part is then I can't use the same integration test as I was using before in the verifier  :p oh well. We will have a bigger overall integration test anyway

tamaralipows (approved):
Thank you!

foundry/src/TychoRouter.sol
```diff
@@ -71,7 +71,22 @@ contract TychoRouter is
...
+        ) = _callVerifyCallback(msgData);
```
royvardhan: Yeah since we are not using offset from the decoding, we should remove it.

royvardhan (approved):

foundry/src/TychoRouter.sol
dianacarvalho1: removed it and updated the tests!

tamaralipows (approved):
Re-approving. Thank you!

PR #32, opened 2025-01-29
feat: USV3 callback verification

tamaralipows:
**Problem:**
We cannot do things the way we planned.
Originally, we were doing this in our main router contract:
```
        CallbackValidationV2.verifyCallback(
            _usv3Factory,
            address(header.tokenIn),
            address(header.tokenOut),
            header.fee
);
```
       
Now, we had the idea to move the verification into the master verification contract, which is called using a `staticcall`. When we call it using a `staticcall`, `msg.sender` is our router contract and not the USV3 pool. This is bad, since the inner workings of `CallbackValidationV2.verifyCallback` rely on `msg.sender` being the pool contract. See here: https://github.com/Uniswap/v3-periphery/blob/0682387198a24c7cd63566a2c58398533860a5d1/contracts/libraries/CallbackValidation.sol#L34

**To overcome this, I can think of three possible solutions:**
1. Copy over the contract methods from v3-periphery to allow us to pass msg.sender (I attempted this in my PR and it seems like a bit much in the end, I don't like it much).
2. Use a `delegatecall` instead of `staticcall` in the dispatcher so that the original `msg.sender` is preserved (Max mentioned we should avoid `delegatecall`s.
3. Hardcode the UniswapV3 callback in the main router like we had in the previous router. This means that for now, there will be no external validation contracts (I think this is fine - at least we have prepared for them existing in the future so maybe it's not so bad?).

**EDIT: We have decided to go with the third option.**


royvardhan:
I would suggest the third option and it's simple.

foundry/src/TychoRouter.sol
```diff
@@ -128,7 +136,9 @@ contract TychoRouter is
         bytes calldata signature,
         bytes calldata swaps
     ) external payable whenNotPaused nonReentrant returns (uint256 amountOut) {
-        require(receiver != address(0), "Invalid receiver address");
+        if (receiver == address(0)) {
+            revert TychoRouter__AddressZero();
+        }
```
dianacarvalho1, line 141: thank you 🙏🏼

foundry/src/TychoRouter.sol
```diff
@@ -340,4 +350,38 @@ contract TychoRouter is
      * @dev Allows this contract to receive native token
      */
     receive() external payable {}
+
+    /**
+     * @dev Called by UniswapV3 pool when swapping on it.
+     * See in IUniswapV3SwapCallback for documentation.
+     */
+    function uniswapV3SwapCallback(
```
dianacarvalho1, line 358: and then where is this called? 🤔

dianacarvalho1: can we have a test as well? 🥺

dianacarvalho1: or do you think having the integration one will be enough?

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 This was a hard baby! 😓 
I have one question

foundry/src/TychoRouter.sol
tamaralipows: This is called by the UniswapV3 pool and I think it doesn't make sense outside of the integration context. I can add one in my next PR after making an integration test though!

tamaralipows: Just feel I'm more likely to catch a bug by making the integration test first. and this PR is getting too messy.

dianacarvalho1 (approved):
Sounds good! ✨

PR #33, opened 2025-01-29
feat: update ExecutorEncoder and test module with contract

royvardhan:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4082)


src/encoding/strategy_encoder.rs
```diff
@@ -1,8 +1,10 @@
+use alloy_primitives::Address;
```
royvardhan: I am using Alloy address, is this good? @dianacarvalho1

dianacarvalho1: soooo, alloy is specific to evm. We want to make tycho as general as possible. so in the interfaces please use Bytes from tycho core `use tycho_core::Bytes;`. then inside the evm specific implementation, yes, we should use alloy!
There is a feature gate that makes sure that this is maintained (more on features [here](https://doc.rust-lang.org/cargo/reference/features.html))! That is why CI is failing

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -43,7 +43,7 @@ contract UniswapV2Executor is IExecutor {
             bool zeroForOne
         )
     {
-        if (data.length != 61) {
+        if (data.length < 61) {
```
dianacarvalho1: can it be more?? in which situations?

src/encoding/strategy_encoder.rs
dianacarvalho1: so here it should be Bytes

dianacarvalho1: this kinda sucks but I hope this generalisation will help us later on..

src/encoding/evm/strategy_encoder/encoder.rs
```diff
@@ -37,10 +39,10 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
 
 /// This strategy encoder is used for solutions that are sent directly to the pool.
 /// Only 1 solution with 1 swap is supported.
-pub struct StraightToPoolStrategyEncoder {}
-impl EVMStrategyEncoder for StraightToPoolStrategyEncoder {}
-impl StrategyEncoder for StraightToPoolStrategyEncoder {
-    fn encode_strategy(&self, solution: Solution) -> Result<Vec<u8>, EncodingError> {
+pub struct ExecutorEncoder {}
```
dianacarvalho1: maybe ExecutorStrategyEncoder?

src/encoding/evm/strategy_encoder/encoder.rs
```diff
@@ -68,10 +70,87 @@ impl StrategyEncoder for StraightToPoolStrategyEncoder {
...
         let protocol_data = swap_encoder.encode_swap(swap.clone(), encoding_context)?;
-        // TODO: here we need to pass also the address of the executor to be used
-        Ok(protocol_data)
+        let executor_address = Address::from_str(swap_encoder.executor_address())
+            .map_err(|_| EncodingError::FatalError("Invalid executor address".to_string()))?;
+        Ok((protocol_data, executor_address))
     }
     fn selector(&self, _exact_out: bool) -> &str {
-        unimplemented!();
+        "swap(uint256, bytes)"
```
dianacarvalho1, line 79: do we need the selector in this case? 🤔

src/encoding/evm/strategy_encoder/encoder.rs
```diff
@@ -68,10 +70,87 @@ impl StrategyEncoder for StraightToPoolStrategyEncoder {
...
+            router_address: Some(Bytes::zero(20)),
+            slippage: None,
+            native_action: None,
+        };
+
+        let (protocol_data, executor_address) = encoder
+            .encode_strategy(solution)
+            .unwrap();
+        let hex_protocol_data = encode(&protocol_data);
+        println!("{}", hex_protocol_data);
```
dianacarvalho1: lost print 👀

src/encoding/evm/strategy_encoder/encoder.rs
```diff
@@ -68,10 +70,87 @@ impl StrategyEncoder for StraightToPoolStrategyEncoder {
...
+
+        let solution = Solution {
+            exact_out: false,
+            given_token: token_in,
+            given_amount: BigUint::from(1000000000000000000u64),
+            expected_amount: BigUint::from(1000000000000000000u64),
+            checked_token: token_out,
+            check_amount: None,
+            sender: Bytes::from_str("0x0000000000000000000000000000000000000000").unwrap(),
+            receiver: Bytes::from_str("0x1d96f2f6bef1202e4ce1ff6dad0c2cb002861d3e").unwrap(),
```
dianacarvalho1, line 119: this corresponds to Bob in the solidity test?

foundry/test/executors/UniswapV2Executor.t.sol
```diff
@@ -93,4 +93,31 @@ contract UniswapV2ExecutorTest is UniswapV2ExecutorExposed, Test, Constants {
         uint256 finalBalance = DAI.balanceOf(BOB);
         assertGe(finalBalance, amountOut);
     }
+
+    function testSwapExecutorEncoderData() public {
+        // Generated by the strategy encoder
```
dianacarvalho1: can you mention here which test created this data please?

dianacarvalho1 (commented):
Thank you @royvardhan ! 🙏🏼 It is looking gooood! Happy that the test works! I have some comments

foundry/src/executors/UniswapV2Executor.sol
royvardhan: In the exactOut situation which is there in the `encode_swap` return [here](https://github.com/propeller-heads/tycho-execution/blob/09b2d8b325776bd3412ac34dc2fd9f6d9e4ba30a/src/encoding/evm/strategy_encoder/encoder.rs#L145)

src/encoding/evm/strategy_encoder/encoder.rs
royvardhan: Haha! Exactly I was thinking 😂

src/encoding/evm/strategy_encoder/encoder.rs
royvardhan: Need to do voting. @tamaralipows

src/encoding/evm/strategy_encoder/encoder.rs
royvardhan: Yes, I got the address from foundry tests first.

foundry/test/executors/UniswapV2Executor.t.sol
royvardhan: Yes, good idea!

src/encoding/evm/strategy_encoder/encoder.rs
royvardhan: Jk, I am good with both.

src/encoding/strategy_encoder.rs
royvardhan: Thanks for clarifying!

foundry/src/executors/UniswapV2Executor.sol
dianacarvalho1: won't it just be 1? but the number of bytes will be the same no?

src/encoding/evm/strategy_encoder/encoder.rs
dianacarvalho1: can you write a comment here that mentions that please?

foundry/src/executors/UniswapV2Executor.sol
royvardhan: I console logged number of bytes and it was 62, one extra because of additional bool.

src/encoding/evm/strategy_encoder/encoder.rs
royvardhan: I am keeping the selector for now, maybe we decide later.

dianacarvalho1: I don't think we will need it, but ok

foundry/src/executors/UniswapV2Executor.sol
dianacarvalho1: I'm confused.. shouldn't the length be the same regardless if it's true or false??

src/encoding/evm/strategy_encoder/encoder.rs
```diff
@@ -68,10 +71,87 @@ impl StrategyEncoder for StraightToPoolStrategyEncoder {
...
+mod tests {
+    use alloy::hex::encode;
+    use num_bigint::BigUint;
+    use tycho_core::{dto::ProtocolComponent, Bytes};
+
+    use super::*;
+    use crate::encoding::models::Swap;
+
+    #[test]
+    fn test_executor_encoder() {
```
dianacarvalho1: test_executor_strategy_encode

dianacarvalho1 (approved):
Thank you @royvardhan ! I think it looks good! Only have a small comment and I'm confused about the encoded data length

foundry/src/executors/UniswapV2Executor.sol
royvardhan: So [_decodeData](https://github.com/propeller-heads/tycho-execution/blob/7ca803be0160633bfc2d0d90ba98077c7b84b2fc/foundry/src/executors/UniswapV2Executor.sol#L40-L43) need only 61 bytes because it returns `(address, address, address, bool)` = 61 and it is strictly checking that, i.e, if `data.length != 61`. But the encoder returns an additional bool with the `exact_out` bool, which makes it 62. So I had to change the if check, otherwise it was reverting. 

Hopefully I was able to understand your question and explain myself.

dianacarvalho1: but then why not make the check `data.length != 62`? 🤔

royvardhan: Because we might change the encoder in the future and remove the `exact_out`? Havn't thought about this much, but can change to `data.length != 62` if you give green signal.

dianacarvalho1: no, I think we will keep the exact out. it's needed for the uniswap v2 swap anyway

src/encoding/models.rs
```diff
@@ -11,7 +11,7 @@ pub struct Solution {
     /// Amount of the given token.
     pub given_amount: BigUint,
     /// The token being bought (exact in) or sold (exact out).
-    checked_token: Bytes,
+    pub checked_token: Bytes,
```
dianacarvalho1, line 14: @royvardhan should we rename the `straight_to_pool` flag further down?

foundry/src/executors/UniswapV2Executor.sol
royvardhan: removed exact_out after team convo.

src/encoding/models.rs
royvardhan: Renamed to `direct_execution`, thanks!

tamaralipows (approved):
Thank you!

PR #34, opened 2025-01-30
feat: UniswapV3 Executor

tamaralipows:


foundry/src/TychoRouter.sol
```diff
@@ -369,8 +369,8 @@ contract TychoRouter is
         int256 amount0Delta,
         int256 amount1Delta,
         bytes calldata data
-    ) internal view returns (uint256 amountOwed, address tokenOwed) {
-        address tokenIn = address(bytes20(data[0:20]));
+    ) internal view returns (uint256 amountOwed, address tokenIn) {
```
dianacarvalho1: why did you rename only the tokenIn and not the amountOwned? can we have consistency pleease?

foundry/test/TychoRouter.t.sol
```diff
@@ -606,4 +606,52 @@ contract TychoRouterTest is TychoRouterTestSetup {
 
         vm.stopPrank();
     }
+
+    function testUSV3Callback() public {
```
dianacarvalho1, line 610: wooop wooppp veery niceeeee

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -20,7 +21,7 @@ contract TychoRouterExposed is TychoRouter {
         return _unwrapETH(amount);
     }
 
-    function ExposedSwap(
+    function exposedSwap(
```
dianacarvalho1: I thought I renamed this in my previous PR 🤔

src/encoding/models.rs
```diff
@@ -65,6 +65,7 @@ pub struct Transaction {
 #[allow(dead_code)]
 pub struct EncodingContext {
     pub receiver: Bytes,
+    // TODO should we keep this?
```
dianacarvalho1: I think yes.. because removing it and adding it later would be an interface change..

dianacarvalho1: or we could remove it and if we add it later we make it default to true? that's actually ok I think

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 looks goooood! only have tiny tiny comments!

src/encoding/evm/swap_encoder/encoders.rs
```diff
@@ -113,7 +113,6 @@ impl SwapEncoder for UniswapV3SwapEncoder {
             bytes_to_address(&encoding_context.receiver)?,
             component_id,
             zero_to_one,
-            encoding_context.exact_out,
```
royvardhan, line 116: @dianacarvalho1 Tamara has removed this already, nice!

royvardhan (approved):
Looks good to me! Thank you!
Just one question, why isn't there a swap test on the` UniswapV3Executor.t.sol`? I noticed the USV3 swap tests on the router.t.sol though.

src/encoding/evm/swap_encoder/encoders.rs
dianacarvalho1: this is for uniswap v3, not v2

dianacarvalho1: this is for uniswap v3, not v2

royvardhan: oh yeah!

foundry/src/TychoRouter.sol
tamaralipows: did it to make u mad

src/encoding/models.rs
tamaralipows: No I think you made a good point about the interface change. And we also need to raise an error if someone is requesting exact out

tamaralipows:
@royvardhan I didn't put the swap test on the UniswapV3Executor.t.sol because USV3 performs a callback and the executor doesn't have the callback interface to handle that, so it would revert every time - only our main router has the necessary entrypoint

foundry/src/TychoRouter.sol
dianacarvalho1: ahahhahahah nooooo

royvardhan:
> @royvardhan I didn't put the swap test on the UniswapV3Executor.t.sol because USV3 performs a callback and the executor doesn't have the callback interface to handle that, so it would revert every time - only our main router has the necessary entrypoint

Thanks for clarifying!

PR #35, opened 2025-01-30
feat: Implement Split Swap strategy encoder

dianacarvalho1:
The strategy works as follows:
- Manage approvals needed
- Compute min amount (if check amount is any):
  - if slippage is defined, apply slippage on the expected amount and take the min value between that and the check amount
  - if not, it's just the check amount
- Iterate through the swaps
  - call the corresponding swap encoder to encode the swap
  - add swap header (tokens indexes and split)
  - ple encode the swaps
- Add extra inputs (amounts, token addresses, min amount, (un)wrap, number of tokens and receiver)

Also done in this PR:
- Remove generalisation on user approvals manager


src/encoding/evm/approvals/permit2.rs
```diff
@@ -1,15 +1,17 @@
...
 use alloy::{
-    primitives::{aliases::U48, Address, Bytes as AlloyBytes, TxKind, U160},
+    primitives::{aliases::U48, Address, Bytes as AlloyBytes, ChainId, TxKind, U160, U256},
     providers::{Provider, RootProvider},
     rpc::types::{TransactionInput, TransactionRequest},
     signers::{local::PrivateKeySigner, SignerSync},
     transports::BoxTransport,
 };
-use alloy_primitives::{ChainId, U256};
+#[allow(deprecated)]
```
dianacarvalho1, line 10: so.. craziest thing here.. Alloy tells me that Signature is deprecated and to use PrimitiveSignature instead. sure. but then further down we use `sign_hash_sync` also from alloy and it returns the deprecated Signature. It seems like the deprecated something without updating their own libraries 🫠

src/encoding/evm/strategy_encoder/encoder.rs
```diff
@@ -75,3 +194,111 @@ impl StrategyEncoder for StraightToPoolStrategyEncoder {
...
+        let expected_swaps = String::from(concat!(
+            // ple encode adds aaalll of this :/ is it correct?
+            "0000000000000000000000000000000000000000000000000000000000000000",
+            "0000000000000000000000000000000000000000000000000000000000120000",
+            "0000000000000000000000000000000000000000000000000000000000020000",
+            "0000000000000000000000000000000000000000000000000000000000060000",
+            "000000000000000000000000000000000000000000000000000000000005b000",
+            "0000000000000000000000000000000000000000000000000000000000080000",
+            "0000000000000000000000000000000000000000000000000000000000000000",
+            "000000000000000000000000000000000000000000000000000000000005b",
```
dianacarvalho1: @tamaralipows does this look correct to you?? it seems tooo many lines for me 😕 but I dunno what else to change.. I need to do an integration test to verify this

src/encoding/evm/strategy_encoder/selector.rs
```diff
@@ -7,11 +11,21 @@ use crate::encoding::{
...
-            Box::new(StraightToPoolStrategyEncoder {})
+            Ok(Box::new(StraightToPoolStrategyEncoder {}))
         } else {
-            Box::new(SplitSwapStrategyEncoder {})
+            let signer = signer.ok_or_else(|| {
+                EncodingError::FatalError(
+                    "Signer is required for SplitSwapStrategyEncoder".to_string(),
+                )
+            })?;
+            Ok(Box::new(SplitSwapStrategyEncoder::new(signer, chain_id).unwrap()))
```
dianacarvalho1: I'm not sure if I like this.. overall I don't like how we are initing our encoders (strategy and swap one).
here for strategy we will be initing one every time the user wants to encode a new swap-
for the swap ones we have a registry that gets created on initialisation.. 
do you guys have ideas on how we could make this better?

src/encoding/evm/utils.rs
```diff
@@ -56,3 +56,19 @@ pub fn encode_input(selector: &str, mut encoded_args: Vec<u8>) -> Vec<u8> {
...
+    let scaled = (percentage / 100.0) * (MAX_UINT24 as f64);
+    U24::from(scaled.round())
+}
+
+pub fn to_chain_id(chain: Chain) -> Result<u64, EncodingError> {
+    match chain {
+        Chain::Ethereum => Ok(1),
+        _ => Err(EncodingError::FatalError("Unsupported chain".to_string())),
+    }
+}
```
dianacarvalho1: Ideally this should be inside tycho-core. Going to raise a discussion about this in the channel

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+    permit2: Permit2,
+}
+
+impl SplitSwapStrategyEncoder {
+    pub fn new(signer_pk: String, chain: Chain) -> Result<Self, EncodingError> {
+        Ok(Self { permit2: Permit2::new(signer_pk, chain)? })
+    }
+}
+impl EVMStrategyEncoder for SplitSwapStrategyEncoder {}
+impl StrategyEncoder for SplitSwapStrategyEncoder {
```
dianacarvalho1, line 68: I know we had a [task](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-4083) to add validations to the router but now I'm actually thinking that the strategy encoders should be responsible for having their own validations. So here I would say we are missing the following validations:
- Check if all the tokens are connected
- Check if all the splits sum to 1

dianacarvalho1: but I would prefer to do this in a separate task please 😬

src/encoding/evm/utils.rs
dianacarvalho1: ok, no. Alan says it should live here. I will improve this then!

src/encoding/evm/approvals/permit2.rs
tamaralipows: oh god 💀

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+            .unwrap();
+
+        let expected_input = String::from(concat!(
+            "0000000000000000000000000000000000000000000000000000000000000020", // offset
+            "0000000000000000000000000000000000000000000000000de0b6b3a7640000", // amount out
+            "000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2", // token in
+            "0000000000000000000000006b175474e89094c44da98b954eedeac495271d0f", // token out
+            "0000000000000000000000000000000000000000000000000000000000000000", // min amount out
+            "0000000000000000000000000000000000000000000000000000000000000000", // wrap
+            "0000000000000000000000000000000000000000000000000000000000000000", // unwrap
```
tamaralipows, line 344: oh god. We just want a bool input and it pads to 32 bytes :( so sad

tamaralipows: Now I understand why we encoded things in the protocolData and decoded them in the method instead of just passing them into the highest level

tamaralipows: it's a good thing we don't care about gas <3

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+        let expected_swaps = String::from(concat!(
+            // ple encode adds aaalll of this :/ is it correct?
+            "0000000000000000000000000000000000000000000000000000000000000000",
+            "0000000000000000000000000000000000000000000000000000000000120000",
+            "0000000000000000000000000000000000000000000000000000000000020000",
+            "0000000000000000000000000000000000000000000000000000000000060000",
+            "000000000000000000000000000000000000000000000000000000000005b000",
+            "0000000000000000000000000000000000000000000000000000000000080000",
+            "0000000000000000000000000000000000000000000000000000000000000000",
+            "000000000000000000000000000000000000000000000000000000000005b",
```
tamaralipows: no this looks sus - something must be wrong with the ple encoding method

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Yes sure! That makes sense!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: should we just encode all of this into bytes and decode it first thing in the method? 😕 it makes it a bit sketchier but we would save quite a lot of gas I think

tamaralipows: Do you really think it would be a lot of gas? These are mostly zeros anyway

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: fixed iiittt ✨

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: true.. we can keep this in mind for a possible future optimization

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: @royvardhan do you think you could take these validations as your next task please?

src/encoding/evm/approvals/permit2.rs
royvardhan: Shall we open pr in alloy to fix this?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Yes I can do that.

src/encoding/evm/router_encoder.rs
```diff
@@ -1,51 +1,61 @@
...
-            let strategy = self
-                .strategy_selector
-                .select_strategy(solution);
-            let method_calldata = strategy.encode_strategy((*solution).clone())?;
+            let router_address = solution
+                .router_address
+                .clone()
+                .unwrap_or(Bytes::from_str(&self.router_address).map_err(|_| {
+                    EncodingError::FatalError("Invalid router address".to_string())
+                })?);
```
tamaralipows: Can't we just initially store our router address as bytes so that we don't need to do this at every `encode`?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+        router_address: Bytes,
+    ) -> Result<Vec<u8>, EncodingError> {
+        let (permit, signature) = self.permit2.get_permit(
+            &router_address,
+            &solution.sender,
+            &solution.given_token,
+            &solution.given_amount,
+        )?;
+        let min_amount_out = if solution.check_amount.is_some() {
+            let mut min_amount_out = solution.check_amount.clone().unwrap();
```
tamaralipows: I'm surprised we can just reuse the same variable names like that in rust. I saw this and instantly got nervous

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+            if solution.slippage.is_some() {
+                let one_hundred = BigUint::from(100u32);
+                let slippage_percent = BigUint::from((solution.slippage.unwrap() * 100.0) as u32);
+                let multiplier = &one_hundred - slippage_percent;
+                let expected_amount_with_slippage =
+                    (&solution.expected_amount * multiplier) / one_hundred;
+                min_amount_out = min(min_amount_out, expected_amount_with_slippage);
```
tamaralipows: I think this is wrong, no? I think it should be the max not the min.

That's how we have been doing it in our old routers

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+            };
+            let protocol_data = swap_encoder.encode_swap(swap.clone(), encoding_context)?;
+            let swap_data = self.encode_swap_header(
+                U8::from(
+                    tokens
+                        .iter()
+                        .position(|t| *t == swap.token_in)
+                        .ok_or_else(|| {
+                            EncodingError::InvalidInput(
+                                "Token in not found in tokens array".to_string(),
```
tamaralipows: I know what this means but it sounds weird lol
maybe like this?
```suggestion
                                "In token not found in tokens array".to_string(),
```

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+                            )
+                        })?,
+                ),
+                U8::from(
+                    tokens
+                        .iter()
+                        .position(|t| *t == swap.token_out)
+                        .ok_or_else(|| {
+                            EncodingError::InvalidInput(
+                                "Token out not found in tokens array".to_string(),
```
tamaralipows: same

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,304 @@
...
+                percentage_to_uint24(swap.split),
+                protocol_data,
+            );
+            swaps.push(swap_data);
+        }
+
+        let encoded_swaps = ple_encode(swaps);
+        let (mut unwrap, mut wrap) = (false, false);
+        if solution.native_action.is_some() {
+            match solution.native_action.unwrap() {
```
tamaralipows: I almost missed this because of all the wraps and unwraps around lol
But pls no unwrap in prod code

src/encoding/evm/strategy_encoder/selector.rs
tamaralipows: Could we not just pass the signer in the solution? Didn't look too deep yet though

tamaralipows (commented):
Thank you!! Some comments 4u

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: ahahah yes! it wooorks. different scopes

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: whyyyyy? I already checked that it is some.. so it should be fine to unwrwap no?

src/encoding/evm/strategy_encoder/selector.rs
dianacarvalho1: the signer is literally a private key. 😕 I think it's too sensitive and more of a part of the setup to put in the solution

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: yeah.. we could do that

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: I think the pattern you are using is bad
```
if let Some(action) = solution.native_action {
    // Safe to use `action` here
}
```

This is the recommended way to proceed. I can change it

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,385 @@
...
+        &self,
+        solution: Solution,
+        _router_address: Bytes,
+    ) -> Result<(Vec<u8>, Bytes), EncodingError> {
+        if solution.router_address.is_none() {
+            return Err(EncodingError::InvalidInput(
+                "Router address is required for straight to pool solutions".to_string(),
+            ));
+        }
+        let swap = solution.swaps.first().unwrap();
```
tamaralipows: There are actually a lot of unwraps :( please be more careful about that. For example here this will fail if the user passes an empty solution (bad input)

> If you call .unwrap() on a None value, your program panics and crashes.

src/encoding/evm/strategy_encoder/strategy_selector.rs
```diff
@@ -0,0 +1,30 @@
...
+    ) -> Result<Box<dyn StrategyEncoder>, EncodingError> {
+        if solution.direct_execution {
+            Ok(Box::new(ExecutorStrategyEncoder {}))
+        } else {
+            let signer_pk = signer.ok_or_else(|| {
+                EncodingError::FatalError(
+                    "Signer is required for SplitSwapStrategyEncoder".to_string(),
+                )
+            })?;
+            Ok(Box::new(SplitSwapStrategyEncoder::new(signer_pk, chain).unwrap()))
```
tamaralipows: Here too

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -0,0 +1,385 @@
...
+            })?;
+        let swap_encoder = registry
+            .get_encoder(&swap.component.protocol_system)
+            .ok_or_else(|| {
+                EncodingError::InvalidInput(format!(
+                    "Swap encoder not found for protocol: {}",
+                    swap.component.protocol_system
+                ))
+            })?;
+        let router_address = solution.router_address.unwrap();
```
tamaralipows: and here

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Ah I think this was just a part of the initial sketch and @royvardhan forgot to update this in his PR and I didn't catch it (and you just moved it around in your PR I guess? the diff is really confusing).

FYI for @royvardhan : unwraps in production code / libraries are bad because they will crash the client's whole system when it could be the issue of just one faulty solution or protocol. Please see more information here:

https://propellerswap.slack.com/archives/GB59UTQ2K/p1737392914620059

tamaralipows (approved):
I replaced all the unwraps I could find with proper error handling. I'm gonna branch from this, and I'll fix more if I find them in the next PR

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: yes, the diff got crazy. and yes, it was in the initial sketch and was not updated 😕

src/encoding/evm/approvals/permit2.rs
royvardhan: But it returns [`PrimitiveSignature as Signature`](https://github.com/alloy-rs/alloy/blob/684dc757188f4cd757a9a19cd4e5b4eeb45c0d7b/crates/signer/src/signer.rs#L3)

PR #37, opened 2025-01-30
feat: replace setExecutor with batchSetExecutor

royvardhan:
- Replaced setExecutor with batchSetExecutor
- Added relevant tests


foundry/test/TychoRouter.t.sol
```diff
@@ -20,15 +20,32 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        address[] memory executors = new address[](1);
+        executors[0] = DUMMY;
         vm.startPrank(EXECUTOR_SETTER);
-        tychoRouter.setExecutor(DUMMY);
+        tychoRouter.batchSetExecutor(executors);
         vm.stopPrank();
         assert(tychoRouter.executors(DUMMY) == true);
     }
 
+    function testSetExecutorMultipleValidRole() public {
```
dianacarvalho1: the method is called `batchSetExecutor`, this test is called `testSetExecutorMultiple` 🥴 can we have coherence please? Maybe I would even simplify the method name and call it `setExecutors`. But I don't have a strong preference here. Will leave it up to you!

dianacarvalho1 (approved):
Thank you @royvardhan ! 🙏🏼 It looks good! I only have a minor comment

foundry/test/TychoRouter.t.sol
royvardhan: My mistake, I didn't look to change the test name. I had initially change the name to `setExecutors` but changed it to batch because the term `batch` is being used by other setters, ex: `batchGrantRole`, to keep it same.

royvardhan: Have changed the naming.

tamaralipows: And the name of this test pls

```suggestion
    function testSetExecutorsValidRole() public {
```

royvardhan: @tamaralipows Same test name exists, so I removed it and merged both.

tamaralipows (approved):
tank yew

PR #38, opened 2025-01-30
test: Router integration tests

tamaralipows:
Diana:
Test passeeess ✨ 
- The sender needs to be Alice. The tokens belong to Alice and she needs to be the one to call the router.
- The block we test in actually doesn't matter 🤔 curious behaviour: it only needs to be before the block where we make the permit. In permit2 code they only check that the current block timestamp < expiration/deadline. So our block is waaayyy behind our deadlines. This is cool for testing and it will never be an issue in production code because you can't go back to the past and run something on an older block.


dianacarvalho1 (approved):
awesooomeeeeeee 🚀

foundry/test/TychoRouter.t.sol
```diff
@@ -671,4 +671,30 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        deal(WETH_ADDR, ALICE, 1 ether);
+        uint256 balancerBefore = IERC20(DAI_ADDR).balanceOf(ALICE);
+
+        // Approve permit2
+        vm.startPrank(ALICE);
+        IERC20(WETH_ADDR).approve(address(permit2Address), type(uint256).max);
+        // Encoded solution generated using `test_split_swap_strategy_encoder` but
+        // manually replacing the executor address `5c2f5a71f67c01775180adc06909288b4c329308`
+        // with the one in this test `5615deb798bb3e4dfa0139dfa1b3d433cc23b72f`
+        (bool success,) = tychoRouterAddr.call(
```
royvardhan, line 689: Amazing stuff!

PR #41, opened 2025-01-31
feat: Router split swap integration test

tamaralipows:


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -400,4 +402,88 @@ mod tests {
...
+
+        let swap_weth_dai = Swap {
+            component: ProtocolComponent {
+                id: "0xA478c2975Ab1Ea89e8196811F51A7B7Ade33eB11".to_string(),
+                protocol_system: "uniswap_v2".to_string(),
+                ..Default::default()
+            },
+            token_in: weth.clone(),
+            token_out: dai.clone(),
+            split: 0.5f64,
```
dianacarvalho1, line 435: Ooohh im a little bit concerned about this. I assumed this would be a percentage! Not a ratio type of thingy

foundry/test/TychoRouter.t.sol
```diff
@@ -695,6 +696,37 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        );
+
+        vm.stopPrank();
+
+        uint256 balancerAfter = IERC20(USDC_ADDR).balanceOf(ALICE);
+
+        assertTrue(success, "Call Failed");
+        assertGe(balancerAfter - balancerBefore, 26173932);
```
dianacarvalho1, line 730: Can you assert that all the WETH was spent please? I think the splits were not encoded nicely

dianacarvalho1 (commented):
Thaaank you!! Amazing that this just works ✨ but I'm concerned about the splits. Can you double check please?

dianacarvalho1 (approved):
yeeeesss amazzingg ✨

PR #43, opened 2025-01-31
feat: Tycho encoder validation

tamaralipows:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4201)

Validates:
- Proper sequence of input, output, and first/last swap tokens for wrap/unwrap cases
- All solutions contain at least one swap
- Only exact in solutions are inputted (since we don't yet support exact out)


src/encoding/evm/tycho_encoder.rs
```diff
@@ -70,15 +67,62 @@ impl<S: StrategySelector> TychoEncoder<S> for EVMTychoEncoder<S> {
...
+                        return Err(EncodingError::FatalError(
+                            "WETH must be the first swap's input in order to wrap".to_string(),
+                        ));
+                    }
+                }
+                if solution.given_token != *NATIVE_ADDRESS {
+                    return Err(EncodingError::FatalError(
+                        "ETH must be the input token in order to wrap".to_string(),
+                    ));
+                }
```
dianacarvalho1: can we just change the order of these two checks? first the solution token and then the swap token? 
I think it makes more sense when you read the function

dianacarvalho1: I got confused when reading it like this and had to do a double take 😅

src/encoding/tycho_encoder.rs
```diff
@@ -10,4 +10,6 @@ pub trait TychoEncoder<S: StrategySelector> {
         &self,
         solutions: Vec<Solution>,
     ) -> Result<Vec<Transaction>, EncodingError>;
+
+    fn validate_solution(&self, solution: &Solution) -> Result<(), EncodingError>;
```
dianacarvalho1: I think we don't need this in the interface..

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 It looks gooood! only have 1 request!

src/encoding/evm/tycho_encoder.rs
royvardhan: Same here, I was wondering that `given_token` and` first swap.token_in` should be the same. No confusion now.

src/encoding/evm/tycho_encoder.rs
```diff
@@ -70,15 +67,62 @@ impl<S: StrategySelector> TychoEncoder<S> for EVMTychoEncoder<S> {
...
 mod tests {
+    use tycho_core::dto::ProtocolComponent;
+
     use super::*;
-    use crate::encoding::strategy_encoder::StrategyEncoder;
+    use crate::encoding::{models::Swap, strategy_encoder::StrategyEncoder};
 
     struct MockStrategySelector;
 
+    fn dai() -> Bytes {
```
royvardhan, line 123: maybe add the network? like `dai_mainnet`?

src/encoding/tycho_encoder.rs
royvardhan: I have added it outside of the trait block, just as a helper function on my SplitSwap pr, does that work?

royvardhan (approved):

src/encoding/evm/tycho_encoder.rs
tamaralipows: this is in the evm dir so I think it's ok! or in which case would this not be okay?
it's just for testing anyway, the bigger problem would be our constants

src/encoding/tycho_encoder.rs
tamaralipows: changed :)

PR #44, opened 2025-02-03
feat: Verify that no amount in is left in the router.

dianacarvalho1:
Verify that no amount in is left in the router. This came after reading the previous audit one more time! This way we assume a zero trust architecture and make sure that the encoded swaps make sense and do what we expect them to do.


tamaralipows (approved):
This is great! Thanks a lot!

royvardhan (approved):

PR #45, opened 2025-02-04
fix: Fix bug with token indexing when wrapping/unwrapping

tamaralipows:
- Found by adding integration tests for the wrapping and unwrapping cases.


tamaralipows:
Wrong branch naming

PR #46, opened 2025-02-04
fix: Fix bug with token indexing when wrapping/unwrapping

tamaralipows:
The issue was that we weren't indexing WETH properly since it was looking for the WETH address in tokens, when only native ETH would be in there

- Found by adding integration tests for the wrapping and unwrapping cases.


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -108,10 +109,27 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
+            tokens.push(solution.given_token.clone());
+        }
         tokens.extend(intermediary_tokens);
-        tokens.push(solution.checked_token.clone());
+
+        if unwrap {
+            tokens.push(WETH_ADDRESS.clone());
+        } else {
+            tokens.push(solution.checked_token.clone());
+        }
```
tamaralipows, line 133: important part

dianacarvalho1: aaahhh great catch!! 👏🏼

dianacarvalho1 (approved):
veeeerryy gooooodd ✨ ✨

royvardhan (approved):

PR #47, opened 2025-02-04
feat: Refactor Registries

dianacarvalho1:
Interface changes:
- Rename StrategySelector to StrategyEncoderRegistry
- Implement clone for SwapEncoder

The StrategyEncoderRegistry needs to be initialised at the highest level and the passed to the TychoEncoder. The TychoEncoder doesn't hold the chain nor the signer_pk as attributes anymore The StrategyEncoderRegistry does:
- Initialises the SwapEncoderRegistry
- Initialises all the strategies and saves them in a HashMap
- Later, the TychoEncoder only reads from this hashmap

The StrategyEncoder now each holds a SwapEncoderRegistry as an attribute and they use this to get the correct SwapEncoder instead of reading from the global SWAP_ENCODER_REGISTRY

Simplified the SwapEncoderRegistry to not need a Config (everything is done inside itself) All SwapEncoders implement clone


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -251,9 +268,15 @@ mod tests {
...
+    fn get_swap_encoder_registry() -> SwapEncoderRegistry {
+        SwapEncoderRegistry::new("src/encoding/config/executor_addresses.json", Chain::Ethereum)
+            .unwrap()
+    }
+
     #[test]
     fn test_executor_strategy_encode() {
-        let encoder = ExecutorStrategyEncoder {};
+        let swap_encoder_registry = get_swap_encoder_registry();
+        let encoder = ExecutorStrategyEncoder::new(swap_encoder_registry);
```
royvardhan, line 293: Noice! So that means anyone can have their own addresses list and pass it here.

royvardhan (approved):
Nothing to add from my side. Looks great!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: exactly!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -116,16 +122,14 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
 
         let mut swaps = vec![];
         for swap in solution.swaps.iter() {
-            let registry = SWAP_ENCODER_REGISTRY
-                .read()
-                .map_err(|_| {
-                    EncodingError::FatalError(
-                        "Failed to read the swap encoder registry".to_string(),
-                    )
```
tamaralipows, line 142: Ahhhhhh we get rid of this terrible thing that can happen at runtime

tamaralipows: This is so beautiful

src/encoding/evm/strategy_encoder/strategy_encoder_registry.rs
```diff
@@ -0,0 +1,53 @@
+use std::collections::HashMap;
```
tamaralipows, line 1: I have one pet peeve. This file is called `strategy_encoder_registry.rs`, but the swap registry is just registry.rs. Maybe we should also just call this `registry.rs`?

src/encoding/evm/strategy_encoder/strategy_encoder_registry.rs
```diff
@@ -0,0 +1,53 @@
...
+        chain: Chain,
+        executors_file_path: &str,
+        signer_pk: Option<String>,
+    ) -> Result<Self, EncodingError> {
+        let swap_encoder_registry = SwapEncoderRegistry::new(executors_file_path, Chain::Ethereum)?;
+
+        let mut strategies: HashMap<String, Box<dyn StrategyEncoder>> = HashMap::new();
+        strategies.insert(
+            "executor".to_string(),
+            Box::new(ExecutorStrategyEncoder::new(swap_encoder_registry.clone())),
```
tamaralipows, line 30: it's not possible just to pass the ref here? We really need to clone a separate one for each strategy?

tamaralipows (approved):
This took forever to understand but I can finally say... me likey 

just a single question and one naming thing

src/encoding/evm/strategy_encoder/strategy_encoder_registry.rs
dianacarvalho1: I prefer to have them both with the big names. I did that mistake with the encoders but then when I had all the files open in my IDE it was soooo confusing

dianacarvalho1: will change the other one!

tamaralipows: true! good choice

src/encoding/evm/strategy_encoder/strategy_encoder_registry.rs
dianacarvalho1: I quickly tried it and got a lot of lifetime issues 😬 let's not go through that hole now please xD

PR #48, opened 2025-02-04
feat: add validations for split swap

royvardhan:
[Issue](https://datarevenue.atlassian.net/browse/ENG-4083)


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -191,6 +201,144 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
     }
 }
 
+fn validate_swaps(swaps: &[Swap]) -> Result<(), EncodingError> {
```
dianacarvalho1: I think the methods could be a part of the SplitSwapStrategyEncoder implementation only

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -191,6 +201,144 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
+    }
+
+    let mut swaps_by_token: HashMap<Bytes, Vec<&Swap>> = HashMap::new();
+
+    for swap in swaps {
+        swaps_by_token
+            .entry(swap.token_in.clone())
+            .or_default()
+            .push(swap);
+    }
```
dianacarvalho1: could we collapse these two for loops into one please?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -191,6 +201,144 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
+
+    for (token, token_swaps) in swaps_by_token {
+        if token_swaps.is_empty() {
+            return Err(EncodingError::InvalidInput(format!(
+                "No swaps found for token {:?}",
+                token
+            )));
+        }
```
dianacarvalho1: in which case could this even happen? 🤔

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -191,6 +201,144 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
+                (true, false) => {
+                    return Err(EncodingError::InvalidInput(format!(
+                        "The 0% split for token {:?} must be the last swap",
+                        token
+                    )))
+                }
+                (true, true) => found_zero_split = true,
+                (false, _) => (),
+            }
+        }
```
dianacarvalho1: instead of looping, couldn't we just check the last swap and that's it?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -191,6 +201,144 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
+                token,
+                total_percentage * 100.0
+            )));
+        }
+    }
+
+    Ok(())
+}
+
+fn validate_token_path_connectivity(
```
dianacarvalho1: this method will fail for the case when we need to wrap ETH. the given token is ETH but the token in of the first swap is WETH

dianacarvalho1: and for the unwrap case as well

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -489,4 +637,275 @@ mod tests {
...
+                split: 0.0,
+            },
+        ];
+        let result = validate_token_path_connectivity(&disconnected_swaps, &weth, &usdc);
+        assert!(matches!(
+            result,
+            Err(EncodingError::InvalidInput(msg)) if msg.contains("not reachable through swap path")
+        ));
+
+        // Test case 2: Unreachable checked token
```
dianacarvalho1: could we parameterize the test using case?? instead of having this. If not, I would prefer to break this test into two tests

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -489,4 +637,275 @@ mod tests {
...
+                token_out: dai.clone(),
+                split: 0.3,
+            },
+        ];
+        assert!(matches!(
+            validate_swaps(&invalid_total_swaps),
+            Err(EncodingError::InvalidInput(msg)) if msg.contains("must have exactly one 0% split")
+        ));
+
+        // Test case 2: Invalid - zero split not at end
```
dianacarvalho1: same here

dianacarvalho1 (commented):
Thank you @royvardhan ! I have some questions..
also, validate_swaps and validate_token_path_connectivity are not the best names here :/ could we rename them? in the first function you are validating the splits and in the second one you are still validating the swaps as well, just if the tokens are connected or not

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Bad user input. This is handled correctly imo. They might have a bug in their solver and in that case we don't want to crash their whole system with a panic - just let them know

**EDIT: I was wrong.** I don't see this happening either

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: If more than one swap has a 0 split it can also be a problem though, I think that is what's tackled here

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: we should probably add test cases for those too

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -191,6 +201,144 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
...
+    // Build directed graph of token flows
+    let mut graph: HashMap<&Bytes, HashSet<&Bytes>> = HashMap::new();
+    for swap in swaps {
+        graph
+            .entry(&swap.token_in)
+            .or_default()
+            .insert(&swap.token_out);
+    }
+
+    // BFS from given_token
```
tamaralipows: What is BFS?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Agreed - in general it's better to keep different test cases separate even if they're testing the same method

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: but how? we are looping by swaps to create `swaps_by_token`.. so I think it will never be empty no?

tamaralipows (commented):
On the right track - thanks! I agree with a lot of Diana's comments

One thing for future note - we use pure kebab case for the branch names, you seem to be using a mix of upper and lower case. We keep the upper case just for the task names. It's a small thing but makes it feel more organized in the long run

In the case of this PR, that would mean the proper name is `encoder/hr/ENG-4083-validate-split-swap`. No need to change it now but just keep it in mind for next time

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Ahhhhh damn you are right! This is entirely on our side! I misread sorry sorry

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Makes sense.

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Sorry, missed this!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Yes correct, we could check if the last swap has split zero, but then other swaps could also have split zero and that's a problem.

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: breadth first search

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Sure, I have now seperated the tests

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Done!

royvardhan:
@dianacarvalho1 @tamaralipows  Ready for review again!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -63,6 +67,141 @@ impl SplitSwapStrategyEncoder {
...
+            for (i, swap) in token_swaps.iter().enumerate() {
+                match (swap.split == 0.0, i == token_swaps.len() - 1) {
+                    (true, false) => {
+                        return Err(EncodingError::InvalidInput(format!(
+                            "The 0% split for token {:?} must be the last swap",
+                            token
+                        )))
+                    }
+                    (true, true) => found_zero_split = true,
+                    (false, _) => (),
```
tamaralipows: Couldn't we perform the non-zero summing here and avoid a second loop down below? Or do you think it's more organized how it is now?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -63,6 +67,141 @@ impl SplitSwapStrategyEncoder {
...
+            let mut total_percentage = 0.0;
+            for swap in token_swaps
+                .iter()
+                .take(token_swaps.len() - 1)
+            {
+                if swap.split <= 0.0 {
+                    return Err(EncodingError::InvalidInput(format!(
+                        "Non-remainder splits must be >0% for token {:?}",
+                        token
+                    )));
```
tamaralipows: I feel that we are checking this twice

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -63,6 +67,141 @@ impl SplitSwapStrategyEncoder {
...
+    }
+
+    fn validate_swap_path(
+        &self,
+        swaps: &[Swap],
+        given_token: &Bytes,
+        checked_token: &Bytes,
+    ) -> Result<(), EncodingError> {
+        // Special case: If given_token is ETH or checked_token is ETH, treat it as WETH for path
+        // validation
```
tamaralipows: Wait a second, in the case of uniswapv4 we may actually want this to stay as ETH. I think in this case it is important to check the native action for wrapping and unwrapping

tamaralipows (commented):
These are a few big things that stick out to me right away

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Yes that is possible and allows us to skip one more loop

royvardhan: Done!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Very good catch! Thank you, updated.

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Not sure where, could you point out?

tamaralipows: Above

```
            for (i, swap) in token_swaps.iter().enumerate() {
                match (swap.split == 0.0, i == token_swaps.len() - 1) {
                    (true, false) => {
                        return Err(EncodingError::InvalidInput(format!(
                            "The 0% split for token {:?} must be the last swap",
                            token
                        )))
                    }
```

royvardhan: It is checking that for any swap, if the split is zero then it must be only the last index.

But for the one which you mentioned, it is checking if the split is not zero, then it must also not be less than zero.

dianacarvalho1 (approved):
Niceeeee, very good! This looks great! 🙏🏼 Thank you @royvardhan

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -69,6 +73,147 @@ impl SplitSwapStrategyEncoder {
...
+                        return Err(EncodingError::InvalidInput(format!(
+                            "The 0% split for token {:?} must be the last swap",
+                            token
+                        )))
+                    }
+                    (true, true) => found_zero_split = true,
+                    (false, _) => {
+                        if swap.split <= 0.0 {
+                            return Err(EncodingError::InvalidInput(format!(
+                                "Non-remainder splits must be >0% for token {:?}",
```
tamaralipows: In this case I think the message is wrong if you want to check that it's less than zero -
I would just say all splits must be greater than or equal to zero. And it's not just for remainder splits

```suggestion
                        if swap.split < 0.0 {
                            return Err(EncodingError::InvalidInput(format!(
                                "All splits must be >= 0% for token {:?}",
```

tamaralipows (approved):
Thanks!!

PR #49, opened 2025-02-04
feat: add swap encoder test in balancer v2 executor

royvardhan:
- Gets the encoded hex data from the BalancerV2 encoder
- Uses it in the executor


foundry/test/executors/BalancerV2Executor.t.sol
```diff
@@ -82,4 +82,40 @@ contract BalancerV2ExecutorTest is
         assertGt(balanceAfter, balanceBefore);
         assertEq(balanceAfter - balanceBefore, amountOut);
     }
+
+    function testSwapExecutorEncoderData() public {
```
dianacarvalho1: the name of this function doesn't really tell me what it is testing.. the name say too many things and it doesn't mention that it is actually testing the decoding

foundry/test/executors/BalancerV2Executor.t.sol
```diff
@@ -82,4 +82,40 @@ contract BalancerV2ExecutorTest is
...
+        ) = balancerV2Exposed.decodeParams(protocolData);
+
+        assertEq(address(tokenIn), WETH_ADDR);
+        assertEq(address(tokenOut), BAL_ADDR);
+        assertEq(poolId, WETH_BAL_POOL_ID);
+        assertEq(receiver, BOB);
+        assertEq(needsApproval, true);
+    }
+
+    function testSwapExecutorSwap() public {
```
dianacarvalho1: also here.. why the `SwapExecutor` part?

foundry/test/executors/BalancerV2Executor.t.sol
dianacarvalho1: you can just call it `testDecodeIntegration`

foundry/test/executors/BalancerV2Executor.t.sol
dianacarvalho1: maybe `testSwapIntegration`?

dianacarvalho1 (commented):
Thank you @royvardhan ! It looks good! Just have some recommendations on the test naming

royvardhan:
> Thank you @royvardhan ! It looks good! Just have some recommendations on the test naming

I have changed the test naming for both the executors, thanks for reviewing!

dianacarvalho1 (approved):
Thank you

tamaralipows (approved):

PR #50, opened 2025-02-04
fix: use PrimitiveSignature instead of deprecated Signature

royvardhan:
Problem: [Signature is deprecated, use PrimitiveSignature instead](https://github.com/propeller-heads/tycho-execution/pull/35#discussion_r1935467352)
The `alloy-signer-local` has already updated the `sign_hash_sync` return to use [PrimitiveSignature as Signature](https://github.com/alloy-rs/alloy/blob/684dc757188f4cd757a9a19cd4e5b4eeb45c0d7b/crates/signer-local/src/lib.rs#L11). I updated our dep to a newer version(not the latest alloy because of breaking changes).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,9 +47,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "0.5.4"
+version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


dianacarvalho1 (approved):
aaaaahh this was the issue 🤦🏼‍♀️ I didn't realise we were using an old version. Thank you for double checking!

royvardhan:
> aaaaahh this was the issue 🤦🏼‍♀️ I didn't realise we were using an old version. Thank you for double checking! 

No problem! 🙏🏻

PR #52, opened 2025-02-04
feat: Get native/wrapped addresses for chain

tamaralipows:
- These were being hardcoded to ETH and WETH, which may not always be the case for EVM-compatible chains.


royvardhan (approved):
Thank you 🙏🏻

src/encoding/evm/constants.rs
```diff
@@ -1,10 +1,18 @@
...
-use lazy_static::lazy_static;
-use tycho_core::Bytes;
+use tycho_core::{models::Chain, Bytes};
 
-lazy_static! {
-    pub static ref NATIVE_ADDRESS: Bytes =
-        Bytes::from(hex!("0000000000000000000000000000000000000000").to_vec());
-    pub static ref WETH_ADDRESS: Bytes =
-        Bytes::from(hex!("c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2").to_vec());
+pub fn native_address(chain: Chain) -> Bytes {
```
dianacarvalho1: huuuuummmmmmm.. I feel we can do better here. In evm/models we already have a ChainId defined. maybe we could combine all of this that we need for execution under the same struct? like to make our own Chain object that has: id, native_address, wrapped_address as functions or attributes?

src/encoding/evm/constants.rs
```diff
@@ -1,10 +1,18 @@
...
+    match chain {
+        Chain::Ethereum => Bytes::from(hex!("0000000000000000000000000000000000000000").to_vec()),
+        // Placeholder values for other chains; update with real addresses
+        _ => Bytes::from(hex!("0000000000000000000000000000000000000000").to_vec()),
```
dianacarvalho1: I think we should raise if the chain is not supported!

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I have a suggestion though

src/encoding/evm/constants.rs
tamaralipows: I raise at a higher level

src/encoding/evm/constants.rs
tamaralipows: uh ohhhh, another chain object eh. sure sure

tamaralipows:
oh my god this alloy issue in the CI is making me crazy

src/encoding/models.rs
```diff
@@ -68,3 +76,65 @@ pub struct EncodingContext {
...
+        match chain {
+            TychoCoreChain::Ethereum => ChainId(1),
+            TychoCoreChain::ZkSync => ChainId(324),
+            TychoCoreChain::Arbitrum => ChainId(42161),
+            TychoCoreChain::Starknet => ChainId(0),
+        }
+    }
+}
+
+impl Chain {
```
dianacarvalho1, line 121: oh I'm sorry @tamaralipows 😢 this is not what I meant. I was suggesting for us to hardcode the wrapped and native token for each chain depending on the value of chain ID. like

````
impl ChainId {
    pub fn id(&self) -> u64 {
        self.0
    }
  
    pub fn wrapped_token(&self) -> u64 {
        match self {
            ChainId(1) => WETH,
            ChainId(324)  => wtv it is on zksync,
            ....
        }
    pub native_token(&self) {
          ....  
    }
}

````

Comment (dianacarvalho1) [reply src/encoding/models.rs:121]: as you asked in your question.. I find it really strange that the user needs to pass the native and wrapped token themself 😕

Comment (dianacarvalho1) [reply src/encoding/models.rs:121]: I'm sorry I wasn't clearer 😞

Comment (tamaralipows) [reply src/encoding/models.rs:121]: What, I did this

> like to make our own Chain object that has: id, native_address, wrapped_address as functions or attributes?

I can simply not allow for the user to pass the native and wrapped token, this isn't a big deal/change at all no?

Comment (dianacarvalho1) [reply src/encoding/models.rs:121]: aah and hardcode them there?? aaah yes that would work!!

File: src/encoding/models.rs [hunk 4]
```diff
@@ -105,36 +101,33 @@ impl From<TychoCoreChain> for ChainId {
...
-        Ok(Chain {
-            id: chain.into(),
-            name: chain.to_string(),
-            native_token: native_token_address,
-            wrapped_token: wrapped_token_address,
-        })
+impl Chain {
+    pub fn native_token(&self) -> Result<Bytes, EncodingError> {
+        match self.id.id() {
+            1 => Ok(Bytes::from(hex::decode("0000000000000000000000000000000000000000").map_err(
```
dianacarvalho1, line 124: I think it could be just Bytes::from_str("0x1234..").map_err(...)

dianacarvalho1 (approved):
Yeeess thank you! 🙏🏼 🙏🏼

src/encoding/models.rs
tamaralipows: Wrong :(

```
124 |             1 => Ok(Bytes::from_str("0000000000000000000000000000000000000000").map_err(
    |                            ^^^^^^^^ function or associated item not found in `Bytes`
```
    
I guess because this is the bytes object form Tycho, not directly from the Bytes lib, idk.

tamaralipows: I've tried a few things. I'm merging for now and will come back to it later if I find a more elegant way

PR #53, opened 2025-02-04
docs: In-code docs for encoders

tamaralipows:


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -213,7 +220,11 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
 }
 
 /// This strategy encoder is used for solutions that are sent directly to the pool.
```
royvardhan: We know that sending directly to the pool mean using the executor directly. Should we replace this and write "that are sent directly to the executor"?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -140,6 +151,11 @@ impl SwapEncoder for UniswapV3SwapEncoder {
     }
 }
 
+/// Encodes a swap on a Balancer pool through the given executor address.
```
royvardhan: Balancer V2 pool perhaps? 🤔

src/encoding/models.rs
```diff
@@ -49,10 +56,16 @@ pub struct Swap {
     pub token_in: Bytes,
     /// Token being output from the pool.
     pub token_out: Bytes,
-    /// Percentage of the amount to be swapped in this operation (for example, 0.5 means 50%)
+    /// Decimal of the amount to be swapped in this operation (for example, 0.5 means 50%)
     pub split: f64,
 }
 
+/// Represents a transaction to be executed on the Ethereum network.
```
royvardhan: This should be "any EVM supported network"

royvardhan (commented):
..just some questions and comments.

src/encoding/models.rs
tamaralipows: Ups this actually shouldn't mention EVM or Ethereum at all, it's more general than that. Thanks for pointing out though

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -74,6 +90,14 @@ impl SplitSwapStrategyEncoder {
         Ok(Self { permit2: Permit2::new(signer_pk, chain)?, selector, swap_encoder_registry })
     }
 
+    /// Raises an error if the split percentages are invalid.
+    ///
+    /// Split percentages are considered valid of all the following conditions are met:
```
royvardhan, line 95: *if* all the following conditions are met.

royvardhan: *if* all the following conditions are met.

src/encoding/evm/tycho_encoder.rs
```diff
@@ -11,20 +11,36 @@ use crate::encoding::{
...
 
 impl<S: StrategyEncoderRegistry> EVMTychoEncoder<S> {
+    /// Raises an `EncodingError` if the solution is not considered valid.
+    ///
+    /// A solution is considered valid if all the following conditions are met:
+    /// * The solution is not exact out.
+    /// * The solution has at least one swap.
+    /// * If the solution is wrapping, the given token is the chain's native token and the first
+    ///   swap's input is the chain's wrapped token.
+    /// * If the solution is unwrapping, the checked token is the chain's native token and the last
```
royvardhan, line 59: This is nice, these same comments apply to `validate_swap_path` on the `strategy_encoder` as well.

royvardhan (approved):

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -58,6 +67,13 @@ pub trait EVMStrategyEncoder: StrategyEncoder {
     }
 }
 
+/// Represents the encoder for a swap strategy which supports single, sequential and split swaps.
+///
+/// # Fields
+/// * `swap_encoder_registry`: SwapEncoderRegistry, containing all possible swap encoders
+/// * `permit2`: Permit2, the object containing necessary information for managing permit2
```
dianacarvalho1: no, not true. Permit2 is the approval manager itself. that will create the Permit object and etc. Should we rename it?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -144,6 +168,12 @@ impl SplitSwapStrategyEncoder {
         Ok(())
     }
 
+    /// Raises an error if swaps do not represent a valid path from the given token to the checked
+    /// token.
+    ///
+    /// A path is considered valid of all the following conditions are met:
+    /// * The checked token is reachable from the given token through the swap path
+    /// * There are tokens which are unconnected from the main path
```
dianacarvalho1:
```suggestion
    /// * There aren't tokens which are unconnected from the main path
```

src/encoding/evm/swap_encoder/swap_encoder_registry.rs
```diff
@@ -7,12 +7,16 @@ use crate::encoding::{
...
+/// Registry containing all supported `SwapEncoders`.
 #[derive(Clone)]
 pub struct SwapEncoderRegistry {
+    /// A hashmap containing the protocol system as a key and the `SwapEncoder` as a value.
     encoders: HashMap<String, Box<dyn SwapEncoder>>,
 }
 
 impl SwapEncoderRegistry {
+    /// Populates the registry with the `SwapEncoders` for the given blockchain by parsing the
+    /// executors in the file at the given path.
```
dianacarvalho1:
```suggestion
    /// Populates the registry with the `SwapEncoders` for the given blockchain by parsing the
    /// executors' addresses in the file at the given path.
```

dianacarvalho1 (approved):
Thank you so much @tamaralipows 🙏🏼 This looks great!! only have minor comments!

src/encoding/evm/tycho_encoder.rs
tamaralipows: Hmm I don't think it is explicitly checking that - it just expects that condition to be true already

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Ah I got confused just seeing the struct, didn't scroll down to the implementation - you're right! I think the name is fine. Changing doc

src/encoding/evm/tycho_encoder.rs
royvardhan: For `validate_swap_path` you could add something like 'If the `given token` or/and `checked token` is Native, then it applies the NativeAction of wrap/unwrap.

royvardhan: Thanks a lot!

PR #57, opened 2025-02-06
feat: Misc changes needed to add execution to quickstart

dianacarvalho1:
see individual commits for info on what was done


src/encoding/evm/approvals/permit2.rs
```diff
@@ -62,15 +71,16 @@ sol! {
...
-        let runtime = tokio::runtime::Handle::try_current()
-            .is_err()
-            .then(|| {
-                tokio::runtime::Runtime::new().map_err(|_| {
-                    EncodingError::FatalError("Failed to create tokio runtime".to_string())
-                })
-            })
-            .ok_or(EncodingError::FatalError("Failed to get tokio runtime".to_string()))??;
-        let client = runtime.block_on(get_client())?;
+        let (handle, runtime) = match Handle::try_current() {
```
dianacarvalho1, line 73: in the end my difficulties were all because we need to make sure both scenarios work: when there is an external runtime already and when it doesn't.
The only way to get the runtime is with `Handle::try_current()` but this doesn't give us the runtime, it gives only a handle to the runtime.
When creating a runtime from scratch, we need to save it somewhere, otherwise it just dies and our handle is useless.
We also need to use block_in_place instead of just block_on because we want to block the main runtime, not a random thread.

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -107,7 +107,7 @@ impl SwapEncoder for UniswapV3SwapEncoder {
         let mut pool_fee_bytes = swap
             .component
             .static_attributes
-            .get("pool_fee")
+            .get("fee")
```
tamaralipows, line 110: Uff thank you!

src/encoding/evm/approvals/permit2.rs
```diff
@@ -61,15 +70,16 @@ sol! {
...
+            Ok(h) => (h, None),
+            Err(_) => {
+                let rt = Arc::new(Runtime::new().map_err(|_| {
+                    EncodingError::FatalError("Failed to create a new tokio runtime".to_string())
+                })?);
+                (rt.handle().clone(), Some(rt))
```
tamaralipows, line 79: I'm a bit shocked that you can just clone the handle like this and it will still be the same handle for the runtime lol. I think I struggled with that before and gave up. Though I am assuming you have tested both cases so this is fine

src/encoding/evm/approvals/permit2.rs
```diff
@@ -61,8 +61,14 @@ sol! {
...
-        let runtime = Runtime::new()
-            .map_err(|_| EncodingError::FatalError("Failed to create runtime".to_string()))?;
+        let runtime = tokio::runtime::Handle::try_current()
+            .is_err()
+            .then(|| {
+                tokio::runtime::Runtime::new().map_err(|_| {
+                    EncodingError::FatalError("Failed to create tokio runtime".to_string())
+                })
+            })
+            .ok_or(EncodingError::FatalError("Failed to get tokio runtime".to_string()))??;
```
tamaralipows: Why are we doing `??`, was a single `?` not sufficient?

src/encoding/config/executor_addresses.json
```diff
@@ -1,6 +1,7 @@
 {
   "ethereum": {
     "uniswap_v2": "0x5C2F5a71f67c01775180ADc06909288B4C329308",
+    "uniswap_v3": "0x5C2F5a71f67c01775180ADc06909288B4C329308",
```
tamaralipows, line 4: I assume these values don't matter since we will overwrite them anyway? Just funny that it's the same as usv2

tamaralipows (approved):
Thank you! This was a hard baby! I just have some non-blocking questions for my own knowledge

src/encoding/evm/approvals/permit2.rs
royvardhan: Thanks, this will also help me when I face a similar issue :)

royvardhan (approved):
I have nothing to add. Looks good!

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: yes yes, this works fine for both cases. the handle is just a pointer to the runtime. you can clone it as much as you want. Here in the [docs](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#sharing) they say that to share a runtime you should use:


> - an [Arc](https://doc.rust-lang.org/nightly/alloc/sync/struct.Arc.html)<Runtime>.
> - a [Handle](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html).
> - Enter the runtime context.

src/encoding/evm/approvals/permit2.rs
dianacarvalho1: no. because there were two results here one after the other. I also didn't like it. the final version doesn't have this though

src/encoding/config/executor_addresses.json
dianacarvalho1: yes yes, I mean, they don't exist yet.. so for now they are only placeholders

src/bin/lib/cli.rs
```diff
@@ -45,6 +44,6 @@ pub struct Cli {
     pub private_key: Option<String>,
 
     /// Path to the executor addresses configuration file
-    #[arg(short, default_value = DEFAULT_CONFIG_PATH)]
-    pub config_path: String,
+    #[arg(short)]
+    pub config_path: Option<String>,
```
royvardhan, line 43: Noice!

src/encoding/evm/approvals/permit2.rs
tamaralipows: Ahhhh thank you! So good to know!

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -113,19 +113,16 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
 
-        let pool_fee_u24: [u8; 3] = pool_fee_bytes[pool_fee_bytes.len() - 3..]
+        let pool_fee_u24: [u8; 3] = padded_fee_bytes[(padded_fee_bytes.len() - 3)..]
             .try_into()
-            .map_err(|_| {
-                EncodingError::FatalError(
-                    "Pool fee static attribute must be at least 3 bytes".to_string(),
-                )
-            })?;
+            .map_err(|_| EncodingError::FatalError("Failed to extract fee bytes".to_string()))?;
```
tamaralipows, line 125: Uff thank you

tamaralipows (approved):
Here, have another approve

PR #58, opened 2025-02-07
feat: add uniswap v4 executor

royvardhan:
**Adds Uniswap V4 Executor support with test for `ExactSingleInput` swap.**

Architecture: This executor inherits [V4Router](https://github.com/Uniswap/v4-periphery/blob/main/src/V4Router.sol) contract, that allows [multi operations](https://github.com/Uniswap/v4-periphery/blob/main/src/interfaces/IV4Router.sol). 

**Flow**: The user encodes the transaction, calls `swap` and the swap function offloads the calldata to `_executeActions` function. Callbacks and settlements are handled by the V4Router contract. 

**Important Notes**: 
- The swap function must return `amountOut` and `_executeActions` do not return it, so we need to determine it ourselves based on the balance changes of the receiver for the `tokenOut` . For this I have used `_decodeData` function to get the `tokenOut` and then determine the `amountOut` based on the balance changes of the `receiver` after the swap.
- Even though using `_executeActions` inheritance unlocks new possibilities, the new blocker was the _decodeData , meaning, different type of swaps need different type of encoding, and getting `tokenOut` and `finalAmountOut` is a bit different and hardcoding `_decodeData` to expect only `SWAP_EXACT_IN_SINGLE` calldata would limit the usage of wide possibilites of `_executeAction` . So for this pr, I have made the _decodeData dynamic.
- After adding the v4 periphery contracts, there were build issues due to solc mismatch, so I updated all the contract to `0.8.26`. I still havn't found a way to fix this. Need help.

**Terms**: The V4 router swap relies on actions, which are mentioned [here](https://github.com/Uniswap/v4-periphery/blob/main/src/libraries/Actions.sol). For our test I have used `SETTLE_ALL` and `TAKE` actions. Explanations below:

1. TAKE: You specify the exact amount you want to take
2. TAKE_ALL: It would take all available credit, which might be more than what we want from the swap
3. SETTLE: You specify the exact amount you want to pay
4. SETTLE_ALL: Automatically calculates and pays your full debt

**Tests**: Test uses ExactSingleInput encoding helper function to achieve the data type needed by the swap function.


foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,114 @@
...
+            currency1: Currency.wrap(zeroForOne ? tokenOut : tokenIn),
+            fee: fee,
+            tickSpacing: int24(tickSpacing),
+            hooks: IHooks(address(0))
+        });
+
+        bytes memory actions = abi.encodePacked(
+            uint8(Actions.SWAP_EXACT_IN_SINGLE),
+            uint8(Actions.SETTLE_ALL),
+            uint8(Actions.TAKE_ALL)
```
dianacarvalho1: we need all of this for a simple swap?? the `SWAP_EXACT_IN_SINGLE` doesn't actually transfer tokens is that it?

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,114 @@
...
+        Currency token,
+        address payer,
+        uint256 amount
+    ) internal override {
+        token.transfer(payer, amount);
+    }
+
+    function msgSender() public view override returns (address) {
+        return msg.sender;
+    }
```
dianacarvalho1: these were the functions that the V4Router needed?

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,114 @@
...
+                zeroForOne: zeroForOne,
+                amountIn: amountIn128,
+                amountOutMinimum: amountOut128,
+                hookData: bytes("")
+            })
+        );
+
+        params[1] = abi.encode(key.currency0, amountIn128);
+        params[2] = abi.encode(key.currency1, amountOut128);
+
```
dianacarvalho1: couldn't we do this in the encoding side?

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,96 @@
...
+            USDE_ADDR,
+            USDT_ADDR,
+            uint24(100), // 0.01% fee tier
+            address(this),
+            true,
+            int24(1)
+        );
+
+        uint256 amountOut = uniswapV4Exposed.swap(amountIn, data);
+        assertEq(USDE.balanceOf(address(uniswapV4Exposed)), 0);
```
dianacarvalho1: and where did the funds go???

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Actually when using `SETTLE` and `TAKE`, i get this [error](https://github.com/Uniswap/v4-periphery/blob/cf451c4f55f36ea64c2007d331e3a3574225fc8b/src/libraries/CalldataDecoder.sol#L68) which originates from [BaseActionsRouter here](https://github.com/Uniswap/v4-periphery/blob/cf451c4f55f36ea64c2007d331e3a3574225fc8b/src/base/BaseActionsRouter.sol#L33). 

Also the Uniswap [docs](https://docs.uniswap.org/contracts/v4/guides/swap-routing) uses the `SETTLE_ALL` and `TAKE_ALL` actions even for the `swapExactInputSingle` demonstration.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Yes. Universal router override this _pay function with permit2 transfer [here](https://github.com/Uniswap/universal-router/blob/344d14179adea70e2626cd87c44499985c846862/contracts/modules/uniswap/v4/V4SwapRouter.sol#L15), I will use it too.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Yes we could do that, another point to note is that `executeActions` need only one param which is `actions`, so we could do all the encoding in the backend and then just call the executeActions inside the swap function.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: `USDE` went to PoolManager because that was `amountIn`, but the `amountOut` is also going back to the PoolManager after the successful swap. Which is funny! Tracing the logic which makes it go away!.

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: I don't understand. So the code that is committed fails? or SETTLE and SETLLE_ALL are different things?
I was not wondering at that difference (since I didn't even know that was a thing). I was wondering if we really need three actions and if SWAP_EXACT_IN_SINGLE alone is not enough

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: oh.. no. I don't think we want permit2 here.
actually.. do we need to approve the pool manager to spend our tokens??

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: yes I would like that

foundry/test/executors/UniswapV4Executor.t.sol
dianacarvalho1: what?? so all the funds are going to the pool manager?? that doesn't sound correct to me.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: My mistake, The commited code here passes, swap works and it requires both these actions. SETTLE and SETTLE_ALL are two different things.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Yes the V4Router needs these two functions, which we can override with our own implementation.

foundry/lib/bytes/LibPrefixLengthEncodedByteArray.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity ^0.8.28;
+pragma solidity ^0.8.26;
```
tamaralipows, line 2: Please for future reference if you are changing versions, or actually for any commit really, try to always include the reason **why** in the commit message.

Reason: someone might try to change this back in the future, being unaware of the issue you struggled with, only to come across the same issue themselves.

For example here, you could say

> feat: update solc and add V4Router into UniswapV4Executor...
> 
> ... because the last version was breaking _______ file. 

(replace with actual reason)

royvardhan: I was going to let you guys know in the PR description. It's just that I am not sure if I am going to keep this solc change because once this executor is complete, I will try to make the solc version back to last version and fix why it was breaking in this same pr, so that it's back to normal.

tamaralipows: ah alright then!

btw - it's good to also put those things in the commit if they become permanent. We have several times tracked changes back to someone's commit from a couple years ago and been confused why they did it, then ended up discovering it on our own with a lot more work haha

tamaralipows: We do rely on the commit messages in those cases, it's easier than looking up a PR from several years ago (let me know if you disagree - open to discussing)

royvardhan: No, you made a great point and I do totally agree with on this. 🙌

Nobody will see the PR description in 2 years even if I include the changes there. Commit history is what we would scan.

royvardhan: It's just that I was focused on making the swap work and thought of tackling the solc issue later.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Fixed it.

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,162 @@
...
+    constructor(IPoolManager _poolManager) V4Router(_poolManager) {}
+
+    function swap(uint256, bytes calldata data)
+        external
+        payable
+        returns (uint256 amountOut)
+    {
+        (
+            ,
```
dianacarvalho1: why is the formatting so weird?

dianacarvalho1: if we don't need the rest, no need to return them

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,162 @@
...
+
+        // First byte of actions determines the swap type
+        uint8 action = uint8(bytes1(actions[0]));
+
+        // Get receiver from params[2] for all cases
+        (, address _receiver,) =
+            abi.decode(params[2], (Currency, address, uint256));
+        receiver = _receiver;
+
+        if (action == uint8(Actions.SWAP_EXACT_IN_SINGLE)) {
```
dianacarvalho1, line 86: I think we could collapse some of these if else statements

dianacarvalho1: on a first glance, they seem very equivalent

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,162 @@
...
+        ) = _decodeData(data);
+
+        uint256 balanceBefore = IERC20(tokenOut).balanceOf(receiver);
+
+        this.executeActions(data);
+
+        if (isExactInput) {
+            amountOut = IERC20(tokenOut).balanceOf(receiver) - balanceBefore;
+        } else {
+            amountOut = amount;
```
dianacarvalho1: ufff this is not necessarily correct. we should return the amount in then

foundry/src/executors/UniswapV4Executor.sol
royvardhan: In Solidity, if you don't need a var, you just mark them by `,` so that the compiler does not throw msg logs. But now I have removed the unnecessary returns as you suggested.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Each swap type struct has different types of [attributes](https://github.com/Uniswap/v4-periphery/blob/main/src/interfaces/IV4Router.sol) and the struct name is needed to decode the bytes. I have tried shortening it a bit in the latest commit, not possible to do more.

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,129 @@
...
+        uint256 balanceAfter = IERC20(tokenOut).balanceOf(receiver);
+
+        if (isExactInput) {
+            amountOut = balanceAfter - balanceBefore;
+        } else {
+            amountOut = amount;
+        }
+
+        // Checks if the amountOut is not 0.
+        // Slither does not allow strict equality checks.
```
dianacarvalho1: what does this mean? `<0` doesn't work?

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,129 @@
...
+        (bytes memory actions, bytes[] memory params) =
+            abi.decode(data, (bytes, bytes[]));
+
+        // First byte of actions determines the swap type
+        uint8 action = uint8(bytes1(actions[0]));
+
+        // Get receiver from params[2] for all cases
+        (, address _receiver,) =
+            abi.decode(params[2], (Currency, address, uint256));
+        receiver = _receiver;
```
dianacarvalho1: why do we need this `_receiver`? I mean the variable name

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,129 @@
...
+    address poolManager = 0x000000000004444c5dc75cB358380D2e3dE08A90;
+
+    function setUp() public {
+        uint256 forkBlock = 21817316;
+        vm.createSelectFork(vm.rpcUrl("mainnet"), forkBlock);
+        uniswapV4Exposed =
+            new UniswapV4ExecutorExposed(IPoolManager(poolManager));
+    }
+
+    function testDecodeParamsUniswapV4() public view {
```
dianacarvalho1: can we add tests for all cases please?

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,129 @@
...
+    function testSwapUniswapV4() public {
+        vm.startPrank(BOB);
+        uint256 amountIn = 100 ether;
+        deal(USDE_ADDR, address(uniswapV4Exposed), amountIn);
+        uint256 usdeBalanceBeforePool = USDE.balanceOf(poolManager);
+        uint256 usdeBalanceBeforeSwapExecutor =
+            USDE.balanceOf(address(uniswapV4Exposed));
+        assertEq(usdeBalanceBeforeSwapExecutor, amountIn);
+        uint256 usdtBalanceBeforeSwapBob = USDT.balanceOf(address(BOB));
+        assertEq(usdtBalanceBeforeSwapBob, 0);
```
dianacarvalho1: these asserts are unnecessary

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,129 @@
...
+            USDE_ADDR, USDT_ADDR, 100, BOB, true, 1, uint128(amountIn)
+        );
+
+        uint256 amountOut = uniswapV4Exposed.swap(amountIn, data);
+        assertEq(USDE.balanceOf(poolManager), usdeBalanceBeforePool + amountIn);
+        assertEq(
+            USDE.balanceOf(address(uniswapV4Exposed)),
+            usdeBalanceBeforeSwapExecutor - amountIn
+        );
+        assertTrue(USDT.balanceOf(BOB) == amountOut && amountOut > 0);
```
dianacarvalho1: the `amountOut > 0` is unnecessary

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,129 @@
...
+            currency1: Currency.wrap(zeroForOne ? tokenOut : tokenIn),
+            fee: fee,
+            tickSpacing: int24(tickSpacing),
+            hooks: IHooks(address(0))
+        });
+
+        bytes memory actions = abi.encodePacked(
+            uint8(Actions.SWAP_EXACT_IN_SINGLE),
+            uint8(Actions.SETTLE_ALL),
+            uint8(Actions.TAKE)
```
dianacarvalho1: why doesn't the TAKE_ALL work??

dianacarvalho1 (commented):
Thank you @royvardhan !! Great work here 👏🏼 👏🏼 Thank you for the thorough PR description! This is gaining shape 🤞🏼 
I have a few comments

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Can you explain a bit, sorry my brain is fried right now 😅

foundry/src/executors/UniswapV4Executor.sol
royvardhan: I want to check if `amountOut == 0` then revert, but slither does not allow `==` strict equality checks, this is why I had to use `< 1` check.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Yeah true, fixed it now.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: All the cases? Meaning for other type of swaps?

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Removed them now.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Removed.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Using `TAKE_ALL` the amountOut is sent back to the PoolManager.

royvardhan: but with `TAKE` it checks the recipient [here](https://github.com/Uniswap/v4-periphery/blob/ea2bf2e1ba6863bb809fc2ff791744f308c4a26d/src/V4Router.sol#L69)

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: So the swap typically returns the calculated amount, which can be the amount out or amount in. If exact output, we should return the amount in here because it's the amount that we didn't pass when calling the method

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: If we are just trying to bypass slither I'd put a conscious slither skip here, or try to deeply understand the reasoning why slither doesn't allow this

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,126 @@
...
+        );
+
+        (address tokenOut, address receiver, bool isExactInput, uint256 amount)
+        = uniswapV4Exposed.decodeData(data);
+
+        assertEq(tokenOut, USDT_ADDR);
+        assertEq(receiver, expectedReceiver);
+        assertTrue(isExactInput);
+        assertEq(amount, expectedAmount);
+    }
```
tamaralipows: If we are supporting all cases (not just single in), I think we should test all possible cases for decoding, especially since they are so different

foundry/test/executors/UniswapV4Executor.t.sol
dianacarvalho1: yes

foundry/test/executors/UniswapV4Executor.t.sol
dianacarvalho1: in take all I see it uses the msg.sender as the receiver.. why isn't our msg.sender the executor?

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: The exploit scenario is described here https://github.com/crytic/slither/wiki/Detector-Documentation#dangerous-strict-equalities

I don't think this applies to us? Trying to think of an exploit one moment

tamaralipows: This could only be used to steal remaining funds from the router if something shady is happening in `executeActions`. I didn't look super deep in it yet. I would suggest just adding the slither-ignore flag for now because checking amount < 1 sounds just as unsafe as checking == 0. Then we can come back to it later I guess

tamaralipows: Just thought about this some more - isn't it safe enough that we have a checked amount at the higher level if we need? Is it really necessary to check if the swap failed at this point? We don't do this for any other pools

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,127 @@
...
+        external
+        payable
+        returns (uint256 amountOut)
+    {
+        (address tokenOut, address receiver, bool isExactInput, uint256 amount)
+        = _decodeData(data);
+
+        uint256 balanceBefore = IERC20(tokenOut).balanceOf(receiver);
+
+        this.executeActions(data);
```
dianacarvalho1: why do we need the `this`? just `executeActions` is not enough? 
now in my tests I noticed that by using the `this` it isn't using the method you defined lower, I guess it's going to the RouterV4 method

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    function testSwap() public {
```
dianacarvalho1, line 58: I just noticed this test is a bit weird..
We don't need the startPrank with Bob. and the receiver should be the executor. we don't need to involve Bob at all here since the funds are already in the executor to begin with

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Yes we can directly use `_executeActions`. The reason why I had to use this.executeActions is because previously we were doing the encoding with exact in, take, settle etc inside the swap function, which was giving us the `bytes memory data` while the _executeActions expects `bytes calldata data`

Now that we have moved the encoding to the test file, we can remove the extra wrapping of _executeActions.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Agreed, I was thinking something else. Will remove it.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: thanks!

foundry/src/executors/UniswapV4Executor.sol
royvardhan: I have added the == 0 check as you suggested with the slither disable warning.

royvardhan: On another thought, you are right, I am removing the check entirely.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: The current executor does not approve the pool manager anything, and the settlement works. So the answer is no.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Hmm, the test passes the receiver [here](https://github.com/propeller-heads/tycho-execution/blob/a41171ad07f30aacbe5634b6658a210b06de5cdb/foundry/test/executors/UniswapV4Executor.t.sol#L128)

And then `TAKE` decodes the receiver [here](https://github.com/Uniswap/v4-periphery/blob/ea2bf2e1ba6863bb809fc2ff791744f308c4a26d/src/V4Router.sol#L69)

Let me know if I am looking at something wrong.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Sure!

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,135 @@
...
+contract UniswapV4Executor is IExecutor, V4Router {
+    using SafeERC20 for IERC20;
+    using CurrencyLibrary for Currency;
+
+    constructor(IPoolManager _poolManager) V4Router(_poolManager) {}
+
+    function swap(uint256, bytes calldata data)
+        external
+        payable
+        returns (uint256 amountInOrOut)
```
tamaralipows: historically we called this `calculatedAmount`, it may be a bit more clear than `amountInOrOut` but of course I'm biased. What do you think?

royvardhan: `calculatedAmount` sounds better tbh. Thanks for suggesting.

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,135 @@
...
+            uint256 amount
+        ) = _decodeData(data);
+
+        uint256 tokenOutBefore = IERC20(tokenOut).balanceOf(receiver);
+        uint256 tokenInBefore = IERC20(tokenIn).balanceOf(address(this));
+
+        _executeActions(data);
+
+        uint256 tokenOutAfter = IERC20(tokenOut).balanceOf(receiver);
+        uint256 tokenInAfter = IERC20(tokenIn).balanceOf(address(this));
```
tamaralipows: Just thought of an issue here - it may not be IERC20 tokens. Native ETH tokens are also supported so we need to cover that case

royvardhan: Great catch! This needs some discussion on how are we handling Native ETH swaps.

foundry/test/executors/UniswapV4Executor.t.sol
dianacarvalho1: I'm questioning why the TAKE_ALL shouldn't work. Afaiu it should work. I think the issue was that you had BOB as the msg.sender. Can you double check this please?

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -0,0 +1,135 @@
...
+    {
+        (
+            address tokenIn,
+            address tokenOut,
+            address receiver,
+            bool isExactInput,
+            uint256 amount
+        ) = _decodeData(data);
+
+        uint256 tokenOutBefore = IERC20(tokenOut).balanceOf(receiver);
```
dianacarvalho1: this naming is misleading.. it seems like the tokens change before and after. these are balances, not tokens

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: what is the question here? 
Here we are only checking the balances.. not the actual swaps

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: When I use `TAKE_ALL` without `BOB` being the msg.sender, the swap works but then it sends the amountOut back to the PoolManager.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: The V4 allows Native token swapping, so for amountIn cases we need to check the native token balance changes, that is what @tamaralipows is suggesting.

foundry/src/executors/UniswapV4Executor.sol
royvardhan: Yes agreed, let me change it asap

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Altough I have removed the `startPrank` as `BOB`

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: yes, and why can't we just check that here?

foundry/test/executors/UniswapV4Executor.t.sol
dianacarvalho1: AH! I understand! it's our `msgSender` that is not correctly implemented in the first place. can you investigate this please? it needs to be the address of the executor in this case. but in the general case it would need to be the router?? I'm not so sure how to handle this without testing. can you have a look please?

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: there are many examples of how we do this in the TychoRouter

royvardhan: I wanted to discuss how V4Router handles the native token swaps. For now I have added the native token balance checks.

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Yes taking a look right now.

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: If you have any specific questions or points to discuss please let us know in the channel! Other than that I am marking this as resolved, thank you 🙏

dianacarvalho1 (approved):
Thank you sooo much @royvardhan 🙏🏼 🙏🏼 I'm very happy with this PR! merge awaaayy

royvardhan:
> Thank you sooo much @royvardhan 🙏🏼 🙏🏼 I'm very happy with this PR! merge awaaayy

Thank you for all your support!

PR #60, opened 2025-02-11
feat: Add selector to Transaction

dianacarvalho1:
Return selector in the StrategyEncoder. This is needed for the ExecutorStrategyEncoder

### DIFF
--- a/src/encoding/models.rs
+++ b/src/encoding/models.rs
@@ -95,11 +95,13 @@ impl Swap {
 /// * `to`: Address of the contract to call with the calldata
 /// * `value`: Native token value to be sent with the transaction.
 /// * `data`: Encoded calldata for the transaction.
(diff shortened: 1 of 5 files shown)


royvardhan (approved):

src/encoding/models.rs
```diff
@@ -95,11 +95,13 @@ impl Swap {
 /// * `to`: Address of the contract to call with the calldata
 /// * `value`: Native token value to be sent with the transaction.
 /// * `data`: Encoded calldata for the transaction.
+/// * `selector`: Only relevant for direct executions. The selector of the function to be called.
```
tamaralipows, line 98: For non-direct executions, is this included in the calldata? Or why is it only relevant for direct executions

dianacarvalho1: yes, for non direct ones that is already included somewhere inside the calldata

tamaralipows (approved):

PR #61, opened 2025-02-11
refactor: (Solution) replace check_amount with checked_amount

tamaralipows:
- This is to have consistency with checked_token


dianacarvalho1 (approved):
i don't even want to see this. I accept

PR #63, opened 2025-02-11
fix: Miscellaneous audit remarks

dianacarvalho1:
- Move pause functions together
- Add missing zero checks
- Use openzepplin's sendValues instead of transfer


tamaralipows (approved):
nice!

foundry/src/TychoRouter.sol
```diff
@@ -161,8 +157,7 @@ contract TychoRouter is
             _unwrapETH(amountOut);
         }
         if (tokenOut == address(0)) {
-            // slither-disable-next-line arbitrary-send-eth
-            payable(receiver).transfer(amountOut);
+            Address.sendValue(payable(receiver), amountOut);
```
royvardhan, line 160: Nice! Didn't know we could do this.

royvardhan (approved):

PR #65, opened 2025-02-13
feat: UniswapV4 encoding

tamaralipows:
- To keep any knowledge of USV4 separate from regular splits, I've made a new USV4 encoding strategy that will be used only if we detect sequential USV4 swaps.

- For single USV4 swaps without necessary optimizations, the regular split swap strategy will be used
- No need to change the swap struct interface to take multiple swaps - this concatenation can be done at the swap strategy level.

TODO:
- [x]  deduplicate code from split strategy
- [x]  test me
- [x]  separate grouping logic to its own method
- [x]  ensure we don't group split swaps
- [x]  generalize this to any protocol (not just usv4)
- [x]  address PR comments


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,6 +258,390 @@ impl SplitSwapStrategyEncoder {
...
+    }
+
+    fn clone_box(&self) -> Box<dyn StrategyEncoder> {
+        Box::new(self.clone())
+    }
+}
+
+// TODO this is all repeated validation code - we should maybe put this somewhere else
+// NO NEED TO REVIEW THIS CODE YET
+impl UniswapV4StrategyEncoder {
```
tamaralipows: no reviewing after this point. it's all copypasta and to stop making my rust rover cry and that's it

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,6 +258,390 @@ impl SplitSwapStrategyEncoder {
...
+        }
+
+        let mut swaps = vec![];
+
+        let mut previous_protocol_data: Vec<u8> = vec![];
+        let mut first_usv4_in_token: Bytes = Bytes::default();
+        let mut last_swap_was_usv4 = false;
+
+
+        // ---------- COPIED CODE ENDS HERE - REVIEW STARTING NOW ----------
```
dianacarvalho1: I feel like I'm in a cooking contest: "YOUR TIME STARTS NOW" 😂

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -80,24 +81,14 @@ pub struct SplitSwapStrategyEncoder {
...
-            permit2: Permit2::new(signer_pk, chain.clone())?,
-            selector,
-            swap_encoder_registry,
-            native_address: chain.native_token()?,
-            wrapped_address: chain.wrapped_token()?,
-        })
-    }
+/// Validates whether a sequence of split swaps represents a valid solution.
+#[derive(Clone)]
+pub struct SplitSwapValidator;
```
dianacarvalho1: Can we put this in its' own file please? 🥺

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,39 +251,251 @@ impl SplitSwapStrategyEncoder {
     }
 }
 
-impl EVMStrategyEncoder for SplitSwapStrategyEncoder {}
+impl SplitSwapStrategyEncoder {
```
dianacarvalho1, line 108: this is a bit out of place. can you move it down so it is with everything else from split swap? or move those things up

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,39 +251,251 @@ impl SplitSwapStrategyEncoder {
...
+    }
+
+    fn clone_box(&self) -> Box<dyn StrategyEncoder> {
+        Box::new(self.clone())
+    }
+}
+
+impl UniswapV4StrategyEncoder {
+    #[allow(dead_code)]
+    pub fn new(
```
dianacarvalho1: can the new go up please? before the `impl StrategyEncoder for UniswapV4StrategyEncoder`

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,39 +251,251 @@ impl SplitSwapStrategyEncoder {
...
+        let min_amount_out = get_min_amount_for_solution(solution.clone());
+
+        // The tokens array is composed of the given token, the checked token and all the
+        // intermediary tokens in between. The contract expects the tokens to be in this order.
+        let solution_tokens: HashSet<Bytes> =
+            vec![solution.given_token.clone(), solution.checked_token.clone()]
+                .into_iter()
+                .collect();
+
+        let intermediary_tokens: HashSet<Bytes> = solution
```
dianacarvalho1: notice that for uniswap v4 this might not be the case. the router doesn't need to know about the intermediary univ4 tokens

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,39 +251,251 @@ impl SplitSwapStrategyEncoder {
...
+        } else {
+            tokens.push(solution.checked_token.clone());
+        }
+
+        let mut swaps = vec![];
+
+        let mut previous_protocol_data: Vec<u8> = vec![];
+        let mut first_usv4_in_token: Bytes = Bytes::default();
+        let mut last_swap_was_usv4 = false;
+        println!("Hello?");
```
dianacarvalho1: oh Hello there! 👋🏼

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,39 +251,251 @@ impl SplitSwapStrategyEncoder {
...
+
+            let current_swap_is_usv4 = swap.component.protocol_system == "uniswap_v4";
+            let encoding_context = EncodingContext {
+                receiver: solution.router_address.clone(),
+                exact_out: solution.exact_out,
+                router_address: solution.router_address.clone(),
+            };
+            let mut protocol_data = swap_encoder.encode_swap(swap.clone(), encoding_context)?;
+            let in_token;
+            println!("Hello?");
```
dianacarvalho1: hello helloooo

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -258,39 +251,251 @@ impl SplitSwapStrategyEncoder {
...
+            } else {
+                in_token = swap.clone().token_in;
+                // This is not a USV4 swap. Clear previous USV4 protocol data.
+                previous_protocol_data = vec![];
+            }
+
+            // This is the hardest part - we will need to have the input token be the first of the
+            // USV4 sequence, and the output token be the last, essentially removing
+            // intermediate tokens and pretending they don't exist... I think?
+            let swap_data = self.encode_swap_header(
```
dianacarvalho1: huuuuummmm 🤔

dianacarvalho1: we only need to this in the first uniswap v4 swap 🤔 
[edit]: ah I see you are replacing it below!

dianacarvalho1: huuuumm this is very complex. mainly because we need to token indexes to match and we can't have the univ4 intermediary tokens there.

I think we need to gather the v4 swaps before anything else. so something like:
- loop through the swaps:
  - if it's v4:
     - if it's the first one: keep the token in and create the protocol data**
     - if it's a middle one: append the protocol data to the previous one and maybe keep the token out
  -  if it's not v4:
     -  if there was no v4 before: skip
     - if there was before then we have a uniswap v4 swap bundle here. I think we could create a new struct: UniswapV4Bundle or something that would be
````
struct UniswapV4Bundle{
    token_in: Bytes,
    token_out: Bytes,
    amount_out_min: U256,
    zero_for_one: bool,
    protocol_data: Bytes,
}
````
  - this would output a vec of Bundles

then to get the intermediary tokens we would need to loop through all the swaps again:
- if it's an uniswap v4 one we get the first bundle (we can assert that the token in matches as well), we use only the token_in and token_out from the bundle for the intermediary tokens. 
- if the next swap is v4, ignore. 
- if it's something else, add the something else
- if we find some new uniswap v4 we go to the next bundle 

then for the main loop: loop through swaps yet again 😩 
- if it's the first uniswap v4:
  - get the corresponding bundle
  - encode it and add callback selector and executor address
  - add the swap header
- if it's a consecutive uniswap v4 - skip
- if it's another swap - do normally and reset the flag for uniswap v4 swaps


** the protocol data needs to be only 
```
        address intermediaryToken;
        uint24 fee;
        int24 tickSpacing;
```

Comment (dianacarvalho1) [reply src/encoding/evm/strategy_encoder/strategy_encoders.rs]: or.. what if instead of the Bundle we made a new Swap object where we can make a custom ProtocolComponent object that would hold the intermediary swaps? 🤔 that's a bit hacky but this way we could replace all the uniswap v4 swaps with this bundled one and then the two last loops from ☝🏼 would be simplified a lot

Comment (dianacarvalho1) [reply src/encoding/evm/strategy_encoder/strategy_encoders.rs]: I'm staring at this for a while now 😓 I think what you have does most of what I described already (so sorry for that!). it only misses the intermediary tokens thingy 😕

**Review by dianacarvalho1 — COMMENTED** [review]
Thank you so much @tamaralipows ! 🙏🏼 this is reaaaally hard hein? I left some suggestions but maybe these things would be best to discuss in a call!

File: src/encoding/evm/strategy_encoder/strategy_encoders.rs (hunk 6 above)
Comment (tamaralipows) [reply src/encoding/evm/strategy_encoder/strategy_encoders.rs]: we take care of this by indexing properly

File: src/encoding/evm/strategy_encoder/strategy_encoders.rs (hunk 9 above)
Comment (tamaralipows) [reply src/encoding/evm/strategy_encoder/strategy_encoders.rs]: I really don't understand this message at all (sorry). What is the issue with the intermediary tokens, easily explained?

Comment (tamaralipows) [reply src/encoding/evm/strategy_encoder/strategy_encoders.rs]: let's have a call in a moment

File: src/encoding/evm/strategy_encoder/strategy_encoders.rs (hunk 7 above)
Comment (royvardhan) [inline src/encoding/evm/strategy_encoder/strategy_encoders.rs]: https://www.youtube.com/watch?v=YQHsXMglC9A

File: src/encoding/evm/strategy_encoder/strategy_encoders.rs (hunk 9 above)
Comment (dianacarvalho1) [reply src/encoding/evm/strategy_encoder/strategy_encoders.rs]: > What is the issue with the intermediary tokens, easily explained?

in the end you are still doing `len(tokens)` for the number of tokens and that is not true

File: src/encoding/evm/strategy_encoder/strategy_encoders.rs [hunk 10]
```diff
@@ -258,39 +251,266 @@ impl SplitSwapStrategyEncoder {
...
+            wrapped_address: chain.wrapped_token()?,
+            split_swap_validator: SplitSwapValidator,
+        })
+    }
+
+    /// Group consecutive swaps which can be encoded into one swap execution for gas optimization.
+    ///
+    /// An example where this applies is the case of USV4, which uses a PoolManager contract
+    /// to save token transfers on consecutive swaps.
+    fn group_swaps(&self, swaps: Vec<Swap>) -> Vec<SwapGroup> {
```
tamaralipows, line 129: I'm starting to think we can actually make this abstract enough such that this doesn't require its own strategy anymore. And that we can perhaps reuse some logic for balancer v3 of necessary. Will look more into this on Monday!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -80,24 +81,14 @@ pub struct SplitSwapStrategyEncoder {
     selector: String,
     native_address: Bytes,
     wrapped_address: Bytes,
+    split_swap_validator: SplitSwapValidator,
```
royvardhan, line 105: This should be added in the code comment/docs above.

royvardhan (approved):
Hey! Thank you for this robust logic, while this completes the encoding part of the USV4 related swap, the only suggestion I have is we could potentially unify SplitSwapStrategyEncoder and UniswapV4StrategyEncoder as both share same struct elements, with UniswapV4StrategyEncoder using optimized `group_swaps`. The struct would look like 

```rust
pub struct UnifiedStrategyEncoder {
    swap_encoder_registry: SwapEncoderRegistry,
    permit2: Permit2,
    selector: String,
    native_address: Bytes,
    wrapped_address: Bytes,
    split_swap_validator: SplitSwapValidator,
    grouping_strategy: SwapGroupingStrategy,
```

And then group_swaps could become: 

```rust
fn group_swaps(&self, swaps: Vec<Swap>) -> Vec<SwapGroup> {
        match self.grouping_strategy {
            SwapGroupingStrategy::Individual => swaps
                .into_iter()
                .map(|swap| SwapGroup {
                    input_token: swap.token_in.clone(),
                    output_token: swap.token_out.clone(),
                    protocol_system: swap.component.protocol_system.clone(),
                    swaps: vec![swap],
                    split: swap.split,
                })
                .collect(),
            SwapGroupingStrategy::GroupConsecutiveV4 => {
                let mut grouped_swaps: Vec<SwapGroup> = Vec::new();
                let mut current_group: Option<SwapGroup> = None;
                let mut last_swap_was_usv4 = false;
                
                for swap in swaps {
                    let current_swap_is_usv4 = swap.component.protocol_system == "uniswap_v4";
                    if current_swap_is_usv4 && last_swap_was_usv4 {
                        // Second or later USV4 pool in a sequence of USV4 pools. Merge to the current group
                        if let Some(group) = current_group.as_mut() {
                            group.swaps.push(swap.clone());
                            // Update the output token of the current group.
                            group.output_token = swap.token_out.clone();
                        }
                    } else {
                        // Not second or later USV4 pool. Push the current group (if it exists) and then
                        // create a new group.
                        if let Some(group) = current_group.as_mut() {
                            grouped_swaps.push(group.clone());
                        }
                        current_group = Some(SwapGroup {
                            input_token: swap.token_in.clone(),
                            output_token: swap.token_out.clone(),
                            protocol_system: swap.component.protocol_system.clone(),
                            swaps: vec![swap.clone()],
                            split: swap.split,
                        });
                    }
                    last_swap_was_usv4 = current_swap_is_usv4;
                }
                if let Some(group) = current_group.as_mut() {
                    grouped_swaps.push(group.clone());
                }
                grouped_swaps
            }
        }
    }
```

This way we could simply do inside `encode_strategy`: 

`let grouped_swaps = self.group_swaps(solution.swaps);`

And `protocol_data` would be extracted this way:
```rust
let protocol_data = match self.grouping_strategy {
                SwapGroupingStrategy::Individual => {
                    swap_encoder.encode_swap(grouped_swap.swaps[0].clone(), encoding_context)?
                }
                SwapGroupingStrategy::GroupConsecutiveV4 => {
                    let mut grouped_protocol_data: Vec<Vec<u8>> = vec![];
                    for swap in grouped_swap.swaps.iter() {
                        let protocol_data =
                            swap_encoder.encode_swap(swap.clone(), encoding_context.clone())?;
                        grouped_protocol_data.push(protocol_data);
                    }
                    grouped_protocol_data.abi_encode_packed()
                }
            };
```

But this suggestion can be applied in a newer PR if we want to save time now, as we are only focusing on the implementation of the encoding here.

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Yesssss, And I have shared an idea for this. Thanks!

dianacarvalho1 (commented):
Oh myyyyyyyy!! I love this ❤️ it's so beautiful now!! Thank you so much for this refactor @tamaralipows 🙏🏼

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -95,213 +118,99 @@ impl SplitSwapStrategyEncoder {
...
-                        total_percentage += swap.split;
-                    }
+            let current_swap_protocol = swap.component.protocol_system.clone();
+            groupable_protocol = GROUPABLE_PROTOCOLS.contains(&current_swap_protocol.as_str());
+
+            // Split 0 can also mean that the swap is the remaining part of a branch of splits,
+            // so we need to check the last swap's out token as well
+            let no_split = swap.split == 0.0 && swap.token_in == last_swap_out_token;
+
+            if current_swap_protocol == last_swap_protocol && groupable_protocol && no_split {
```
royvardhan, line 143: It is checking if the current protocol is same as last protocol and also if it belongs to groupable protocol, so far so good but what about no_split part? Is this condition specifically for the last swap?

royvardhan (approved):

dianacarvalho1 (approved):
Aweesoomeeee! I love this PR 🙏🏼 Thank you so so much for pushing through and coming up with such a clean and beautiful solution! ✨ ✨ 
I have only one question (that can be addressed in a separate PR): Do you think we should allow for grouping swaps in the `ExecutorStrategyEncoder` as well? I'm thinking that yes

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: The second swap in a sequence must get 100% of the funds of the first swap, so we can't optimize situations like this

```
          /---> USV4 
USV4 -----
          \---> USV4
```
          
It's basically just checking for that.

PR #66, opened 2025-02-13
feat: refactor usv3 callback

royvardhan:
- Rename `ExecutionDispatcher` to `Dispatcher` and added `_handleCallback` fn inside it to handle callback data from both router fallback fn and protocol specific calls such as USV3 and USV4. `_handleCallback` uses `delegatecall` to call the provided address with the provided selector with the provided data, otherwise uses default `handleCallback` from `ICallback`. 
- Updated both USV3 and USV4 callback function to use `_handleCallback` directly.
- Removed `CallbackVerificationDispatcher` since all the pool verifications now happen inside Executor's callback functions.
- Added USV3 executor level swap and callback tests.

Note: There were discussions to add `verifyCallback` fn inside `ICallback`, but when I implementing it i found it would be similar looking function to `handleCallback`, as both would accept one param as calldata. So I have thought to keep the verification inside the `handleCallback` only.


foundry/src/ExecutionDispatcher.sol
```diff
@@ -79,4 +79,29 @@ contract ExecutionDispatcher {
 
         calculatedAmount = abi.decode(result, (uint256));
     }
+
+    function _handleCallback(bytes calldata data) internal {
+        // Take last 20 bytes (excluding the final byte)
+        address executor =
+            address(bytes20(data[data.length - 21:data.length - 1]));
```
dianacarvalho1: why are they at the end of the callback data?
and why are we excluding the last byte?

foundry/src/ExecutionDispatcher.sol
```diff
@@ -79,4 +79,29 @@ contract ExecutionDispatcher {
...
+
```
dianacarvalho1, line 88: we should have the selector as well. if it's empty, use the `IExecutor.handleCallback.selector`

foundry/src/TychoRouter.sol
```diff
@@ -240,17 +229,7 @@ contract TychoRouter is
...
-    }
-
-    /**
-     * @dev Check if the sender is correct and executes callback actions.
-     *  @param msgData encoded data. It must includes data for the verification.
-     */
-    function _executeGenericCallback(bytes calldata msgData) internal {
-        (uint256 amountOwed, address tokenOwed) = _callVerifyCallback(msgData);
-
-        IERC20(tokenOwed).safeTransfer(msg.sender, amountOwed);
```
dianacarvalho1, line 253: where did this go to?

foundry/src/executors/BalancerV2Executor.sol
```diff
@@ -59,6 +59,11 @@ contract BalancerV2Executor is IExecutor {
             IVault(VAULT).swap(singleSwap, funds, limit, block.timestamp);
     }
 
+    function handleCallback(bytes calldata data)
+        external
+        returns (bytes memory result)
+    {}
```
dianacarvalho1: can we raise instead of doing nothing?

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -50,6 +65,44 @@ contract UniswapV3Executor is IExecutor {
         }
     }
 
+    function handleCallback(bytes calldata msgData)
+        external
+        returns (bytes memory result)
+    {
+        // Skip first 4 bytes of function selector and decode the two int256 values
```
dianacarvalho1: why this skip?

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -50,6 +65,44 @@ contract UniswapV3Executor is IExecutor {
...
+        int256 amount1Delta,
+        bytes calldata data
+    ) internal view returns (uint256 amountIn, address tokenIn) {
+        // Skip the first 64 bytes (32 bytes offset + 32 bytes length)
+        data = data[64:];
+
+        tokenIn = address(bytes20(data[0:20]));
+        address tokenOut = address(bytes20(data[20:40]));
+        uint24 poolFee = uint24(bytes3(data[40:43]));
+
```
dianacarvalho1: we could simplify this logic now? and decode everything at once instead of passing things around

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -75,9 +128,9 @@ contract UniswapV3Executor is IExecutor {
 
     function _makeV3CallbackData(address tokenIn, address tokenOut, uint24 fee)
         internal
-        pure
+        view
         returns (bytes memory)
     {
-        return abi.encodePacked(tokenIn, tokenOut, fee);
+        return abi.encodePacked(tokenIn, tokenOut, fee, self);
```
dianacarvalho1: why is the self not needed here anymore?

foundry/test/executors/UniswapV3Executor.t.sol
```diff
@@ -22,17 +24,19 @@ contract UniswapV3ExecutorExposed is UniswapV3Executor {
...
 }
 
-contract UniswapV3ExecutorTest is UniswapV3ExecutorExposed, Test, Constants {
+contract UniswapV3ExecutorTest is Test, Constants {
     using SafeERC20 for IERC20;
 
     UniswapV3ExecutorExposed uniswapV3Exposed;
     IERC20 WETH = IERC20(WETH_ADDR);
     IERC20 DAI = IERC20(DAI_ADDR);
+    address factory = 0x1F98431c8aD98523631AE4a59f267346ea31F984;
```
dianacarvalho1: can we move this to constants actually please?

dianacarvalho1 (commented):
Thank you @royvardhan ! 🙏🏼 I have a few suggestions!

foundry/src/executors/UniswapV3Executor.sol
dianacarvalho1: or do we want to keep this as `uniswapV3SwapCallback` so that if we use only the executor the whole swap works?? Can you add a test for that case please?

dianacarvalho1:
> Shall I now delete the CallbackVerificationDispatcher related functions such as setCallbackVerifier and removeCallbackVerifier ?

yes, we can remove that!

foundry/src/ExecutionDispatcher.sol
dianacarvalho1, line 1: I would rename the ExecutionDispatcher to just Dispatcher

foundry/src/ExecutionDispatcher.sol
royvardhan: Somehow it includes extra '00' in the calldata in the end.

foundry/src/executors/UniswapV3Executor.sol
royvardhan: As we are expecting the calldata to include int256,int256, and then other datas. The first 4 bytes are selector, and to get the ints and other data we need to skip the selector byte.

foundry/src/executors/UniswapV3Executor.sol
royvardhan: Yes sure

royvardhan: But that would mean adding a `fallback` function in order to run `handleCallback` function

foundry/src/ExecutionDispatcher.sol
royvardhan: Roger that!

foundry/src/ExecutionDispatcher.sol
dianacarvalho1: is it to pad to 32 bytes?

foundry/src/ExecutionDispatcher.sol
tamaralipows: Wait why? This is to distinguish from the callback verification dispatcher @dianacarvalho1

foundry/src/executors/UniswapV3Executor.sol
dianacarvalho1: I don't understand why the selector is making it inside of this function..

foundry/src/executors/UniswapV3Executor.sol
dianacarvalho1: no no, you just need to rename the function to be `uniswapV3SwapCallback` I think

foundry/src/ExecutionDispatcher.sol
tamaralipows: Uhhhh I think we need to understand what this 00 means

foundry/interfaces/IExecutor.sol
```diff
@@ -24,6 +24,10 @@ interface IExecutor {
         uint256 givenAmount,
         bytes calldata data
     ) external payable returns (uint256 calculatedAmount);
+
+    function handleCallback(
+        bytes calldata callbackData
+    ) external returns (bytes memory result);
```
tamaralipows: Butting this in the IExecutor means that all executors will have to implement this, which isn't the case for those that don't have callback such as USV2 right?

foundry/src/executors/BalancerV2Executor.sol
tamaralipows: Why are we putting this in the interface at all

foundry/src/executors/UniswapV3Executor.sol
tamaralipows: The selector should not be here o_o

tamaralipows (commented):
Just a few things! Looking good though

foundry/src/ExecutionDispatcher.sol
tamaralipows: In general it's not a good idea to delete things without understanding what they mean

foundry/interfaces/IExecutor.sol
royvardhan: yes and this is why I have left as blank, @dianacarvalho1 suggested to add a revert inside the ones which don't need callback.

royvardhan:
@tamaralipows @dianacarvalho1 I have updated the PR description with all new changes according to the design change discussion we had. Will require a new review.

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -50,6 +66,52 @@ contract UniswapV3Executor is IExecutor {
...
 
+    function uniswapV3SwapCallback(
+        int256 amount0Delta,
+        int256 amount1Delta,
+        bytes calldata data
+    ) external {
+        // slither-disable-next-line low-level-calls
+        (bool success, bytes memory result) = self.delegatecall(
+            abi.encodeWithSelector(
+                ICallback.handleCallback.selector,
```
royvardhan: I have to use delegatecall to call it's own function because `handleCallback` needs calldata and if I call it with `this.handleCallback` then the `msg.sender` changes to the executor's own address and pool verification fails. With delegatecall the msg.sender which is pool, is preserved here.

royvardhan: Btw, why not just completely remove the extra added `ICallback` which means getting rid of `handleCallback` function. And just use the protocol specification function names such as `uniswapV3SwapCallback` inside the executor.

With these changes the flow would be:

1. Router catches USV3 callback in uniswapV3SwapCallback,creates selector with `IUniswapCallback. uniswapV3SwapCallback.selector` and offloads the data to `_handleCallback(selector, data)`

And then _handleCallback delegate calls to the executor.

2. On the executor side, things are normal. It catches callbacks inside `uniswapV3SwapCallback` as standard.

foundry/src/Dispatcher.sol
```diff
@@ -79,4 +80,35 @@ contract ExecutionDispatcher {
 
         calculatedAmount = abi.decode(result, (uint256));
     }
+
+    function _handleCallback(bytes4 selector, bytes memory data) internal {
```
dianacarvalho1: I'm confused.. why is the selector passed but the executor address is not?

foundry/src/Dispatcher.sol
```diff
@@ -79,4 +80,35 @@ contract ExecutionDispatcher {
...
+        // Using assembly to access the last 20 bytes of the bytes memory data
+        address executor;
+        // slither-disable-next-line assembly
+        assembly {
+            let pos := sub(add(add(data, 0x20), mload(data)), 20)
+            executor := mload(pos)
+            executor := shr(96, executor)
+        }
+
```
dianacarvalho1: why are you using assembly? let's try to use solidity as much as we can. assembly is just shady

foundry/interfaces/ICallback.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+interface ICallback {
+    function handleCallback(
+        bytes calldata data
+    ) external returns (bytes memory result);
+}
```
dianacarvalho1, line 25: please add a `verifyCallback` function as well.  I saw your comment in the description but I still think it's best to have it explicitly.

foundry/interfaces/ICallback.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+interface ICallback {
```
dianacarvalho1, line 4: please add docs

foundry/src/TychoRouter.sol
```diff
@@ -240,17 +219,8 @@ contract TychoRouter is
...
+        bytes4 selector = bytes4(msg.data[:4]);
+        _handleCallback(selector, msg.data[4:]);
```
dianacarvalho1: it's weird that you are decoding only the selector

foundry/src/TychoRouter.sol
```diff
@@ -418,50 +366,20 @@ contract TychoRouter is
...
-        if (!executors[executor]) {
-            revert ExecutionDispatcher__UnapprovedExecutor();
-        }
-
-        // slither-disable-next-line controlled-delegatecall,low-level-calls
-        (bool success,) = executor.delegatecall(
-            abi.encodeWithSelector(selector, protocolData)
-        );
-        require(success, "delegatecall to uniswap v4 callback failed");
+        _handleCallback(selector, abi.encodePacked(protocolData, executor));
```
dianacarvalho1: why are you encoding again?? the data already has the selector and the executor

foundry/src/TychoRouter.sol
```diff
@@ -25,14 +22,7 @@ error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
 error TychoRouter__AmountInNotFullySpent(uint256 leftoverAmount);
 error TychoRouter__MessageValueMismatch(uint256 value, uint256 amount);
 
-contract TychoRouter is
-    AccessControl,
-    ExecutionDispatcher,
-    CallbackVerificationDispatcher,
-    Pausable,
-    ReentrancyGuard,
-    SafeCallback
```
dianacarvalho1, line 34: why did you remove the UniV4 callback?? In the meeting we agreed to keep the univ4 and v3 specific callbacks in the router

dianacarvalho1: you said it yourself in your [own message](https://propellerswap.slack.com/archives/C05QSEPUP37/p1739550120794519)

> Add univ3swap callback fn back to the router, whatever data it receives make it in bytes and use the dispatcher to delegatecall to the executor

foundry/src/executors/UniswapV3Executor.sol
dianacarvalho1: I'm so confused with this.
Why are you doing a delegate call to itself??? 😵‍💫

dianacarvalho1: this function should be in the Router. not here.
Here we need a function with the same name but that literally just calls the `handleCallback` (defined below)

dianacarvalho1 (changes requested):
Thank you @royvardhan! At this point I'm confused with this PR. Let me know if you need a call to discuss it. 
Also, please update your branch with main because I'm pretty sure the test CI will fail

foundry/src/Dispatcher.sol
royvardhan: The executor address is passed in the calldata through [_makeV3CallbackData](https://github.com/propeller-heads/tycho-execution/blob/14e5c127c6273df11383864503ddf1f13e9026a1/foundry/src/executors/UniswapV3Executor.sol#L138) which includes self address

foundry/src/Dispatcher.sol
royvardhan: Had to use assembly because of `bytes memory` not being directly slicable using plain Solidity. Will look for solutions.

foundry/interfaces/ICallback.sol
royvardhan: Okay will do.

foundry/src/TychoRouter.sol
royvardhan: Because as far as I know we are handling v3 and v4 callbacks in their respective callback functions. 

So I don't know what's left for `fallback`. Any data could be passed in this function right?

foundry/src/TychoRouter.sol
royvardhan: My bad, I misread the `protocolData` as sliced data.

foundry/src/Dispatcher.sol
dianacarvalho1: exactly, why isn't the selector passed like that as well?

foundry/src/TychoRouter.sol
royvardhan: The v4 call is there [right](https://github.com/propeller-heads/tycho-execution/blob/14e5c127c6273df11383864503ddf1f13e9026a1/foundry/src/TychoRouter.sol#L374)? I just removed the [SafeCallback](https://github.com/Uniswap/v4-periphery/blob/main/src/base/SafeCallback.sol) that validates the callback. Now the validation happens inside executors.

foundry/src/Dispatcher.sol
royvardhan: Ahh so you want it to pass the selector as well. Will add it.

foundry/src/TychoRouter.sol
dianacarvalho1: where are you validating? I don't see any added code in the v4 executor

foundry/src/executors/UniswapV3Executor.sol
royvardhan: You need `uniswapV3SwapCallback` in the executor that just calls the `handleCallback` which sounds plain and simple. But the problem are in types.

`handleCallback` expects `bytes calldata` but `uniswapV3SwapCallback` provides `bytes memory` which is a real pain. This is why I had to use `deletgatecall`

foundry/src/TychoRouter.sol
royvardhan: `SafeCallback` is inherited by `BaseActionsRouter` which is inherited by the `V4Router`

When we use [constructor](https://github.com/Uniswap/v4-periphery/blob/fdb5d3ea09d07eef380edf1a329140c98ff39dbb/src/V4Router.sol#L29) in our executor it is for the callback validation.

dianacarvalho1: But that is only used when we use the executor directly. When we use the router the verification is not done I think

dianacarvalho1: But that is only used when we use the executor directly. When we use the router the verification is not done I think

foundry/src/TychoRouter.sol
```diff
@@ -240,17 +219,8 @@ contract TychoRouter is
      *  caller is not a pool.
      */
     fallback() external {
-        _executeGenericCallback(msg.data);
```
dianacarvalho1, line 243: @royvardhan if you keep this logic you can keep this dat as calldata instead of memory. I think this would solve most of your problems

foundry/src/TychoRouter.sol
royvardhan: V3 also needs callback verification and that is handled inside the executor as well. 

Swaps go through the executors and verification happens when executors are used. I remember we agreed upon this in the meeting.

For reference:

V4 verification happens when unlockCallback at the executor level is called: [here](https://github.com/Uniswap/v4-periphery/blob/fdb5d3ea09d07eef380edf1a329140c98ff39dbb/src/base/SafeCallback.sol#L15)

V3 verification happens [here](https://github.com/propeller-heads/tycho-execution/blob/14e5c127c6273df11383864503ddf1f13e9026a1/foundry/src/executors/UniswapV3Executor.sol#L106)

dianacarvalho1: huuum ok ok, I see

foundry/src/TychoRouter.sol
royvardhan: I don't understand your suggestion. In our current logic, where V3 and V4 are handled by their own respective callback functions, what could be the use of a fallback function here?

foundry/src/TychoRouter.sol
royvardhan: No it is actually sliced data, the `data[:data.length - 24]` takes everything from the start up to, but not including, the last 24 bytes (which are the selector and executor). But I will update the `_handleCallback` to take the executor address as a param so that we don't have to encode it again.

foundry/src/TychoRouter.sol
dianacarvalho1: There are more protocols than the ones we have implemented that will have callbacks. Ideally we would be able to catch them here.
But that was not my suggestion. Most of your current problems are because the `data` is `bytes memory`. If you do this call here that converts it to `bytes calldata`

foundry/src/TychoRouter.sol
dianacarvalho1: no, you can refactor this method as well. it would make more sense not to decode any data here and pass everything to `_handleCallback` and do all the decoding there

royvardhan: If _handleCallback is enabled to receive `selector`, `executor` and `protocolData` as params, it will save us headache of decoding bytes memory in assembly. We could just use delegatecall with the protocolData, selector and the address without any decoding.

foundry/src/Dispatcher.sol
tamaralipows: What do you mean about bytes memory not being directly slicable? I think we are doing this in other places. What is the error you get - can you post it please?

royvardhan: Here is the error: `Index range access is only supported for dynamic calldata arrays.`

royvardhan: For code
```solidity
address executor = address(uint160(bytes20(data[data.length - 20:])));
```

dianacarvalho1 (approved):
Thank you @royvardhan 🙏🏼 This hard baby is ready to be merged!! ✨

PR #67, opened 2025-02-13
feat(univ4): Refactor execution input and handle all swap cases

dianacarvalho1:
Construct the uniswap v4 specific objects inside the executor The swap test in the executor alone doesn't work anymore because of the callback data prepended to the rest of he calldata

TODO:
- [x] Handle multiple swaps case
- [x] Decide where to put the executor and the selector for the callback
- [x] polish and make slither happy

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -449,9 +449,9 @@ contract TychoRouter is
         returns (bytes memory)
     {
         require(data.length >= 20, "Invalid data length");
-        bytes4 selector = bytes4(data[data.length - 24:data.length - 20]);
-        address executor = address(uint160(bytes20(data[data.length - 20:])));
-        bytes memory protocolData = data[:data.length - 24];
+        address executor = address(uint160(bytes20(data[0:20])));
+        bytes4 selector = bytes4(data[20:24]);
+        bytes memory protocolData = data[24:];
 
         if (!executors[executor]) {
             revert ExecutionDispatcher__UnapprovedExecutor();
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -449,9 +449,9 @@ contract TychoRouter is
         returns (bytes memory)
     {
         require(data.length >= 20, "Invalid data length");
-        address executor = address(uint160(bytes20(data[0:20])));
-        bytes4 selector = bytes4(data[20:24]);
-        bytes memory protocolData = data[24:];
+        bytes4 selector = bytes4(data[data.length - 4:]);
+        address executor = address(uint160(bytes20(data[data.length - 24:data.length - 4])));
+        bytes memory protocolData = data[:data.length - 24];
 
         if (!executors[executor]) {
             revert ExecutionDispatcher__UnapprovedExecutor();
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -450,7 +450,8 @@ contract TychoRouter is
     {
         require(data.length >= 20, "Invalid data length");
         bytes4 selector = bytes4(data[data.length - 4:]);
-        address executor = address(uint160(bytes20(data[data.length - 24:data.length - 4])));
+        address executor =
+            address(uint160(bytes20(data[data.length - 24:data.length - 4])));
         bytes memory protocolData = data[:data.length - 24];
 
         if (!executors[executor]) {
--- a/foundry/src/executors/UniswapV4Executor.sol
+++ b/foundry/src/executors/UniswapV4Executor.sol
@@ -16,7 +16,6 @@ import {V4Router} from "@uniswap/v4-periphery/src/V4Router.sol";
 import {Actions} from "@uniswap/v4-periphery/src/libraries/Actions.sol";
 import {IV4Router} from "@uniswap/v4-periphery/src/interfaces/IV4Router.sol";
 import {PathKey} from "@uniswap/v4-periphery/src/libraries/PathKey.sol";
-import "lib/forge-std/src/console.sol";
 
 error UniswapV4Executor__InvalidDataLength();
 
@@ -47,7 +46,7 @@ contract UniswapV4Executor is IExecutor, V4Router {
             UniswapV4Executor.UniswapV4Pool[] memory pools
         ) = _decodeData(data);
 
-        bytes memory fullData;
+        bytes memory swapData;
         if (pools.length == 1) {
             PoolKey memory key = PoolKey({
                 currency0: Currency.wrap(zeroForOne ? tokenIn : tokenOut),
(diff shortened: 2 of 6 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -449,9 +449,9 @@ contract TychoRouter is
         returns (bytes memory)
     {
         require(data.length >= 20, "Invalid data length");
-        bytes4 selector = bytes4(data[data.length - 24:data.length - 20]);
-        address executor = address(uint160(bytes20(data[data.length - 20:])));
-        bytes memory protocolData = data[:data.length - 24];
+        address executor = address(uint160(bytes20(data[0:20])));
+        bytes4 selector = bytes4(data[20:24]);
```
dianacarvalho1: I decided to put them in the beginning of the data. I think it's what makes the most sense considering our flow. BUT this breaks using the executor alone. Should I keep it at the end? even though it's hacky, it keeps that door open at least. Not sure tbh

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -40,39 +44,58 @@ contract UniswapV4ExecutorTest is Test, Constants {
...
+            UniswapV4Executor.UniswapV4Pool[] memory decodedPools
+        ) = uniswapV4Exposed.decodeData(data);
 
         assertEq(tokenIn, USDE_ADDR);
         assertEq(tokenOut, USDT_ADDR);
-        assertTrue(isExactInput);
-        assertEq(amount, expectedAmount);
-    }
-
-    function testSwap() public {
```
dianacarvalho1, line 59: because of the executor address and the selector (used for callback) now this test dies dies dies

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -24,108 +25,128 @@ contract UniswapV4Executor is IExecutor, V4Router {
...
+                    hookData: bytes("")
+                })
+            );
+            params[1] = abi.encode(key.currency0, amountIn);
+            params[2] = abi.encode(key.currency1, 0);
+            bytes memory swapData = abi.encode(actions, params);
+            fullData =
+                abi.encodePacked(callbackExecutor, callbackSelector, swapData);
+        } else {
+            console.log("do later");
```
tamaralipows: haha nice

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -24,108 +25,128 @@ contract UniswapV4Executor is IExecutor, V4Router {
...
     {
-        (address tokenIn, address tokenOut, bool isExactInput, uint256 amount) =
-            _decodeData(data);
+        (
+            address tokenIn,
+            address tokenOut,
+            uint256 amountOutMin,
+            bool zeroForOne,
+            address callbackExecutor,
+            bytes4 callbackSelector,
```
tamaralipows, line 45: I think we will actually need one of these per pool unfortunately, since each pool will have its token in and token out.

Another reason - the way I'm doing it on the encoder side:
if one param is duplicated for each pool, they all must be.

This is to avoid complicated logic of having to parse the output of the individual encoded swap. I would love if everything were just a list...

But maybe you'll understand better after deeply reviewing my PR

foundry/src/TychoRouter.sol
dianacarvalho1: yeah I will put it in the end again actually!

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -18,114 +18,164 @@ import {IV4Router} from "@uniswap/v4-periphery/src/interfaces/IV4Router.sol";
...
+            address intermediaryToken;
+            uint24 fee;
+            int24 tickSpacing;
+
+            // slither-disable-next-line assembly
+            assembly {
+                intermediaryToken := mload(add(poolsData, add(offset, 20)))
+                fee := shr(232, mload(add(poolsData, add(offset, 52))))
+                tickSpacing := shr(232, mload(add(poolsData, add(offset, 55))))
+            }
```
dianacarvalho1, line 176: I couldn't, for the life of me, decode the tickSpacing without assembly 😩 it's an int24 and I couldn't find another way tbh. Open to suggestions!

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: actually we don't need eheheh
we only need the token out of the intermediary pools!

foundry/test/TychoRouter.t.sol
```diff
@@ -857,15 +857,68 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        swaps[0] = swap;
+
+        tychoRouter.exposedSwap(amountIn, 2, pleEncode(swaps));
+
+        assertEq(IERC20(USDT_ADDR).balanceOf(tychoRouterAddr), 99943852);
+    }
+
+    function testSwapMultipleUSV4Callback() public {
+        // This test has two uniswap v4 hops that will be executed inside of the V4 pool manager
+        // USDE -> USDT -> WBTC
```
royvardhan, line 897: I wish they had v4 pool filter somewhere in their official site. Would be easy to find more pools.

royvardhan (approved):

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: did you try

`int24(uint24(bytes3(data[40:43])))`

(replacing with proper indices)

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: perhaps on the encoding side I can just pass a first_swap flag and include those things only if it's the first swap, and for every subsequent swap just append to the array?

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: yup :( it doesn't work.. for example in the test, it decodes to 0 instead of 60..

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: what? no no, the pool object is only created at the contract level, it doesn't need to be in encoding. in encoding you can just encode packed everything. See the `UniswapV4Utils` for how I'm doing in the tests

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: what the heck :/ 
ok keep the assembly then!

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: Removed that part of my message after reading more lol

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: I know.. I'm also baffled. was expecting you could tell me I was doing something really stupid 😓

tamaralipows (approved):
beaut

PR #69, opened 2025-02-18
feat: usv4 swap encoder

royvardhan:
- Added USV4 SwapEncoder that handles both simple and grouped swaps.
- Updated EncodingContext to take new attributes such as `group_token_in`, `group_token_out`, `amount_out_min`

Logic: For grouped swaps: It checks whether the `group_token_in` and `token_in` matches, if not then it only returns the required encoded [pool params](https://github.com/propeller-heads/tycho-execution/blob/4f37cd392bb5d578fe707a7686ad674f92fe31de/foundry/src/executors/UniswapV4Executor.sol#L26)
Otherwise, it returns the full encoded [swap params](https://github.com/propeller-heads/tycho-execution/blob/main/foundry/src/executors/UniswapV4Executor.sol#L40-L46)

Moved the following to utils:
- Selector encoder
- Static attributes getter
- Padding a byte to fix size


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -257,6 +259,9 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
                 receiver: solution.router_address.clone(),
                 exact_out: solution.exact_out,
                 router_address: solution.router_address.clone(),
+                group_token_in: Some(tokens.first().unwrap().clone()),
```
royvardhan: I haven't touched the tests of Strategy encoder, just added these Encoding Context values to keep the ci happy.

dianacarvalho1: I understand you didn't want the wrath of CI on you, but can you write here what these values are supposed to be please?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+///
+/// # Fields
+/// * `executor_address` - The address of the executor contract that will perform the swap.
+/// * `executor_selector` - The selector of the swap function in the executor contract.
+/// * `callback_selector` - The pre-computed selector of the callback function in the executor
+///   contract.
+#[derive(Clone)]
+pub struct UniswapV4SwapEncoder {
+    executor_address: String,
+    executor_selector: String,
```
dianacarvalho1: maybe now we could rename this to `swap_selector` actually

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+    ) -> Vec<u8> {
+        let mut encoded = Vec::with_capacity(26);
+        // Encode intermediary token (20 bytes)
+        encoded.extend_from_slice(intermediary_token.as_ref());
+        // Encode fee (3 bytes)
+        encoded.extend_from_slice(fee.as_ref());
+        // Encode tick spacing (3 bytes)
+        encoded.extend_from_slice(tick_spacing.as_ref());
+        encoded
+    }
```
dianacarvalho1: why do we need to encode like this? wouldn't `args.abi_encode_packed()` work?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+}
+
+impl SwapEncoder for UniswapV4SwapEncoder {
+    fn new(executor_address: String) -> Self {
+        // Pre-compute the callback selector for "unlockCallback(bytes)"
+        // This matches how Solidity computes function selectors
+        let callback_selector = keccak256(b"unlockCallback(bytes)")[..4]
+            .try_into()
+            .unwrap();
```
dianacarvalho1: please don't use unwrap in production code

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+/// # Fields
+/// * `executor_address` - The address of the executor contract that will perform the swap.
+/// * `executor_selector` - The selector of the swap function in the executor contract.
+/// * `callback_selector` - The pre-computed selector of the callback function in the executor
+///   contract.
+#[derive(Clone)]
+pub struct UniswapV4SwapEncoder {
+    executor_address: String,
+    executor_selector: String,
+    callback_selector: [u8; 4],
```
dianacarvalho1: it's a bit weird that the executor_selector is a `String` and the callback_selector is a `[u8; 4]`. Can we keep them consistent please?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+        sell_token_address < buy_token_address
+    }
+
+    fn encode_pool_params(
+        intermediary_token: Address,
+        fee: [u8; 3],
+        tick_spacing: [u8; 3],
+    ) -> Vec<u8> {
+        let mut encoded = Vec::with_capacity(26);
+        // Encode intermediary token (20 bytes)
```
dianacarvalho1: sorry but in my opinion, these comments are redundant. they make it look like chatgpt did this for you 👀

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+        &self,
+        swap: Swap,
+        encoding_context: EncodingContext,
+    ) -> Result<Vec<u8>, EncodingError> {
+        let mut first_swap = false;
+        if encoding_context.group_token_in == Some(swap.token_in.clone()) &&
+            encoding_context.group_token_out == Some(swap.token_out.clone())
+        {
+            first_swap = true;
+        }
```
dianacarvalho1: how would this ever be true? The group token in should be the **first** swap's token in but the group token out should be the **last** swap's token out

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+            .to_vec();
+
+        // Pad on the left with zeros if the fee is less than 3 bytes
+        let mut padded_fee_bytes = [0u8; 3];
+        let start = 3 - fee.len();
+        padded_fee_bytes[start..].copy_from_slice(&fee);
+
+        let pool_fee_u24: [u8; 3] = padded_fee_bytes[(padded_fee_bytes.len() - 3)..]
+            .try_into()
+            .map_err(|_| EncodingError::FatalError("Failed to extract fee bytes".to_string()))?;
```
dianacarvalho1: maybe we can make a function for all of this since we are using the exact same in many places now?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -147,6 +147,159 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+            .map_err(|_| {
+                EncodingError::FatalError("Failed to extract tick spacing bytes".to_string())
+            })?;
+
+        let pool_params =
+            Self::encode_pool_params(token_out_address, pool_fee_u24, pool_tick_spacing_u24);
+
+        if !first_swap {
+            return Ok(Self::encode_pool_params(
+                token_in_address,
```
dianacarvalho1: it should be the token out here

foundry/test/executors/UniswapV4Executor.t.sol
```diff
@@ -133,6 +133,27 @@ contract UniswapV4ExecutorTest is Test, Constants {
         assertTrue(USDT.balanceOf(address(uniswapV4Exposed)) == amountOut);
     }
 
+    function testSingleSwapIntegration() public {
```
dianacarvalho1, line 136: Can we have a complex swap test as well? You need to make a rust test that encodes the two swaps separately and then concatenate the data and use that as an input of the solidity test.

dianacarvalho1 (changes requested):
Thank you @royvardhan ! I left a few comments. Based on what I see here, it seems like there's still some room to deepen your understanding of Uniswap v4. Let me know if you have any questions, I’m happy to help clarify anything.

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: Sorry but I got the inspiration to add token_in from [here](https://propellerswap.slack.com/archives/C05QSEPUP37/p1739905917581789?thread_ts=1739898433.472069&channel=C05QSEPUP37&message_ts=1739905917.581789)

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: I used `executor_selector` to align with other encoders using the same naming.

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: Actually in solidity, the selector is always 4 bytes. The reason why I added it as fixed bytes is because we would immediately use the callback selector in the calldata. While `executor_selector` is not used anywhere in the swap encoder but in the strategy encoder [here](https://github.com/propeller-heads/tycho-execution/blob/4f37cd392bb5d578fe707a7686ad674f92fe31de/src/encoding/evm/strategy_encoder/strategy_encoders.rs#L275), converted to 4 bytes. 

To keep them consistent, I will make `callback_selector` as `string`, and then do the encoding in the returns.

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: haha true!

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: good catch! Suppose this is only single swap: 

`group_token_in == swap.token_in && group_token_out == swap.token_out`

But if this is part of more than one swap then what must I match?

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: yes I know. I'm suggesting changing it all together

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: your goal is to be able to identify the first swap. what do you think you need to match?

foundry/test/executors/UniswapV4Executor.t.sol
royvardhan: Yes I can do that!

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: only the `group_token_in` and  `swap_token_in`

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: I see. Tamara was providing conceptual guidance, but ultimately, you're responsible for ensuring everything makes sense. If you take a look at `testMultipleSwap`, you'll notice that it should be the token out.

royvardhan: yes it is my responsibility

tamaralipows: thanks for the correction guys :) I'll update the excalidraw for our reference

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -139,31 +134,163 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+            .ok_or_else(|| {
+                EncodingError::FatalError(
+                    "Pool tick spacing not found in Uniswap v4 static attributes".to_string(),
+                )
+            })?
+            .to_vec();
+
+        let pool_tick_spacing_u24 = pad_to_fixed_size::<3>(&tick_spacing).map_err(|_| {
+            EncodingError::FatalError("Failed to extract tick spacing bytes".to_string())
+        })?;
```
tamaralipows: For the params that are only necessary for the first swap, we should put them in an if block, otherwise we are computing them unnecessarily I think?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: I will take care of the strategy encoder @dianacarvalho1, no worries
I'll be working on it after this is merged, in order to do the integration test

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: Yes, agreed. Fixed it in the lastest commit.

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -139,9 +127,121 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+        sell_token_address < buy_token_address
+    }
+
+    fn encode_pool_params(
+        intermediary_token: Address,
+        fee: [u8; 3],
+        tick_spacing: [u8; 3],
+    ) -> Vec<u8> {
+        (intermediary_token, fee, tick_spacing).abi_encode_packed()
+    }
```
tamaralipows: feels overkill to have an extra method for this haha. Can we just move this one line to the main function?

tamaralipows (approved):
Just a minor comment - thank you!

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -139,31 +127,132 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+        let group_token_in = if let Some(group_token_in) = encoding_context.group_token_in {
+            bytes_to_address(&group_token_in)?
+        } else {
+            token_in_address
+        };
+        let group_token_out = if let Some(group_token_out) = encoding_context.group_token_out {
+            bytes_to_address(&group_token_out)?
+        } else {
+            token_out_address
+        };
```
dianacarvalho1: why is this necessary?? it should always be the group tokens, not the swap tokens. 
Or are you accounting for the single case? that in the single case the group tokens are not set?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -139,31 +127,132 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+        let mut amount_out_min = vec![0u8; 32]; // Zero-filled buffer of 32 bytes
+        let min_value = encoding_context
+            .amount_out_min
+            .unwrap_or_default()
+            .to_bytes_be();
+        amount_out_min[(32 - min_value.len())..].copy_from_slice(&min_value);
```
dianacarvalho1: why is this necessary? it should be just a U256

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -344,4 +442,249 @@ mod tests {
...
+                // - fee (3 bytes)
+                "000bb8",
+                // - tick spacing (3 bytes)
+                "00003c"
+            ))
+        );
+    }
+
+    #[test]
+    fn test_encode_uniswap_v4_combined() {
```
dianacarvalho1: can you improve this test a bit please?
naming is a bit strange: simple and grouped? it looked like you copy pasted the two previous tests all together.
it should be initial and second swap or something like this.
Also usually the tests have the following structure:
- setup
- calling the method that is being tested
- assertions

Here we are defining the first swap and encoding it, then defining the second swap (with the same named variables 😬 ) and then encoding it.
Can we define all the swaps first, encode them and then assert?

dianacarvalho1 (commented):
Thank you @royvardhan ! It is looking much better! I have a couple of requests still

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: Yes for single swaps, the group tokens are single swap token in and out.

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: Can you tell me how to directly convert BigUint to U256?

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: the group tokens will always be set

tamaralipows: so you can simplify this

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: using `biguint_to_u256`

royvardhan: thanks!

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: Thanks, I wasn't sure if I should make group tokens as non optional.

src/encoding/evm/swap_encoder/swap_encoders.rs
royvardhan: I have now defined the setup first and then encoded it.

dianacarvalho1:
Thank you 🙏🏻

dianacarvalho1 (approved):
Thank you! 🙏🏻

PR #70, opened 2025-02-18
feat: Tycho encoder builder

dianacarvalho1:
Create a EVMEncoderBuilder
- It has two three methods to be created:
  - new: where the user can pass any custom StrategyEncoder
  - tycho_router: where the default SplitSwapStrategyEncoder with Tycho Router will be used
  - direct_execution: where the user can encode only the execution data and integrate this into their workflow.
- This replaces StrategyEncoderRegistry
- EVMTychoEncoder holds directly the StrategyEncoder and not the registry (per one init of the EVMTychoEncoder there is only one possible StrategyEncoder)
- Update quickstart
- Update bin (add subcommands to bin to create a different instance of EVMEncoderBuilder)
- Remove direct_execution from Solution

TODO:
- After this is merged, update gitbook docs


src/bin/tycho-encode.rs
```diff
@@ -3,21 +3,16 @@ use std::io::{self, Read};
...
-    },
-    models::Solution,
-    strategy_encoder::StrategyEncoderRegistry,
-    tycho_encoder::TychoEncoder,
-};
+use tycho_execution::encoding::{models::Solution, tycho_encoder::TychoEncoder};
 
 mod lib {
     pub mod cli;
 }
```
kayibal: I just noticing this here by accident - but why is this called lib? I find it confusing since it isn't a library... I would get rid of this and just keep cli as a sibling here.

README.md
kayibal, line 1: Are we maintaining documentation here? Shouldn't this just link to the gitbook instead?

src/encoding/evm/encoder_builder.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub struct EVMEncoderBuilder {
+    strategy: Box<dyn StrategyEncoder>,
+    chain: Chain,
+}
+
+impl EVMEncoderBuilder {
+    pub fn new(chain: Chain, strategy: Box<dyn StrategyEncoder>) -> Self {
+        EVMEncoderBuilder { chain, strategy }
+    }
```
kayibal: Hm I would slightly prefer if we keep new just without any params and add two methods to set the chain and the strategy. This usually results in nice readable code.

```rust
let builder = EncoderBuilder::new();
builder.chain(Chain::from_str(cli.chain)?)?;
if ... {
    builder.strategy(...)
}
```

or

```rust
let builder = EncoderBuilder::new()
     .chain(Chain::Ethereum)
     .tycho_router(pk, executors_path)?;
```

src/encoding/evm/tycho_encoder.rs
```diff
@@ -4,41 +4,42 @@ use tycho_core::Bytes;
...
-/// * `strategy_registry`: S, the strategy registry to use to select the best strategy to encode a
-///   solution, based on its supported strategies and the solution attributes.
+/// * `strategy_encoder`: Strategy encoder to follow for encoding the solution
 /// * `native_address`: Address of the chain's native token
 /// * `wrapped_address`: Address of the chain's wrapped native token
-#[derive(Clone)]
-pub struct EVMTychoEncoder<S: StrategyEncoderRegistry> {
-    strategy_registry: S,
+pub struct EVMTychoEncoder {
+    strategy_encoder: Box<dyn StrategyEncoder>,
```
kayibal, line 18: Hm drastic change here.... Why was the type erasion necessary?

src/bin/tycho-encode.rs
```diff
@@ -33,8 +28,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let encoded = encode_swaps(&buffer, cli.config_path, cli.private_key)?;
-
+    let encoded = match cli.command {
+        Commands::TychoRouter { config_path, private_key } => {
+            encode_swaps(&buffer, config_path, Some(private_key), true)?
+        }
+        Commands::DirectExecution { config_path } => {
+            encode_swaps(&buffer, config_path, None, false)?
+        }
+    };
```
kayibal: This code block is weird, the flag seems strange here... but it's not too important since it isn't library code.

src/encoding/evm/swap_encoder/swap_encoder_registry.rs
```diff
@@ -19,8 +19,9 @@ impl SwapEncoderRegistry {
     /// executors' addresses in the file at the given path.
     pub fn new(
         executors_file_path: Option<String>,
-        blockchain: Chain,
+        blockchain: tycho_core::dto::Chain,
```
kayibal, line 22: Why do we use a data transfer object here?

src/encoding/evm/tycho_encoder.rs
```diff
@@ -4,41 +4,42 @@ use tycho_core::Bytes;
...
     native_address: Bytes,
     wrapped_address: Bytes,
 }
 
-impl<S: StrategyEncoderRegistry> EVMTychoEncoder<S> {
-    pub fn new(strategy_registry: S, chain: tycho_core::dto::Chain) -> Result<Self, EncodingError> {
+impl EVMTychoEncoder {
+    pub fn new(
+        chain: tycho_core::dto::Chain,
```
kayibal, line 25: We shouldn't use dto objects in higher level code...

kayibal (approved):
Thx looks a lot better already! I have a few comments though

src/encoding/evm/tycho_encoder.rs
kayibal: Not saying we should revert this now but probably it wasn't necessary. Not a big deal though

tamaralipows (commented):
I like this thanks! I agree with making a chain setter and strategy setter method in the builder. That's how we set attributes in a few of our other builders so it'll be a bit more consistent.

One question - do we want to have the split strategy as the default? Or did we consciously decide against it?

README.md
dianacarvalho1: yes but the gitbook has not been published yet 😐 so nothing to point to

src/encoding/evm/tycho_encoder.rs
dianacarvalho1: I thought having the generic S was a bit of an overkill when we know more stuff about what's allowed here or not (the `StrategyEncoder`). I thought it makes the code less obscure

src/bin/tycho-encode.rs
dianacarvalho1: yeah.. do you have any suggestion on how to do this in a different way?

src/encoding/evm/swap_encoder/swap_encoder_registry.rs
dianacarvalho1: isn't that what we are supposed to use in the first place? I thought that was the interface with tycho core 😬 We were using models initially but then changed to dto everywhere. Not long ago, I made a [PR](https://github.com/propeller-heads/tycho-execution/pull/54) for it and you [agreed](https://propellerswap.slack.com/archives/C05LQALE9L1/p1738761691390239?thread_ts=1738750328.370419&cid=C05LQALE9L1)

dianacarvalho1:
> One question - do we want to have the split strategy as the default? Or did we consciously decide against it?

@tamaralipows yes, it's the default when using the `tycho_router` option

royvardhan (approved):
Thank you for this, lgtm!

src/encoding/evm/encoder_builder.rs
```diff
@@ -11,34 +11,66 @@ use crate::encoding::{
...
-            Box::new(SplitSwapStrategyEncoder::new(swapper_pk, chain, swap_encoder_registry)?);
-        Ok(EVMEncoderBuilder { chain, strategy })
+        if let Some(chain) = self.chain {
+            let swap_encoder_registry = SwapEncoderRegistry::new(executors_file_path, chain)?;
+            let strategy =
+                Box::new(SplitSwapStrategyEncoder::new(swapper_pk, chain, swap_encoder_registry)?);
+            Ok(EVMEncoderBuilder { chain: Some(chain), strategy: Some(strategy) })
+        } else {
+            Err(EncodingError::FatalError(
+                "Please set the chain before setting the strategy".to_string(),
```
tamaralipows: Before setting the tycho router, no?

src/encoding/evm/encoder_builder.rs
```diff
@@ -11,34 +11,66 @@ use crate::encoding::{
...
         executors_file_path: Option<String>,
     ) -> Result<Self, EncodingError> {
-        let swap_encoder_registry = SwapEncoderRegistry::new(executors_file_path, chain)?;
-        let strategy =
-            Box::new(SplitSwapStrategyEncoder::new(swapper_pk, chain, swap_encoder_registry)?);
-        Ok(EVMEncoderBuilder { chain, strategy })
+        if let Some(chain) = self.chain {
+            let swap_encoder_registry = SwapEncoderRegistry::new(executors_file_path, chain)?;
+            let strategy =
+                Box::new(SplitSwapStrategyEncoder::new(swapper_pk, chain, swap_encoder_registry)?);
```
tamaralipows, line 55: I think we need to be careful not to overwrite the strategy here if it's already set (yes this works now since the strategy has to be split swap if we use the router, but in the future we may have other strategies that shouldn't be overwritten)

We could keep it as the default though

src/encoding/evm/encoder_builder.rs
```diff
@@ -11,34 +11,66 @@ use crate::encoding::{
...
     }
 
     pub fn build(self) -> Result<EVMTychoEncoder, EncodingError> {
-        EVMTychoEncoder::new(self.chain, self.strategy)
+        if let (Some(chain), Some(strategy)) = (self.chain, self.strategy) {
+            EVMTychoEncoder::new(chain, strategy)
+        } else {
+            Err(EncodingError::FatalError(
+                "Please set the chain and strategy before building the encoder".to_string(),
+            ))
```
tamaralipows, line 89: In addition to this, should we add some more checks? Like that either `tycho_router` or `direct_execution` is called? Or perhaps set the default to `tycho_router`? Just thinking out loud.

src/encoding/evm/encoder_builder.rs
```diff
@@ -11,34 +11,66 @@ use crate::encoding::{
...
+        self.chain = Some(chain);
+        self
+    }
+    pub fn strategy_encoder(mut self, strategy: Box<dyn StrategyEncoder>) -> Self {
+        self.strategy = Some(strategy);
+        self
     }
     pub fn tycho_router(
-        chain: Chain,
+        self,
```
tamaralipows, line 48: Not really sure what's the point of this method if we already have a strategy setter?

tamaralipows (commented):
Just a few more questions

src/encoding/evm/encoder_builder.rs
dianacarvalho1: it's to have a default behaviour. a one liner people can use. the new method + setters is for advanced users that want to do fancy things

src/encoding/evm/encoder_builder.rs
dianacarvalho1: huumm this is the default for the tycho_router option. I think if we have then a `cowswap` (or no approvals one, we need to talk naming) the strategy there would already be different

src/encoding/evm/encoder_builder.rs
dianacarvalho1: I think we should allow for people to create their own. and not force tycho_router or direct_execution

src/encoding/evm/encoder_builder.rs
tamaralipows: I'm thinking more if we have a single encoding strategy... here it would be overwritten to use the split encoding strategy

src/encoding/evm/encoder_builder.rs
tamaralipows: hmmmm ok that's not super clear to me just by looking at it. I would have been confused about what I need to set and what not. Should we comment it or add a docstring then?

src/encoding/evm/encoder_builder.rs
dianacarvalho1: aaah yes, we would need to find a way to support it. like make another of these functions or a multiplex strategy that then chooses the best sub strategy

src/encoding/evm/encoder_builder.rs
dianacarvalho1: yes, docstring is definitely missing. sorry

src/encoding/evm/encoder_builder.rs
tamaralipows: I guess that can wait until the moment comes where we add another strategy!

tamaralipows (approved):

PR #71, opened 2025-02-18
feat: Support swap grouping optimizations for direct execution

tamaralipows:
- In preparation for also using this for the executor encoder. This also make test setup much easier for swap grouping - not having to init the encoder every time.


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -329,31 +261,47 @@ impl StrategyEncoder for ExecutorStrategyEncoder {
...
-        let protocol_data = swap_encoder.encode_swap(swap.clone(), encoding_context)?;
+        let mut grouped_protocol_data: Vec<Vec<u8>> = vec![];
+        for swap in grouped_swap.swaps.iter() {
+            let encoding_context = EncodingContext {
+                receiver: receiver.clone(),
+                exact_out: solution.exact_out,
+                router_address: router_address.clone(),
+            };
+            let protocol_data = swap_encoder.encode_swap(swap.clone(), encoding_context.clone())?;
+            grouped_protocol_data.push(protocol_data);
```
dianacarvalho1: couldn't we just extend? this way we wouldn't need the `grouped_protocol_data.abi_encode_packed()` a few lines down

src/encoding/evm/strategy_encoder/utils.rs
dianacarvalho1, line 1: Maybe we could rename this to be specific of groups and not just utils.. `group_swaps.rs`?

dianacarvalho1 (approved):
Thank you so much @tamaralipows! 🙏🏼 only have small small comments

royvardhan (approved):

PR #74, opened 2025-02-20
feat: add multichain support in encoder

royvardhan:
Added `native_token`, `wrapped_token`, `dai_address`, `usdc_address` for the following networks:
 - Base
 - ZkSync
 - Arbitrum

### DIFF
--- a/src/encoding/models.rs
+++ b/src/encoding/models.rs
@@ -153,22 +153,33 @@ impl From<TychoCoreChain> for Chain {
 }
 
 impl Chain {
+    fn decode_hex(&self, hex_str: &str, err_msg: &str) -> Result<Bytes, EncodingError> {
+        Ok(Bytes::from(
+            hex::decode(hex_str).map_err(|_| EncodingError::FatalError(err_msg.to_string()))?,
+        ))
+    }
+
     pub fn native_token(&self) -> Result<Bytes, EncodingError> {
+        let decode_err_msg = "Failed to decode native token";
         match self.id.id() {
-            1 => Ok(Bytes::from(hex::decode("0000000000000000000000000000000000000000").map_err(
-                |_| EncodingError::FatalError("Failed to decode native token".to_string()),
-            )?)),
+            1 | 8453 | 42161 => {
+                self.decode_hex("0000000000000000000000000000000000000000", decode_err_msg)
+            }
+            324 => self.decode_hex("000000000000000000000000000000000000800A", decode_err_msg),
             _ => Err(EncodingError::InvalidInput(format!(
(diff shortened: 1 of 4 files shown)


src/encoding/models.rs
```diff
@@ -153,28 +153,67 @@ impl From<TychoCoreChain> for Chain {
...
 impl Chain {
+    fn decode_hex(&self, hex_str: &str, err_msg: &str) -> Result<Bytes, EncodingError> {
+        Ok(Bytes::from(
+            hex::decode(hex_str).map_err(|_| EncodingError::FatalError(err_msg.to_string()))?,
+        ))
+    }
+
     pub fn native_token(&self) -> Result<Bytes, EncodingError> {
+        let decode_err_msg = "Failed to decode native token";
         match self.id.id() {
```
dianacarvalho1: this is strange `self.id.id()`
I know you didn't do this. but do we need this? can't we simplify the Chain object? (not to have `self.id.id()`)

dianacarvalho1 (approved):
Thank you @royvardhan 🙏🏼 This looks good to me! I only have a request that is not even related to what you did here

src/encoding/models.rs
royvardhan: Ohh yeah, let me fix it. Thanks for noticing.

src/encoding/models.rs
```diff
@@ -153,28 +153,67 @@ impl From<TychoCoreChain> for Chain {
...
+            1 => self.decode_hex("A0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48", decode_err_msg),
+            8453 => self.decode_hex("833589fCD6eDb6E08f4c7C32D4f71b54bdA02913", decode_err_msg),
+            324 => self.decode_hex("1d17CBcF0D6D143135aE902365D2E5e2A16538D4", decode_err_msg),
+            42161 => self.decode_hex("af88d065e77c8cC2239327C5EDb3A432268e5831", decode_err_msg),
+            _ => Err(EncodingError::InvalidInput(format!(
+                "Usdc address not set for chain {:?}. Double check the chain is supported.",
+                self.name
+            ))),
+        }
+    }
```
tamaralipows: This is just for testing purposes? Feels weird that we are expanding a core model just for testing purposes tbh.
idk what you think @dianacarvalho1 @royvardhan

dianacarvalho1: I agree tbh 😕

royvardhan: Okay, then I will revert last two commits

tamaralipows (approved):
Thank you!

PR #76, opened 2025-02-20
refactor: use chain and protocol component from tycho_core::models

royvardhan:
Use tycho_core::models instead to tycho_core::dto objects


dianacarvalho1 (approved):
Thank you! 🙏🏼

PR #77, opened 2025-02-20
feat: Adapt SplitSwapStrategyEncoder to have optional permit2 logic

dianacarvalho1:
- If the signer_pk is not passed -> use the swap method that expects the tokens to be already in the Router
- If it is passed -> compute permit2 and use swap method that does the token in transfer
- Extend builder to have another shortcut
- Add integration test with contract
- Update bin (and simplified it) and quickstart


tamaralipows (approved):
Thank you! Looks like this is it!

royvardhan (approved):
Looks good to me!

PR #78, opened 2025-02-21
fix: Native ETH input/output integration tests/fixes

tamaralipows:
- Only for single swaps
- Used USV4 for this because it's the only DEX we support that allows native ETH swaps
- For Native ETH input single swaps, we need to properly check the remaining amount (we were treating them wrongly like ERC20 tokens)
- For Native ETH output single swaps, we were passing the incorrect currency (the settle always needs to be the out token and the take always needs to be the in token, this should not depend on the zeroForOne value).


foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -71,8 +71,8 @@ contract UniswapV4Executor is IExecutor, V4Router {
                     hookData: bytes("")
                 })
             );
-            params[1] = abi.encode(key.currency0, amountIn);
-            params[2] = abi.encode(key.currency1, uint256(0));
+            params[1] = abi.encode(tokenIn, amountIn); // currency to settle
+            params[2] = abi.encode(tokenOut, uint256(0)); // currency to take
             swapData = abi.encode(actions, params);
```
tamaralipows, line 76: This was an absolute horrific thing to figure out. I needed console logs eeevvvveryyywhere in the V4Router and DeltaResolver to find out what was wrong. In the end this fixed it. Let me know if this makes sense to you. 

This needs to match the order of the actions passed above (swap, settle, take). This means the settle token should always be second and the take token should always be third. We were changing this based on the zeroForOne, which isn't good. I'll double check if there is a similar issue tomorrow for the non-single swap case.

Why didn't our other integration tests catch this? Was it just luck? Didn't deeply check tbh.

Please let me know if this makes sense to you.

foundry/test/TychoRouter.t.sol
```diff
@@ -764,6 +762,61 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        //   USDC ───(USV4)──> ETH
+        //
+        deal(USDC_ADDR, ALICE, 3000_000000);
+        uint256 balancerBefore = ALICE.balance;
+
+        // Approve permit2
+        vm.startPrank(ALICE);
+        IERC20(USDC_ADDR).approve(address(permit2Address), type(uint256).max);
+
+        // Encoded solution generated using `test_split_encoding_strategy_usv4_eth_in`
```
dianacarvalho1: eth_out

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: ooh shiiittt! great catch here!! this makes a lot of sense tbh!! 😬 yeah this was just wrong before

dianacarvalho1: just checked and for the multi swap it is correct though 🙏🏼

dianacarvalho1 (approved):
Thank you thank you thank you much @tamaralipows 🙏🏼 awesome that you found these bugs!! 
I think we were just lucky in the single swap cases (we always used the same pool and same trade direction in all tests 😕 )

foundry/test/TychoRouter.t.sol
```diff
@@ -764,6 +762,61 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        uint256 balancerAfter = IERC20(PEPE_ADDR).balanceOf(ALICE);
+
+        assertTrue(success, "Call Failed");
+        assertEq(balancerAfter - balancerBefore, 242373460199848577067005852);
+    }
+
+    function testUSV4IntegrationOutputETH() public {
+        // Test created with calldata from our router encoder.
+
+        // Performs a single swap from ETH to PEPE without wrapping or unwrapping
```
royvardhan: from USDC to PEPE

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -1006,8 +1006,129 @@ mod tests {
...
+            swaps: vec![swap_eth_pepe],
+            ..Default::default()
+        };
+
+        let (calldata, _, _) = encoder
+            .encode_strategy(solution)
+            .unwrap();
+        let hex_calldata = encode(&calldata);
+
+        println!("{}", hex_calldata);
```
royvardhan, line 1179: printer goes brrrrrr

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -1006,8 +1006,129 @@ mod tests {
...
+            swaps: vec![swap_usdc_eth],
+            ..Default::default()
+        };
+
+        let (calldata, _, _) = encoder
+            .encode_strategy(solution)
+            .unwrap();
+
+        let hex_calldata = encode(&calldata);
+        println!("{}", hex_calldata);
```
royvardhan, line 1243: printooor

royvardhan (changes requested):
This is what giga-brain prs look like! Amazing catch!

royvardhan (approved):

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: i keep these prints here because they are important for the test - since I'm not asserting anything I want to print the calldata to plug into the solidity side

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: A lesson to check both swap directions in the future

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: okay!

PR #80, opened 2025-02-21
feat: add target verification for Uniswap V2 and V3

royvardhan:
Added target(pool) check using `_computePairAddress` in Uniswap V2 and V3. Uses `Factory` and protocol's official pool init hash for computing pair address. 

And since Uniswap V4, Balancer V2 uses a single entrypoint for all swaps(pool manager / vault), they didn't need such checks.


royvardhan:
Omg! How many times this will go out-of-date

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -146,4 +152,27 @@ contract UniswapV3Executor is IExecutor, ICallback {
...
+            tokenA < tokenB ? (tokenA, tokenB) : (tokenB, tokenA);
+        pool = address(
+            uint160(
+                uint256(
+                    keccak256(
+                        abi.encodePacked(
+                            hex"ff",
+                            factory,
+                            keccak256(abi.encode(token0, token1, fee)),
+                            hex"e34f199b19b2b4f47f68442619d555527d244f78a3297ea89325f843f87b8b54"
```
tamaralipows: What's this?

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -22,6 +26,10 @@ contract UniswapV2Executor is IExecutor {
         IERC20 tokenIn;
 
         (tokenIn, target, receiver, zeroForOne) = _decodeData(data);
+
+        if (target != _computePairAddress(target)) {
+            revert UniswapV2Executor__InvalidTarget();
+        }
```
tamaralipows: Feels a bit weird to me that `_computePairAddress` already takes the supposed pair address (target). I would either pass the tokens there, or rename that method to `_verifyPairAddress` (leaning towards the latter)

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -74,4 +82,28 @@ contract UniswapV2Executor is IExecutor {
...
+        bytes32 salt = keccak256(abi.encodePacked(token0, token1));
+        pair = address(
+            uint160(
+                uint256(
+                    keccak256(
+                        abi.encodePacked(
+                            hex"ff",
+                            FACTORY,
+                            salt,
+                            hex"96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f"
```
dianacarvalho1: what is this?

foundry/src/executors/UniswapV2Executor.sol
royvardhan: The function actually computes the pair address with token0 and token1 extracted from the supposed pair address.

For computing address, it needs both tokens, but we are only receiving tokenIn in this function, so we need the target to extract the other token and then do the computing.

foundry/test/mock/MockUniswapV2Pool.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Unlicense
```
dianacarvalho1: this is not really a mock. it's a fake uniswap v2 pool. We should name it accordingly please. Also I don't think it warrants its' own file tbh

dianacarvalho1 (commented):
Thank you @royvardhan ! Left a couple of questions.
Could you please share what you found to be the most challenging part of this task? It seems like this PR took the entire day to complete, was there a particular obstacle that caused the delay?

foundry/src/executors/UniswapV3Executor.sol
royvardhan: It is the keccak hash of the creation code of the pool, taken from [here](https://github.com/Uniswap/v3-periphery/blob/0682387198a24c7cd63566a2c58398533860a5d1/contracts/libraries/PoolAddress.sol#L6)

foundry/src/executors/UniswapV2Executor.sol
tamaralipows: Yes I see what the function does, I don't think it's wrong, I think my point still stands that it's strange

foundry/test/mock/MockUniswapV2Pool.sol
royvardhan: I can change the naming, but since we had MockERC20, I added this file to maintain consistency.

dianacarvalho1: >  I added this file to maintain consistency

I like consistency when it makes sense. These two use cases, MockERC20 and "Mock"UniswapV2Pool (which is not a mock but a fake) are not even in the same scope. It makes no sense to have consistency across them because they are completely unrelated.

royvardhan: So should I move this small contract inside the usv2 test contract?

foundry/src/executors/UniswapV2Executor.sol
royvardhan: https://github.com/propeller-heads/tycho-execution/pull/80#discussion_r1965985447

foundry/src/executors/UniswapV2Executor.sol
royvardhan: I am moving verification inside the function and then renaming it.

foundry/src/executors/UniswapV3Executor.sol
tamaralipows: Can you keep it as a constant at the top of this contract then so that we know, similar to how it's done in that lib. Else we'll be super confused coming back to this one day

royvardhan: yes yes!

foundry/test/mock/MockUniswapV2Pool.sol
tamaralipows: yes

royvardhan:
> Thank you @royvardhan ! Left a couple of questions. Could you please share what you found to be the most challenging part of this task? It seems like this PR took the entire day to complete, was there a particular obstacle that caused the delay?

I have been spending more time writing thoughtful PRs, to minimize extra efforts and time to address pr comment rounds. This is a win-win situation for both the reviewer and the author, opening a pr is easy but satisfactorily closing it is very tough and blocks from taking other tasks, until I close the open ones since the pr change requests keep coming in.

This task wasn't hard, but minor things keep me blocked from merging. Thought to improve on it.

Additionally, I got busy in company games, so there's a 90-minute pr delay there. And I won the poker game!

foundry/test/executors/UniswapV2Executor.t.sol
```diff
@@ -120,4 +144,16 @@ contract UniswapV2ExecutorTest is UniswapV2ExecutorExposed, Test, Constants {
         uint256 finalBalance = DAI.balanceOf(BOB);
         assertGe(finalBalance, amountOut);
     }
+
+    function test_RevertIf_Swap_InvalidTarget() public {
```
tamaralipows: why are there underscores in this test name? :(

foundry/test/executors/UniswapV3Executor.t.sol
```diff
@@ -113,6 +128,25 @@ contract UniswapV3ExecutorTest is Test, Constants {
         assertGe(IERC20(DAI_ADDR).balanceOf(address(this)), expAmountOut);
     }
 
+    function test_RevertIf_InvalidTarget() public {
```
tamaralipows: :(

royvardhan: It's a foundry test naming convention that I am following for the reverts: https://github.com/propeller-heads/tycho-execution/blob/2241c50df1b57a8160e1c2041598f504658b6fb9/foundry/test/LibPrefixLengthEncodedByteArray.t.sol#L54

tamaralipows: oh goodness, but we haven't been following this the whole repo, now our repo is inconsistent :( using this naming sometimes but not always. I think that's pretty bad. We should decide as a team to either use it for every test or never

royvardhan: Remember when the test for the lib was breaking? We had to adapt the revert if convention for the reverts test. 

Had to use this convention for reverts.

tamaralipows: There are countless other tests where we are not using this though. See `testSwapCheckedFailure` for just one of many examples. Is it not valuable to be consistent within our repo?

If yes, do you recommend we rename all those tests? I don't think inconsistency is a good thing here so we need to make a decision altogether which way to go...

royvardhan: As far as I remember [this](https://github.com/propeller-heads/tycho-execution/commit/69d03f060872bf9cce313f2420dc95c6d7554dec) fixed the breaking test issue

tamaralipows: I'm not asking that - I'm asking if you think we should go that way for all the tests to be consistent

royvardhan: Ahh so it was `testFail` tests that were failing after foundry update.

royvardhan: > There are countless other tests where we are not using this though. See `testSwapCheckedFailure` for just one of many examples. Is it not valuable to be consistent within our repo?
> 
> 
> 
> If yes, do you recommend we rename all those tests? I don't think inconsistency is a good thing here so we need to make a decision altogether which way to go...

It's just that I used the revert if convention for the reverts tests because it reminded me of our lib tests breaking after foundry update and Diana fixed it with the revert if convention.

royvardhan: > I'm not asking that - I'm asking if you think we should go that way for all the tests to be consistent

Yes, we should rename all our tests someday to follow the foundry convention: https://book.getfoundry.sh/guides/best-practices#general-test-guidance

tamaralipows (approved):
Thanks!

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -22,6 +29,9 @@ contract UniswapV2Executor is IExecutor {
         IERC20 tokenIn;
 
         (tokenIn, target, receiver, zeroForOne) = _decodeData(data);
+
+        _verifyPairAddress(target);
```
kayibal, line 33: Why do we do this here? Is there a problem calling some other contract with the v2 executor?

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -6,10 +6,17 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "@uniswap-v2/contracts/interfaces/IUniswapV2Pair.sol";
 
 error UniswapV2Executor__InvalidDataLength();
+error UniswapV2Executor__InvalidTarget();
 
 contract UniswapV2Executor is IExecutor {
     using SafeERC20 for IERC20;
 
+    bytes32 internal constant POOL_INIT_CODE_HASH =
+        0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f;
```
kayibal, line 15: This means sushiswap and other v2 clones can't be handled anymore...

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -146,4 +152,31 @@ contract UniswapV3Executor is IExecutor, ICallback {
...
+                            keccak256(abi.encode(token0, token1, fee)),
+                            POOL_INIT_CODE_HASH
+                        )
+                    )
+                )
+            )
+        );
+        if (pool != target) {
+            revert UniswapV3Executor__InvalidTarget();
+        }
```
kayibal, line 180: Isn't this the same logic that verifyCallback uses?

foundry/src/executors/UniswapV2Executor.sol
royvardhan: Yes, if an attacker can provide any target, they may have access to the `givenAmount` [here](https://github.com/propeller-heads/tycho-execution/blob/ceef68320288273d5a7f9a98f230d5ce7c7d8158/foundry/src/executors/UniswapV2Executor.sol#L45). The attacker can make the target a smart contract and make the swap succeed.

This can be used to steal tokens already owned by the Router.

foundry/src/executors/UniswapV2Executor.sol
royvardhan: Yes. We are using both Factory and the pool init hash to verify if the target is indeed a uni v2 pool.

foundry/src/executors/UniswapV3Executor.sol
royvardhan: Yes that is correct! Good catch! I will remove this function.

PR #82, opened 2025-02-24
feat: hardcode callback and swap selection in dispatcher

royvardhan:
**Router**
- Hardcoded `handleCallback` and `swap` selector in dispatcher, and updated integration tests for both Uniswap V3 and V4. 
- The router no longer parses swap selector in `_swap` function. 
- Selector parser function has also been removed from the LibSwap library.
 
**Encoder:**
The Swap and Strategy encoders are updated to not have the selectors in the encoded data. Also fixed the encoder tests.


foundry/test/ExecutionDispatcher.t.sol
dianacarvalho1, line 1: we didn't rename this test!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -328,11 +313,7 @@ impl StrategyEncoder for ExecutorStrategyEncoder {
         let executor_address = Bytes::from_str(swap_encoder.executor_address())
             .map_err(|_| EncodingError::FatalError("Invalid executor address".to_string()))?;
 
-        Ok((
-            grouped_protocol_data,
-            executor_address,
-            Some(swap_encoder.swap_selector().to_string()),
-        ))
+        Ok((grouped_protocol_data, executor_address))
     }
```
dianacarvalho1, line 317: @royvardhan I think it's still necessary in this case.. otherwise, how will the user know which one to use? I mean.. they could also hardcode it in their "Dispatcher"

src/encoding/evm/tycho_encoder.rs
```diff
@@ -118,7 +118,7 @@ impl TychoEncoder for EVMTychoEncoder {
                 value,
                 data: contract_interaction,
                 to: target_address,
-                selector,
+                selector: None,
```
dianacarvalho1: We could remove it from transaction as well no? (If we assume the user will hardcode it as well)

foundry/test/TychoRouter.t.sol
```diff
@@ -752,7 +706,7 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
...0000000cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000000000000000000000000000000000003b9aca000000000000000000000000000000000000000000000000000000000067e4237600000000000000000000000000000000000000000000000000000000000000000000000000000000000000003ede3eca2a72b3aecc820e955b36f38437d013950000000000000000000000000000000000000000000000000000000067bc9d7e00000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000280000000000000000000000000000000000000000000000000000000000000004166b5d3bb274c323e08eeba45d308cc9c11216f9aaafad2a22e94b94fec39293e5480f65f6238d7c8f1e8177f39118373e1041b0ab3a674d3041d119bdb6bc39c1b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008c008a0001000000f62849f9a0b5bf2913b396098f7c7019b51a820aa0b86991c6218b36c1d19d4a2e9eb0ce3606eb486982508145454ce325ddbe47a25d4ec3d231193300f62849f9a0b5bf2913b396098f7c7019b51a820a0000000000000000000000000000000000000000000bb800003c6982508145454ce325ddbe47a25d4ec3d23119330061a80001f40000000000000000000000000000000000000000"
```
dianacarvalho1, line 709: uuff where these ones a pain to update?

dianacarvalho1 (commented):
Thank you @royvardhan ! 🙏🏼 Only have a small small comment

src/encoding/evm/tycho_encoder.rs
royvardhan: Yes, I have removed it now after your suggestion

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Yes, they could hardcode it.

foundry/test/TychoRouter.t.sol
royvardhan: This was a bit time consuming, because even after removing the selector from the asserts, the tests were failing because of ple encode return changed due to data length. And also had to pad extra bytes wherever the calldata(bytes4) was removed.

royvardhan: Plus, these integration test expects the correct executor address

dianacarvalho1: yuupp

src/encoding/models.rs
```diff
@@ -96,7 +96,6 @@ pub struct Transaction {
     pub to: Bytes,
     pub value: BigUint,
     pub data: Vec<u8>,
-    pub selector: Option<String>,
```
dianacarvalho1, line 99: I think we need to update the gitbook docs now as well!

dianacarvalho1 (approved):
Thank you 🙏🏼 LGTM

src/encoding/models.rs
royvardhan: +100

PR #83, opened 2025-02-25
fix: Restrict receive callers to have code

dianacarvalho1:


royvardhan (approved):

tamaralipows (approved):
For our future reference this only costs 100 additional gas

PR #85, opened 2025-02-25
feat: support base deployment

tamaralipows:
gas limit had to be set when setting executors on base: kept getting UNPREDICTABLE_GAS_LIMIT error. It should definitely not cost more than 100000 gas to set limit... but this may be a problem on other chains in the future.


foundry/scripts/roles.json
```diff
@@ -1,15 +1,32 @@
...
+    ],
+    "FEE_SETTER_ROLE": [],
+    "PAUSER_ROLE": [
+      "0xb0A77f867Fcec1e9b271Ee17354bC6bBC0dD5662"
+    ],
+    "UNPAUSER_ROLE": [
+      "0xb0A77f867Fcec1e9b271Ee17354bC6bBC0dD5662"
+    ],
+    "FUND_RESCUER_ROLE": [
+      "0xb0A77f867Fcec1e9b271Ee17354bC6bBC0dD5662"
```
tamaralipows, line 44: These shall be changed eventually... this was just for my test deployment.

foundry/hardhat.config.js
```diff
@@ -13,13 +13,21 @@ module.exports = {
     },
 
     networks: {
-        tenderly: {
+        tenderly_mainnet: {
+            url: process.env.RPC_URL,
+            accounts: [process.env.PRIVATE_KEY]
+        },
+        tenderly_base: {
```
tamaralipows, line 24: I hate that this is an underscore but it doesn't accept a hyphen

foundry/scripts/roles.json
```diff
@@ -1,15 +1,62 @@
...
+      "0x58Dc7Bf9eD1f4890A7505D5bE4E4252978eAF655"
+    ],
+    "UNPAUSER_ROLE": [
+      "0x58Dc7Bf9eD1f4890A7505D5bE4E4252978eAF655"
+    ],
+    "FUND_RESCUER_ROLE": [
+      "0x58Dc7Bf9eD1f4890A7505D5bE4E4252978eAF655"
+    ]
+  },
+  "tenderly_mainnet": {
```
tamaralipows: I don't love that these are repeated for tenderly. Maybe tomorrow I'll check if the network name has tenderly and if yes remove "tenderly_" from it to find the role. Too much for today though. lmk what you think

foundry/scripts/README.md
```diff
@@ -5,23 +5,24 @@
...
-1. Make a new [fork](https://dashboard.tenderly.co/) in tenderly dashboard.
+1. Make a new [fork](https://dashboard.tenderly.co/) in tenderly dashboard for the 
+   chain that you wish to deploy on.
 2. Set the following environment variables:
 
 ```
-export RPC_URL=<fork-rpc-from-tenderly>
-export DEPLOY_WALLET=<wallet-address>
-export PRIVATE_KEY=<private-key>
+export TENDERLY_RPC_URL=https://rpc.tenderly.co/fork/de5f123f-1c95-412b-97c1-ddd64bb2ff42
```
royvardhan: is this intentional?

foundry/scripts/roles.json
royvardhan: I have no problem with `base` and `tenderly_base`

royvardhan (approved):
Hey! This looks good to me, thanks!

foundry/hardhat.config.js
```diff
@@ -17,13 +17,21 @@ module.exports = {
     },
 
     networks: {
-        tenderly: {
-            url: process.env.RPC_URL,
+        tenderly_mainnet: {
+            url: process.env.TENDERLY_RPC_URL,
```
dianacarvalho1: but won't this mean that even when you want to deploy to mainnet you need to define the `TENDERLY_RPC_URL`? why did you make them a different env var and not just RPC_URL for everyone?

foundry/hardhat.config.js
```diff
@@ -17,13 +17,21 @@ module.exports = {
...
+            accounts: [process.env.PRIVATE_KEY]
+        },
+        tenderly_base: {
+            url: process.env.TENDERLY_RPC_URL,
             accounts: [process.env.PRIVATE_KEY]
         },
         mainnet: {
```
dianacarvalho1: I think we will have to change this to be "ethereum" to match what is expected in the executors file (which is what tycho-indexer gives us)

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I only have small concerns (one that I introduced myself 😅 sooorryy)

foundry/scripts/README.md
tamaralipows: upsi. it's a burner wallet though, i'll toss it out

foundry/hardhat.config.js
tamaralipows: i could have sworn it was set to TENDERLY_RPC_URL somewhere else before this change. maybe I hallucinated

tamaralipows: Oh. It was in `fund_tenderly_fork.js`. I'll change it in there to just `RPC_URL` then.

PR #87, opened 2025-02-26
refactor: use swap core logic in swapPermit2

royvardhan:
Since `swapPermit2` uses all the core logic of `swap`, I have refactored it to call `swap` after using permit transfer. Addresses the risk of inconsistencies that could occur when one function is modified while the other is forgotten 👍

Also removed `nonReentrant` modifier from `swapPermit2` since `swap` already has it, and if `swap` reverts then everything reverts.

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -151,37 +151,16 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
             tokenIn = address(_weth);
         }
 
-        amountOut = _swap(amountIn, nTokens, swaps);
-
-        if (fee > 0) {
-            uint256 feeAmount = (amountOut * fee) / 10000;
-            amountOut -= feeAmount;
-            IERC20(tokenOut).safeTransfer(feeReceiver, feeAmount);
-        }
-
-        if (minAmountOut > 0 && amountOut < minAmountOut) {
-            revert TychoRouter__NegativeSlippage(amountOut, minAmountOut);
-        }
-
-        uint256 leftoverAmountIn;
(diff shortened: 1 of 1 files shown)


dianacarvalho1 (approved):
beeeerryy niceeee! 🙏🏼

dianacarvalho1 (commented):
I would just move the _execute_swap definition a bit down. so that we have both of our public methods first

royvardhan:
> I would just move the _execute_swap definition a bit down. so that we have both of our public methods first

hmmm, both the public methods are first in the contract.

dianacarvalho1:
ah true sorry! I got confused with the diff! all good!

royvardhan:
> ah true sorry! I got confused with the diff! all good!

All good, thanks! 🙏

foundry/src/TychoRouter.sol
```diff
@@ -242,6 +205,47 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
+        bool wrapEth,
+        bool unwrapEth,
+        uint256 nTokens,
+        address receiver,
+        bytes calldata swaps
+    ) internal returns (uint256 amountOut) {
+        if (receiver == address(0)) {
+            revert TychoRouter__AddressZero();
+        }
+        // For native ETH, assume funds are already in the router.
```
tamaralipows: Maybe we can move this comment up to where the permit2 transfer is? Seems a bit misplaced here now

tamaralipows (approved):
Thank you! Let's just wait for @dianacarvalho1 to give another green light since it changed significantly

dianacarvalho1 (approved):
Thanks! LGTM

PR #88, opened 2025-02-26
refactor: check if amountIn was fully consumed based on balance changes

royvardhan:
We were doing an absolute check of `leftoverAmount` ignoring the fact that there could be dust amount of `tokenIn` in the router before the swap. This could lead to unexpected reverts.

This pr checks if `amountIn` was fully consumed based on balance changes


dianacarvalho1 (approved):
Thank you! This looks good to me! 👍🏼

tamaralipows (approved):
Thank you! Very good!

PR #89, opened 2025-02-26
refactor: remove unnecessary checks from _unwrapEth

royvardhan:
Remove unnecessary checks from _unwrapEth, and only check the amount

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -441,9 +441,6 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
      * @param amount of WETH to unwrap.
      */
     function _unwrapETH(uint256 amount) internal {
-        if (amount == 0) {
-            revert TychoRouter__AmountZero();
-        }
         _weth.withdraw(amount);
     }
 
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -56,6 +56,7 @@ import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
 //           ✷✷✷✷✷✷               ✷✷✷✷✷              ✷✷✷✷✷✷✷✷        ✷✷✷✷✷✷      ✷✷✷✷✷✷         ✷✷✷✷✷✷✷✷
 
 error TychoRouter__AddressZero();
+error TychoRouter__AmountZero();
 error TychoRouter__EmptySwaps();
 error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
 error TychoRouter__AmountInNotFullySpent(uint256 leftoverAmount);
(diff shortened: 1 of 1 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -464,13 +465,14 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
-        _weth.withdraw(unwrapAmount);
+        if (amount == 0) {
+            revert TychoRouter__AmountZero();
+        }
+        _weth.withdraw(amount);
     }
 
     /**
-     * @dev Allows this contract to receive native token
+     * @dev Allows this contract to receive native token from contracts
```
tamaralipows: This in the end is a bit misleading since the contract can still receive native token as long as the calldata is not empty.

Maybe specify "with empty calldata"? i dunno

tamaralipows: (I know you didnt write this btw)

royvardhan: It checks if the msg.sender is contract [here](https://github.com/propeller-heads/tycho-execution/blob/22c48e80d279ccfb43c2a8c51fea4f93c05ba6c8/foundry/src/TychoRouter.sol#L476)

tamaralipows: Yes I know, I wasn't talking about that part.
I was talking about the fact that native token can still be received from contracts even without this comment, this comment implies that it can't.

tamaralipows:
```suggestion
     * @dev Allows this contract to receive native token with empty msg.data from contracts
```

royvardhan: Ahh yes got it now

royvardhan: committing away!

foundry/src/TychoRouter.sol
```diff
@@ -464,13 +465,14 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
      * @param amount of WETH to unwrap.
      */
     function _unwrapETH(uint256 amount) internal {
-        uint256 unwrapAmount =
-            amount == 0 ? _weth.balanceOf(address(this)) : amount;
-        _weth.withdraw(unwrapAmount);
+        if (amount == 0) {
+            revert TychoRouter__AmountZero();
+        }
```
dianacarvalho1: why do we need to revert if it's zero? did I miss any thread?

royvardhan: I add it myself, just to revert for 0 amountOuts

dianacarvalho1: I think it's unnecessary. what happens if it's truly zero?

royvardhan: if it's truly zero then `weth` contract calls `msg.sender.transfer(0)`

dianacarvalho1: and then what happens? or what is the problem with that?

royvardhan: when `msg.sender.transfer(0)` is called, nothing happens. The call goes fine with no revert or whatsoever. I was thinking of the case when the swap happened, amountIn was consumed but amountOut came out to be zero, I know it is not possible but still was thinking of edge cases.

royvardhan: I can remove it though! :)

dianacarvalho1: from the meeting: I understand the check you are doing but then it shouldn't be done here.

dianacarvalho1 (approved):
Thank you

PR #93, opened 2025-02-28
fix: make USV2 factory configurable in Executor

tamaralipows:
- This factory is not the same for Ethereum and Base, so Base txs were failing when verifying pool addresses.
- I've double checked that we don't have this problem for Balancer V2 - the vault address in the same on Base and on Ethereum Mainnet.


royvardhan (approved):

config/executor_addresses.json
```diff
@@ -12,7 +12,10 @@
     "vm:balancer_v2": "0x543778987b293C7E8Cf0722BB2e935ba6f4068D4"
   },
   "base": {
-    "uniswap_v2": "0x2DaE954eCF457276F90B68Cee68981C0aA07f2ef"
+    "uniswap_v2": "0xa2D6c55676D0d2A1D090e3157aCC9a41bB53BE3B",
+    "uniswap_v3": "0x374fFd7224422dF5dE83B43C7Fa8379F9CAA826a",
+    "uniswap_v4": "0xf1531642018C9Ab6a1B07666755d975eae16a01b",
+    "vm:balancer_v2": "0x6920209CaAF45872006eC507f2f18595af030418"
```
tamaralipows, line 18: Yes I actually deployed these to base but I didn't verify them on basescan because I was too lazy to fetch a basescan API key. Problem for tomorrow

dianacarvalho1 (approved):
oh perfect! Great catch @tamaralipows 🙏🏼 thank you

PR #95, opened 2025-02-28
feat: Rename ETH_RPC_URL -> RPC_URL

tamaralipows:
We support multiple chains so this may not just be ETH, but perhaps a Base URL for example.


dianacarvalho1 (approved):
merciii

PR #98, opened 2025-03-04
fix: remove redundant target verification in USV3 executor

royvardhan:
Removed the target verification from the Uniswap V3 executor because we are already doing it [here](https://github.com/Uniswap/v3-periphery/blob/0682387198a24c7cd63566a2c58398533860a5d1/contracts/libraries/CallbackValidation.sol#L33) in the `verifyCallback`.


royvardhan:
@tamaralipows shall I close this? Since we already discussed the new attack scenario.

dianacarvalho1 (commented):
I think we should close this PR for now. Later if we decide to remove the verification we can come back to it

PR #99, opened 2025-03-04
fix: inequality check for amountConsumed and amountIn

royvardhan:
Checks if `amountConsumed` is not equal to `amountIn`, reverts with `amountConsumed`

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -59,7 +59,7 @@ error TychoRouter__AddressZero();
 error TychoRouter__AmountZero();
 error TychoRouter__EmptySwaps();
 error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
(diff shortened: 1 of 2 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -59,7 +59,7 @@ error TychoRouter__AddressZero();
 error TychoRouter__AmountZero();
 error TychoRouter__EmptySwaps();
 error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
-error TychoRouter__AmountInNotFullySpent(uint256 leftoverAmount);
+error TychoRouter__AmountInDiffersFromConsumed(uint256 amountConsumed);
```
tamaralipows: Perhaps we can also return the amountIn here actually

royvardhan: Yes sure!

tamaralipows (approved):
Thanks!

PR #100, opened 2025-03-05
feat: add safeTransferFrom in swap and move core swap logic inside _swapChecked

royvardhan:
Moved the common swap logic inside `_swapChecked` and added `safeTransferFrom` function inside `swap`

The router does not assume that the `tokenIn` are in the router anymore.

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -143,7 +143,6 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
         address receiver,
         bytes calldata swaps
     ) public payable whenNotPaused nonReentrant returns (uint256 amountOut) {
-        // Transfer the tokenIn token to the router
         IERC20(tokenIn).safeTransferFrom(msg.sender, address(this), amountIn);
         return _swapChecked(
             amountIn,
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -56,7 +56,6 @@ import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
 //           ✷✷✷✷✷✷               ✷✷✷✷✷              ✷✷✷✷✷✷✷✷        ✷✷✷✷✷✷      ✷✷✷✷✷✷         ✷✷✷✷✷✷✷✷
 
 error TychoRouter__AddressZero();
-error TychoRouter__AmountZero();
 error TychoRouter__EmptySwaps();
 error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
 error TychoRouter__AmountInDiffersFromConsumed(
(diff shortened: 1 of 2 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -144,51 +143,19 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
-
-        if (unwrapEth) {
-            _unwrapETH(amountOut);
-        }
-        if (tokenOut == address(0)) {
-            Address.sendValue(payable(receiver), amountOut);
-        } else {
-            IERC20(tokenOut).safeTransfer(receiver, amountOut);
-        }
+        // Transfer the tokenIn token to the router
```
tamaralipows: Unnecessary comment I think
```suggestion

```

tamaralipows (approved):
Thanks, just a minor comment!

foundry/test/TychoRouter.t.sol
```diff
@@ -457,6 +457,43 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        );
+
+        bytes memory swap = encodeSwap(
+            uint8(0), uint8(1), uint24(0), address(usv2Executor), protocolData
+        );
+        bytes[] memory swaps = new bytes[](1);
+        swaps[0] = swap;
+
+        uint256 minAmountOut = 2600 * 1e18;
+        vm.expectRevert();
```
dianacarvalho1, line 481: can you specify which error you expect here please?

foundry/test/TychoRouter.t.sol
```diff
@@ -457,6 +457,43 @@ contract TychoRouterTest is TychoRouterTestSetup {
         vm.stopPrank();
     }
 
+    function testSwapCheckedLessApprovalFailure() public {
```
dianacarvalho1, line 460: why is this test necessary now? there was already a `testSwapCheckedFailure`

dianacarvalho1 (commented):
Thank you @royvardhan ! I have two questions

foundry/test/TychoRouter.t.sol
royvardhan: It should be [this error](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/fa995ef1fe66e1447783cb6038470aba23a6343f/contracts/token/ERC20/utils/SafeERC20.sol#L22) with which it should revert, I had added it. But then the test complains that expected revert error and actual revert error does not match. Because it reverts with our favourite error `EvmError: Revert`.
image

foundry/test/TychoRouter.t.sol
royvardhan: `testSwapCheckedFailure` is for the `minAmountOut` check and it reverts with `TychoRouter__NegativeSlippage` when actual `amountOut` does not match `minAmountOut`. While this new test reverts because not sufficient allowance was granted by the user

dianacarvalho1: oh that is not evident to me from the naming. can you improve it a bit or add a comment please?

foundry/test/TychoRouter.t.sol
```diff
@@ -457,6 +457,43 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        // Trade 1 WETH for DAI with 1 swap on Uniswap V2
+        // Checks amount out at the end
+        uint256 amountIn = 1 ether;
+
+        // Approve less than the amountIn
```
dianacarvalho1: it's strange that the comment is a few lines higher than the line of code it refers to. (the same happens in the other test as well)

foundry/test/TychoRouter.t.sol
dianacarvalho1: huuumm that is weird.. you should get that error 🤔

foundry/test/TychoRouter.t.sol
```diff
@@ -457,6 +457,43 @@ contract TychoRouterTest is TychoRouterTestSetup {
         vm.stopPrank();
     }
 
+    function testSwapCheckedLessApprovalFailure() public {
+        // Trade 1 WETH for DAI with 1 swap on Uniswap V2
+        // Checks amount out at the end
```
dianacarvalho1: it doesn't check amount out at the end because it reverts

royvardhan: that is copy pasta issue. Removing it

foundry/test/TychoRouter.t.sol
royvardhan: yeahhhhh

tamaralipows: this is indeed bizarre

foundry/test/TychoRouter.t.sol
royvardhan: yesss

foundry/test/TychoRouter.t.sol
tamaralipows: Just double checked this - this is expected. `SafeERC20FailedOperation` is only thrown if the token contract doesn't provide any data. In this case, I guess the token contract is providing us useless data (EVMError)

foundry/test/TychoRouter.t.sol
dianacarvalho1: 🍝 🍝 🍝 🍝

dianacarvalho1 (approved):
Loooks goood! Thank you 🙏🏼

PR #101, opened 2025-03-05
feat: Check min amount out is not zero

dianacarvalho1:
Update tests


tamaralipows (approved):
Yeehaw LFG

tamaralipows (commented):
Wait wait. How about just one test with a 0 min amount and expect revert? Should be an easy copypasta

foundry/src/TychoRouter.sol
```diff
@@ -243,6 +244,10 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
         if (receiver == address(0)) {
             revert TychoRouter__AddressZero();
         }
+        if (minAmountOut == 0) {
```
royvardhan, line 247: can we not unify this with the above check and make them like
```solidity
if (receiver == address(0) || minAmountOut == 0) 
{
    revert TychoRouter__AddressOrMinAmountOutZero();
}
```

royvardhan (approved):
Looks good! Just a question.

foundry/src/TychoRouter.sol
dianacarvalho1: I think the gas usage should be very very similar. I prefer to have a more descriptive error tbh.

royvardhan: Okay cool!

PR #102, opened 2025-03-05
feat: Support cyclic swaps

royvardhan:
While adding a cyclic swap test, I discovered that we are adding the `amountIn` to `amountOut` when `tokenInIndex` and `tokenOutIndex` are same. I added a `cyclicSwapAmountOut` tracker to track `amountOut` in case of cyclic swap, and added both sequential and split swap tests that circles back to USDC(the start token of the swap)


foundry/test/TychoRouter.t.sol
```diff
@@ -1192,6 +1192,127 @@ contract TychoRouterTest is TychoRouterTestSetup {
         assertEq(IERC20(WBTC_ADDR).balanceOf(tychoRouterAddr), 102718);
     }
 
+    function testCyclicSequentialSwap() public {
+        // This test has start and end tokens that are the same
+        // The flow is:
+        // USDC -> WETH -> USDC -> WETH -> USDC using two pools, and four swaps
```
tamaralipows: can't we just simplify this to have 2 swaps? haha this seems complicated unnecessarily to achieve the purpose of knowing if we can start and end on the same token in a sequence?

foundry/src/TychoRouter.sol
```diff
@@ -345,7 +345,9 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
             currentAmountOut = _callExecutor(
                 swapData.executor(), currentAmountIn, swapData.protocolData()
             );
-            amounts[tokenOutIndex] += currentAmountOut;
+            amounts[tokenOutIndex] = tokenOutIndex == 0
```
tamaralipows: I'm not sure it would work in the scenario where we have a split like this for the output token. Can you double check, please? Imagine WETH - USDC - WETH but we split on the output

```
               ┌─ WETH
WETH ── USDC ──┤
               └─ WETH

**Review by tamaralipows — CHANGES_REQUESTED** [review]
Just one major comment, the other one more minor. Please have a look!

File: foundry/test/TychoRouter.t.sol (hunk 1 above)
Comment (dianacarvalho1) [reply foundry/test/TychoRouter.t.sol]: oh yes.. I think it doesn't make sense to pass through USDC three times 😕

**Review by dianacarvalho1 — COMMENTED** [review]
thank you @royvardhan ! Can we improve the PR name a bit please? the first part of the name doesn't mean much without more context. it should be a feat and something like: feat: Support cyclic swaps

File: foundry/test/TychoRouter.t.sol (hunk 1 above)
Comment (royvardhan) [reply foundry/test/TychoRouter.t.sol]: Done, removed them

File: foundry/src/TychoRouter.sol (hunk 2 above)
Comment (royvardhan) [reply foundry/src/TychoRouter.sol]: Thank you for this, updated `_swap` now.

File: foundry/test/TychoRouter.t.sol [hunk 3]
```diff
@@ -1192,6 +1192,162 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+
+    function testSplitOutputCyclicSwap() public {
+        // This test has start and end tokens that are the same
+        // The flow is:
+        //           ┌─── WETH (Pool 2, 60% split) ───┐
+        //           │                                │
+        // USDC ----─┤                                ├─> USDC
+        //           │                                │
+        //           └─── WETH (Pool 2, the rest)  ───┘
+        //
```
tamaralipows: This looks really similar to the previous diagram, it's hard to tell the difference. I think the pool should go on the lines and not directly next to the token perhaps?

tamaralipows (commented):
Thanks! It's looking good. Just a little request.

foundry/test/TychoRouter.t.sol
royvardhan: Okay, let's see what I can make.

foundry/src/TychoRouter.sol
```diff
@@ -329,6 +329,7 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
 
         uint256[] memory remainingAmounts = new uint256[](nTokens);
         uint256[] memory amounts = new uint256[](nTokens);
+        uint256 cyclicSwapAmountOut = 0;
```
dianacarvalho1, line 335: can we add some docs on why this is necessary please? it's not obvious

dianacarvalho1 (approved):
Thank you 🙏🏼

foundry/test/TychoRouter.t.sol
```diff
@@ -1192,6 +1192,161 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+        );
+
+        // WETH -> USDC
+        swaps[3] = encodeSwap(
+            uint8(1),
+            uint8(0),
+            uint24(0),
+            address(usv3Executor),
+            usdcWethV3Pool1OneZeroData
+        );
```
tamaralipows: Looking at this with a fresh brain, I think this is wrong

I think swap index 1 shouldn't exist.

The two swaps from USDC to WETH are right beside each other, and the third swap will take the entirety of the out amount of both pools and swap it all at once back to USDC, does that make sense?

royvardhan: Thanks for spotting this, i have changed it now.

foundry/test/TychoRouter.t.sol
```diff
@@ -1192,6 +1192,146 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+    }
+
+    function testSplitInputCyclicSwap() public {
+        // This test has start and end tokens that are the same
+        // The flow is:
+        //            ┌─── 60% ──> WETH ───┐
+        //            │                    │
+        // USDC ──────┤                    ├──> USDC
+        //            │                    │
+        //            └─── 40% ──> WETH ───┘
```
tamaralipows:
```suggestion
        //            ┌─ (USV3, 60% split) ──> WETH ─┐
        //            │                              │
        // USDC ──────┤                              ├──(USV2)──> USDC
        //            │                              │
        //            └─ (USV3, 40% split) ──> WETH ─┘
```

tamaralipows (approved):
Thanks!

PR #103, opened 2025-03-05
feat: enforce checked amount when encoding to router

tamaralipows:
- If using the tycho router, you must either set the slippage with the expected amount in the solution, or you must set the checked amount
- Decided to put this in the split swap strategy validator, since this is not relevant when it comes to direct execution (no checks are performed inside the executors)


royvardhan (approved):
Thanks!🙏🏻

dianacarvalho1 (approved):
veeery goooodd!

PR #106, opened 2025-03-07
[DO NOT MERGE] BatchExecute ActionType tests

tamaralipows:
Find out if it's much more expensive to use the ActionType pattern than to use a simple bool

TODO:
- Write the same test for passing a bool instead of using ActionType... this part should be much faster.


tamaralipows:
Closing since we are going another route to perform transfers from the user to the first pool. This can be resurrected when we would like to support batch execution.

tamaralipows:
Closing since we are going another route to perform transfers from the user to the first pool. This can be resurrected when we would like to support batch execution.

PR #108, opened 2025-03-10
feat: support cyclic swap encoding

royvardhan:
- Updated the `validate_swap_path`  to allow Cyclic swap encoding.
- Added Integration tests for Sequential, SplitInput, SplitOutput cyclic swaps.


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -206,6 +206,11 @@ impl StrategyEncoder for SplitSwapStrategyEncoder {
         }
 
         let encoded_swaps = self.ple_encode(swaps);
+        let tokens_len = if solution.given_token == solution.checked_token {
+            tokens.len() - 1
+        } else {
+            tokens.len()
+        };
```
tamaralipows, line 213: There's one thing that I am unsure about:

Imagine a WETH -> ETH non-cyclical swap with an unwrap.

However, because of this line, it will be considered cyclical:

```
        if unwrap {
            tokens.push(self.wrapped_address.clone());
        } else {
            tokens.push(solution.checked_token.clone());
        }
```

Is this problematic? Or should it be considered cyclical? (It might be okay - can you just double check my concern please?)

On the contrary, what about a cyclical swap where ETH -> ETH with unwrapping is considered?

Perhaps it would be good to add an integration test for these scenarios?

tamaralipows (commented):
I just have one concern that I am not sure is a valid concern - please double check. 

Also - we like to keep our commit headers to less than 70 characters for ease of reviewing. We then follow up with more detailed information after a blank line.

Could you set up a warning for this in your RustRover?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Yeah, thanks for this scenario. I will do an integration for this.

royvardhan:
> I just have one concern that I am not sure is a valid concern - please double check. 
> 
> 
> 
> Also - we like to keep our commit headers to less than 70 characters for ease of reviewing. We then follow up with more detailed information after a blank line.
> 
> 
> 
> Could you set up a warning for this in your RustRover?

I don't use RustRover but is this what you suggest?: 


```
Fix cyclic swaps integration tests

- Detailed information one
- Another detailed information two
- One more information three
```

dianacarvalho1:
> > I just have one concern that I am not sure is a valid concern - please double check.
> > Also - we like to keep our commit headers to less than 70 characters for ease of reviewing. We then follow up with more detailed information after a blank line.
> > Could you set up a warning for this in your RustRover?
> 
> I don't use RustRover but is this what you suggest?:
> 
> ```
> Fix cyclic swaps integration tests
> 
> - Detailed information one
> - Another detailed information two
> - One more information three
> ```

yes that's what @tamaralipows meant! I'm sure you can add a warning in your IDE as well though

foundry/test/TychoRouter.t.sol
```diff
@@ -1332,6 +1332,63 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
+    function testCyclicSequentialSwapIntegration() public {
+        deal(USDC_ADDR, ALICE, 100 * 10 ** 6);
+
+        // Approve permit2
+        vm.startPrank(ALICE);
+        IERC20(USDC_ADDR).approve(PERMIT2_ADDRESS, type(uint256).max);
+        // Encoded solution generated using `test_cyclic_sequential_swap`
+        // but manually replacing the executor address
+        // `dD8559c917393FC8DD2b4dD289c52Ff445fDE1B0` with the one in this test
+        // `2e234DAe75C793f67A35089C9d99245E1C58470b`
```
dianacarvalho1: the addresses should be lower case I think

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -1212,4 +1217,306 @@ mod tests {
...
+            sender: Bytes::from_str("0xcd09f75E2BF2A4d11F3AB23f1389FcC1621c0cc2").unwrap(),
+            receiver: Bytes::from_str("0xcd09f75E2BF2A4d11F3AB23f1389FcC1621c0cc2").unwrap(),
+            ..Default::default()
+        };
+
+        let (calldata, _) = encoder
+            .encode_strategy(solution)
+            .unwrap();
+
+        println!("{}", hex::encode(&calldata));
```
dianacarvalho1: could we assert something here please? to make sure that it worked as expected? if we don't, then we might change the original code in a way that breaks this logic and we won't notice at all. that would be a suuuper hard bug to find tbh

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -1212,4 +1217,306 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_split_output_cyclic_swap() {
+        // This test has start and end tokens that are the same
+        // The flow is:
+        //                        ┌─── (USV3, 60% split) ───┐
```
dianacarvalho1, line 1510: I'm confused about this drawing. which token is the upper split?

dianacarvalho1: ah I think it should be something like

````

        //                                                          ┌─── (USV3, 60% split) ──> USDC
        //                                                            │                        
        // USDC ──(USV2) ── WETH ──
        //                                                            │
        //                                                           └─── (USV3 Pool 2, 40%s split) ──> USDC

````
right??
edit: goddamit this formating is annoying

dianacarvalho1 (commented):
Thank you @royvardhan ! I have a few requests

src/encoding/evm/strategy_encoder/strategy_encoders.rs
royvardhan: Agreed!

dianacarvalho1 (approved):
Thank you @royvardhan ! It looks good to me

tamaralipows (approved):
Thanks!

PR #111, opened 2025-03-11
feat: TokenTransfer helper contract

tamaralipows:
### Uniswap V2
- Supports **transfer**, **transferFrom**, **permit2 transferFrom**, and **no transfer**.
### Uniswap V3
- Supports **transfer**, **transferFrom**, and **permit2 transferFrom**.

Unfortunately cannot handle the "no transfer" situation (i.e. if the previous pool would like to send directly to the next USV3 pool. The transfer must be done in the callback.

This is because it checks the balance right before the callback and uses this as the “starting point”, so if a transfer happens before even calling swap, it won't be considered.
```
        if (zeroForOne) {
            if (amount1 < 0) TransferHelper.safeTransfer(token1, recipient, uint256(-amount1));

            uint256 balance0Before = balance0();
            IUniswapV3SwapCallback(msg.sender).uniswapV3SwapCallback(amount0, amount1, data);
            require(balance0Before.add(uint256(amount0)) <= balance0(), 'IIA');
        }
```

### Uniswap V4 - implement transient storage solution in separate PR

In order to support different ways of in-transfers, we would need to implement it this override method:
```
    function _pay(Currency token, address, uint256 amount) internal override {
        IERC20(Currency.unwrap(token)).safeTransfer(
            address(poolManager), amount
        );
    }
```
This method does not allow us to pass it the transfer type. Thus - we can't know which transfer type to use, unless we perhaps use transient storage here to store it in the executor contract. 

### Ekubo - implement transient storage solution in separate PR

Similar case as USV4: Ekubo is also limited as to what info sent in the callback. Remember - this is why we needed to have a special case for Ekubo to be able to know which executor to call. There is no easy way to fit the transfer method in here (as far as I understand).

### Balancer V2
We are not allowed to transfer funds from anywhere other than `msg.sender`. This is ensured during authentication when calling the swap method on the Balancer Vault. Unfortunately for us - `msg.sender` will always be our router contract.
```
    function swap(
        SingleSwap memory singleSwap,
        FundManagement memory funds,
        uint256 limit,
        uint256 deadline
    )
        external
        payable
        override
        nonReentrant
        whenNotPaused
        authenticateFor(funds.sender) <------ it doesn't allow this to be anything other than msg.sender
        returns (uint256 amountCalculated)
    {
```

### Curve
Curve only performs `transferFrom` `msg.sender`. It seems there is no way to override that or pass it any other address to do so (or even to not perform this `transferFrom`).


foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -32,16 +35,19 @@ contract UniswapV2Executor is IExecutor {
...
 
-        (tokenIn, target, receiver, zeroForOne) = _decodeData(data);
+        (tokenIn, target, receiver, zeroForOne, method) = _decodeData(data);
 
         _verifyPairAddress(target);
 
         calculatedAmount = _getAmountOut(target, givenAmount, zeroForOne);
+        _transfer(tokenIn, target, givenAmount, method);
+
         tokenIn.safeTransfer(target, givenAmount);
```
royvardhan: Was just checking, I think you forgot to remove this transfer after adding the new one above.

tamaralipows: Haha yes I caught that later but forgot to push. Thanks a lot!

foundry/src/executors/ExecutorTransferMethods.sol
```diff
@@ -0,0 +1,54 @@
...
+        } else if (method == TransferMethod.TRANSFERFROM) {
+            // slither-disable-next-line arbitrary-send-erc20
+            tokenIn.safeTransferFrom(sender, receiver, amount);
+        } else if (method == TransferMethod.TRANSFERPERMIT2) {
+            // Permit2.permit is already called from the TychoRouter
+            permit2.transferFrom(
+                sender, receiver, uint160(amount), address(tokenIn)
+            );
+        } else {
+            // Funds are likely already in pool. Do nothing.
```
dianacarvalho1: likely or should be in the pool for real? 😅

foundry/src/executors/ExecutorTransferMethods.sol
```diff
@@ -0,0 +1,54 @@
...
+import "@permit2/src/interfaces/IAllowanceTransfer.sol";
+
+error ExecutorTransferMethods__InvalidPermit2();
+
+contract ExecutorTransferMethods {
+    using SafeERC20 for IERC20;
+
+    IAllowanceTransfer public immutable permit2;
+
+    enum TransferMethod {
```
dianacarvalho1: maybe `TransferType`?

dianacarvalho1: I feel that method reminds more of a function/method and not a methodology to do the transfer

foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -35,17 +38,18 @@ contract UniswapV2Executor is IExecutor {
         payable
         returns (uint256 calculatedAmount)
     {
+        IERC20 tokenIn;
         address target;
         address receiver;
         bool zeroForOne;
-        IERC20 tokenIn;
+        TransferMethod method;
```
dianacarvalho1: calling the variable just `method` is tooo general for me 😕

foundry/test/executors/UniswapV2Executor.t.sol
```diff
@@ -140,24 +168,104 @@ contract UniswapV2ExecutorTest is Test, Constants {
...
+        vm.startPrank(ALICE);
+        (
+            IAllowanceTransfer.PermitSingle memory permitSingle,
+            bytes memory signature
+        ) = handlePermit2Approval(
+            WETH_ADDR, address(uniswapV2Exposed), amountIn
+        );
+
+        // Assume the permit2.approve method will be called from the TychoRouter
+        // Replicate this secnario in this test.
```
dianacarvalho1:
```suggestion
        // Replicate this scenario in this test.
```

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -14,6 +14,15 @@ use crate::encoding::{
     swap_encoder::SwapEncoder,
 };
 
+#[allow(dead_code)]
+#[repr(u8)]
+pub enum TransferType {
```
dianacarvalho1, line 27: ah here it's TransferType 😅

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -54,6 +63,7 @@ impl SwapEncoder for UniswapV2SwapEncoder {
             component_id,
             bytes_to_address(&encoding_context.receiver)?,
             zero_to_one,
+            (TransferType::Transfer as u8).to_be_bytes(),
```
dianacarvalho1, line 78: this will be changed in upcoming PRs right??

dianacarvalho1 (commented):
Thank you sooo much @tamaralipows 🙏🏼 this was a biiigg boooiii
I have a few nitpick comments and then a couple of big ones:
1. What about ETH transfers? it doesn't really matter at this point because none of the protocols that support this trade ETH right?
2. I think it is worth it to try to use transient storage for the uniswap v4 case (maybe even for ekubo) - but this can be done in a separate PR though!

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: yesss I've been going back and forth between whether I like type or method better lol. I thought type was too generic but you made a good point of it reminding you of a function call. I'll change those!

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: Oui. I already have a PR open changing it

tamaralipows:
@dianacarvalho1 As far as I understand, we can't support ETH `transferFrom` from the user into the first pool, since this is sent as a `{value: x}` when calling the swap method of our contract. We would need set up an additional transaction to send ETH into the first pool - which wouldn't work for chains like `Base` where flashbots is not supported. There is no ETH equivalent of `transferFrom` afaik.

I will add an extra check in the encoding PR to ensure we are not attempting to optimize this case for ETH.

The transfers between pools of the same solution (for example from USV4 ----(ETH)---> Ekubo) is also something to think about though - we should look into this when implementing the transient storage solution (in a separate PR like you said).

foundry/src/executors/TokenTransfer.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.26;
+
+import "@interfaces/IExecutor.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@permit2/src/interfaces/IAllowanceTransfer.sol";
+
+error TokenTransfer__InvalidPermit2();
```
dianacarvalho1: I think we are calling these type of errors just `AddressZero`. `InvalidPermit2` makes it sound like something is wrong with permit2 at execution level not initing

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -91,12 +96,20 @@ contract UniswapV3Executor is IExecutor, ICallback {
 
         address tokenIn = address(bytes20(msgData[128:148]));
 
+        require(
+            uint8(msgData[171]) <= uint8(TransferType.NONE),
+            "InvalidTransferMethod"
+        );
```
dianacarvalho1: can you write a comment explaining why this is invalid please?

dianacarvalho1: also it's type now, not method

foundry/test/executors/UniswapV3Executor.t.sol
```diff
@@ -19,7 +21,8 @@ contract UniswapV3ExecutorExposed is UniswapV3Executor {
             uint24 fee,
             address receiver,
             address target,
-            bool zeroForOne
+            bool zeroForOne,
+            TransferType method
```
dianacarvalho1: should be `transferType` not `method`

foundry/test/executors/UniswapV3Executor.t.sol
```diff
@@ -67,7 +80,8 @@ contract UniswapV3ExecutorTest is Test, Constants {
             uint24 fee,
             address receiver,
             address target,
-            bool zeroForOne
+            bool zeroForOne,
+            TokenTransfer.TransferType method
```
dianacarvalho1: here too

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -722,7 +722,7 @@ mod tests {
             "a478c2975ab1ea89e8196811f51a7b7ade33eb11", // component id
             "3ede3eca2a72b3aecc820e955b36f38437d01395", // receiver
             "00",                                       // zero2one
-            "00",                                       // exact out
+            "00",                                       // transfer method
```
dianacarvalho1: type

dianacarvalho1: and there are a bunch more in this file!

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 I like this!!
Only have really small comments now

foundry/src/executors/UniswapV3Executor.sol
tamaralipows: I think it's obvious that this would be referring to a transfer type that doesn't exist, no? Since NONE is the highest one

dianacarvalho1: oooohh no xD I didn't even realise that was the reason for this error 😬

PR #112, opened 2025-03-12
feat: add curve executor

royvardhan:
Added curve executor with swap tests for all the pools we index.


foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,62 @@
...
+import "@interfaces/ICurveRouter.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+error CurveExecutor__InvalidAddresses();
+
+contract CurveExecutor is IExecutor {
+    using SafeERC20 for IERC20;
+
+    ICurveRouter public immutable curveRouter;
+    address public immutable ethAddress;
```
tamaralipows: We expect this to work on other EVM-compatible chains right? In that case it should probably we native token, not eth?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+    // Get the number of coins in the pool
+    function get_n_coins(address pool) external view returns (uint256);
+    // Get the indices of the coins in the pool
+    function get_coin_indices(address pool, address from, address to)
+        external
+        view
+        returns (int128, int128, bool);
+}
+
+// Aave lending pool
```
tamaralipows: These comments are not very meaningful imo, this info is already in the interface name

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+}
+
+// Curve pool registry
+// This is the registry that contains the information about the pool
+// The naming convention is different because it is in vyper
+interface MetaRegistry {
+    // Get the number of coins in the pool
+    function get_n_coins(address pool) external view returns (uint256);
+    // Get the indices of the coins in the pool
+    function get_coin_indices(address pool, address from, address to)
```
tamaralipows: Same with these method comments - they don't add any additional info. It's a concept in the [book](chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://milkov.tech/assets/psd.pdf) - by the way, under the section `Comment Repeats Code`.

https://milkov.tech/assets/psd.pdf

The interface comment is fine

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        assertEq(params.route[0], WETH_ADDR);
+        assertEq(params.route[1], TRICRYPTO_POOL);
+        assertEq(params.route[2], USDC_ADDR);
+        assertEq(params.swapParams[0][0], 2);
+        assertEq(params.swapParams[0][1], 0);
+        assertEq(params.swapParams[0][2], 1);
+        assertEq(params.swapParams[0][3], 3);
+    }
+
+    // The following pools are unique and do not have a factory
```
tamaralipows: What does it mean if they don't have a factory? Do we have to execute differently for these pools?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+
+        assertEq(amountOut, 998096);
+        assertEq(
+            IERC20(USDC_ADDR).balanceOf(address(curveExecutorExposed)),
+            amountOut
+        );
+    }
+
+    // The following pool is from CryptoSwapNG, deployed by factory 0x6A8cbed756804B16E05E741eDaBd5cB544AE21bf
+    // - It is a plain pool
```
tamaralipows: What does plain pool vs meta pool mean? For my own knowledge

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        uint256 amountOut = curveExecutorExposed.swap(amountIn, data);
+
+        assertEq(amountOut, 6081816039338);
+        assertEq(
+            IERC20(WETH_ADDR).balanceOf(address(curveExecutorExposed)),
+            amountOut
+        );
+    }
+
+    // The following pool is from Tricrypto, deployed by factory 0x0c0e5f2fF0ff18a3be9b835635039256dC4B4963
```
tamaralipows: Why do we need to know these factory addresses?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        uint256 amountOut = curveExecutorExposed.swap(amountIn, data);
+
+        assertEq(amountOut, 999910);
+        assertEq(
+            IERC20(USDT_ADDR).balanceOf(address(curveExecutorExposed)),
+            amountOut
+        );
+    }
+
+    // The following pools were taken from the old curve executor tests
```
tamaralipows: Not a meaningful comment since future devs have no idea what the old tests are. Better to write what they are actually testing

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+            metaRegistry.get_coin_indices(pool, tokenIn, tokenOut);
+
+        swapParams[0][0] = uint256(int256(coinInIndex));
+        swapParams[0][1] = uint256(int256(coinOutIndex));
+        swapParams[0][2] = swapType;
+        swapParams[0][3] = poolType;
+        swapParams[0][4] = nCoins;
+    }
+
+    function dealAaveDai() internal {
```
tamaralipows: This function is only used in one place AFAIU - is it worth to make it a separate function?

tamaralipows (commented):
Thanks for this! I have mostly questions for my own understanding 🙏

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,62 @@
...
+    // slither-disable-next-line locked-ether
+    function swap(uint256 amountIn, bytes calldata data)
+        external
+        payable
+        returns (uint256)
+    {
+        ICurveRouter.CurveRouterParams memory params = _decodeData(data);
+        if (params.route[0] != ethAddress) {
+            // slither-disable-next-line unused-return
+            IERC20(params.route[0]).approve(address(curveRouter), amountIn);
```
tamaralipows: Missed one thing - we will actually be doing dynamic approvals here to save on gas (i.e. we approve once for MAX if we need to, otherwise we don't approve anything).

Please have a look at how this is done in Balancer!

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,62 @@
...
+contract CurveExecutor is IExecutor {
+    using SafeERC20 for IERC20;
+
+    ICurveRouter public immutable curveRouter;
+    address public immutable ethAddress;
+
+    constructor(address _curveRouter, address _ethAddress) {
+        if (_curveRouter == address(0) || _ethAddress == address(0)) {
+            revert CurveExecutor__InvalidAddresses();
+        }
```
dianacarvalho1: what if they are both 0?? this will pass 👀

foundry/src/executors/CurveExecutor.sol
dianacarvalho1: you need to pass a bool from encoding to decide if we need to approve or not!

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        vm.createSelectFork(vm.rpcUrl("mainnet"), forkBlock);
+        curveExecutorExposed = new CurveExecutorExposed(CURVE_ROUTER, ETH_ADDR);
+        metaRegistry = MetaRegistry(CURVE_META_REGISTRY);
+    }
+
+    function testDecodeParams() public view {
+        address[11] memory route =
+            _getRoute(WETH_ADDR, USDC_ADDR, TRICRYPTO_POOL);
+
+        // The meta registry does not have information about the pool.
```
dianacarvalho1: why is this relevant? Can I have a bit more context please? 😅  what does the meta registry have to do with anything here?

foundry/test/executors/CurveExecutor.t.sol
dianacarvalho1: we might. not having a factory means that the curve devs decided to deploy a pool on its own just because. and it has unique code 😬 it's terrible

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        address[5] memory pools;
+
+        deal(address(curveExecutorExposed), amountIn);
+        bytes memory data = abi.encode(
+            route, swapParams, amountIn, minAmountOut, pools, address(this)
+        );
+
+        uint256 amountOut = curveExecutorExposed.swap(amountIn, data);
+
+        assertTrue(amountOut >= 1 ether);
```
dianacarvalho1: can we have a = check instead of a >= please?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        assertEq(IERC20(STETH_ADDR).balanceOf(address(this)), amountOut - 1); // Gets 1 wei less than amountOut
```
dianacarvalho1: why?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,594 @@
...
+        uint256 amountOut = curveExecutorExposed.swap(amountIn, data);
+
+        assertEq(amountOut, 998096);
+        assertEq(
+            IERC20(USDC_ADDR).balanceOf(address(curveExecutorExposed)),
+            amountOut
+        );
+    }
+
+    // The following pool is from CryptoSwapNG, deployed by factory 0x6A8cbed756804B16E05E741eDaBd5cB544AE21bf
```
dianacarvalho1: can we have these comments inside of the test please? not outside

dianacarvalho1 (commented):
Thank you so much @royvardhan ! 🙏🏼  this looks great! I have some comments and one overall question:
How did you figure out which pool type (from the router types) were the pools?

foundry/src/executors/CurveExecutor.sol
tamaralipows: no? || isn't XOR?

foundry/test/executors/CurveExecutor.t.sol
royvardhan: Plain pool is a standard curve pool like `DAI/USDC/USDT` while meta pool is a pool that is based on top of a plain pool. ex: FRAX/3pool

foundry/test/executors/CurveExecutor.t.sol
royvardhan: Nothing, I took these comments from the indexer.

foundry/test/executors/CurveExecutor.t.sol
royvardhan: agreed!

foundry/test/executors/CurveExecutor.t.sol
royvardhan: It could be merged into the function where this is used, but then I thought of keeping the `deal` logic separate. So that we read the function name and know what exactly it is doing, just like`deal(token, address_to, amount)`

royvardhan: I prefer to keep it seperate, do you want me to merge it where it is used?

foundry/src/executors/CurveExecutor.sol
royvardhan: if any of these vals are zero address then we revert. If the first val is zero, it will revert without checking the other val.

foundry/test/executors/CurveExecutor.t.sol
royvardhan: Meta registry helps us get `n_coins` and `coin_indices` for any registered curve pool. And these vals are needed for a swap. For all the other tests I am using `_getSwapParams` which uses the meta registry, but for this particular pool it is reverting.

foundry/test/executors/CurveExecutor.t.sol
royvardhan: No idea, this was the same issue in older tests

foundry/src/executors/CurveExecutor.sol
royvardhan: Yeah I forgot, updating tests to supply a bool which indicates if approval is needed or not.

tamaralipows:
I see you are also having problems with slither, this is how I fixed it in the feature branch: https://github.com/propeller-heads/tycho-execution/pull/117/commits/f57cc213ec7868c886f3d99a66d354439f853d28

royvardhan:
> I see you are also having problems with slither, this is how I fixed it in the feature branch: [f57cc21](https://github.com/propeller-heads/tycho-execution/commit/f57cc213ec7868c886f3d99a66d354439f853d28)

thanks! fixed it.

foundry/test/CurveRouterGasTest.t.sol
```diff
@@ -0,0 +1,42 @@
...
+
+contract CurveRouterGasTest is Constants {
+    ICurveRouter curveRouter = ICurveRouter(CURVE_ROUTER);
+
+    function setUp() public {
+        uint256 forkBlock = 22031795;
+        vm.createSelectFork(vm.rpcUrl("mainnet"), forkBlock);
+    }
+
+    function testCurveRouter() public {
```
tamaralipows: It's nice to have a docstring explaining the swap path at the top of each test.

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,68 @@
...
+contract CurveExecutor is IExecutor {
+    using SafeERC20 for IERC20;
+
+    ICurveRouter public immutable curveRouter;
+    address public immutable nativeToken;
+
+    struct SwapParams {
+        address[11] route;
+        uint256[5][5] swapParams;
+        uint256 amountIn;
```
tamaralipows: we don't need this `amountIn`, do we?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,626 @@
...
+        uint256[5][5] memory swapParams;
+        swapParams[0][0] = 2; // tokenIn Index
+        swapParams[0][1] = 0; // tokenOut Index
+        swapParams[0][2] = 1; // swap type
+        swapParams[0][3] = 3; // pool type
+        swapParams[0][4] = 3; // n_coins
+
+        uint256 amountIn = 1 ether;
+        uint256 minAmountOut = 0;
+        address[5] memory pools;
```
tamaralipows: Is there not a single swap case where these pools are necessary to set?

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,626 @@
...
+
+    function _getRoute(address tokenIn, address tokenOut, address pool)
+        internal
+        pure
+        returns (address[11] memory route)
+    {
+        route[0] = tokenIn;
+        route[2] = tokenOut;
+        route[1] = pool;
+    }
```
tamaralipows: What are the other 8 addresses of the route? Is 11 / 3 just the max number of swaps you can make on a single call to the curve router?

foundry/test/executors/CurveExecutor.t.sol
tamaralipows: It's not clear from the function name to be exactly what this is doing though, maybe a docstring would be better? It seems we are dealing DAI to the curve executor, and then already sending it to to the lending pool? Why are we doing that? I could be misunderstanding something

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,626 @@
...
+        uint256 amountOut = curveExecutorExposed.swap(amountIn, data);
+
+        assertEq(amountOut, 999910);
+        assertEq(
+            IERC20(USDT_ADDR).balanceOf(address(curveExecutorExposed)),
+            amountOut
+        );
+    }
+
+    function testSwapAavePool() public {
```
tamaralipows: I see these tests don't correspond with any of the pools listed in https://github.com/propeller-heads/tycho-protocol-sdk/blob/main/substreams/ethereum-curve/integration_test.tycho.yaml

Are they just carried over from the previous repo? Are they necessary do you think?

tamaralipows (commented):
Thanks for this @royvardhan 

I have some minor questions, though my biggest concern is that I don't find the following [test cases](https://github.com/propeller-heads/tycho-protocol-sdk/blob/main/substreams/ethereum-curve/integration_test.tycho.yaml) here:

1. Metapool factory 0xB9fC157394Af804a3578134A6585C0dc9cc990d4 - PlainPool
2. CryptoPool factory 0xF18056Bbd320E96A48e3Fbf8bC061322531aac99 - with ETH
  
Could you please let me know if I've missed them?

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,68 @@
...
+        uint256 minAmountOut;
+        address[5] pools;
```
tamaralipows: If this isn't needed right now then I say we don't need to decode it. Instead, we can perhaps just hardcode it as input to the exchange method

foundry/test/executors/CurveExecutor.t.sol
royvardhan: Yes, they were carried over from the previous repo. If we are not indexing them anymore then we should remove them

royvardhan: I am removing them

royvardhan:
> Thanks for this @royvardhan
> 
> I have some minor questions, though my biggest concern is that I don't find the following [test cases](https://github.com/propeller-heads/tycho-protocol-sdk/blob/main/substreams/ethereum-curve/integration_test.tycho.yaml) here:
> 
> 1. Metapool factory 0xB9fC157394Af804a3578134A6585C0dc9cc990d4 - PlainPool
> 2. CryptoPool factory 0xF18056Bbd320E96A48e3Fbf8bC061322531aac99 - with ETH
> 
> Could you please let me know if I've missed them?

You have mentioned the factory address, but I get your point, these are the following pools that I haven't added in tests due to significantly low liquidity:

royvardhan:
> Thanks for this @royvardhan
> 
> I have some minor questions, though my biggest concern is that I don't find the following [test cases](https://github.com/propeller-heads/tycho-protocol-sdk/blob/main/substreams/ethereum-curve/integration_test.tycho.yaml) here:
> 
> 1. Metapool factory 0xB9fC157394Af804a3578134A6585C0dc9cc990d4 - PlainPool
> 2. CryptoPool factory 0xF18056Bbd320E96A48e3Fbf8bC061322531aac99 - with ETH
> 
> Could you please let me know if I've missed them?

You have mentioned the factory address, but I get your point, these are the following pools that I haven't added in tests due to significantly low liquidity:

1. [wstTAO/wTAO](https://etherscan.io/address/0xf2DCf6336D8250754B4527f57b275b19c8D5CF88) - $5k liquidity
2. [pxETH/mkUS](https://etherscan.io/address/0x71db3764d6841d8b01dc27c0fd4a66a8a34b2be0) - $178 liquidity

foundry/src/executors/CurveExecutor.sol
royvardhan: nope, we dont

foundry/test/executors/CurveExecutor.t.sol
royvardhan: resolving notes: had a discussion about this In chat

foundry/test/executors/CurveExecutor.t.sol
royvardhan: You can do upto 5 swaps in a single call.

royvardhan: Flow: 
 - tokenIn
 - pool
 - tokenOut
 - pool
 - tokenOut
 - pool
 - tokenOut
 - pool
 - tokenOut
 - pool
 - tokenOut

tamaralipows:
We still support those pool types despite the low liquidity, don't we?

foundry/src/executors/CurveExecutor.sol
royvardhan: yes sure. Thanks for this suggestion, i am removing it

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -252,7 +252,7 @@ contract CurveExecutorTest is Test, Constants {
     }
 
     function testSwapWethXyoPool() public {
-        // The following pool is from CryptoPool, deployed by factory 0xF18056Bbd320E96A48e3Fbf8bC061322531aac99
+        // The following pool is from CryptoPool, deployed by factory 0xF18056Bbd320E96A48e3Fbf8bC061322531aac99 - with ETH
```
tamaralipows: I'm a bit confused - is this a `WETH/XYO` pool or an `ETH/XYO` pool? In the sdk integration tests it seems to be an `ETH/XYO` pool but here you are testing with `WETH`. Does it support both? I think this test case should test ETH since you have already added a non-ETH test case, right?

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,67 @@
...
+contract CurveExecutor is IExecutor {
+    using SafeERC20 for IERC20;
+
+    ICurveRouter public immutable curveRouter;
+    address public immutable nativeToken;
+
+    struct SwapParams {
+        address[11] route;
+        uint256[5][5] swapParams;
+        uint256 minAmountOut;
```
tamaralipows: We should remove this from the swap params since we aren't checking min amount out at the executor level.

tamaralipows (commented):
Thanks - looks like we are almost there! See my two comments please

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.26;
+
+import "@interfaces/IExecutor.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+error CurveExecutor__InvalidAddresses();
+
+interface CryptoPool {
```
dianacarvalho1, line 9: do you think these should be in a lib? I kinda felt like no. we literally only need them here

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,133 @@
...
+    // slither-disable-next-line naming-convention
```
dianacarvalho1, line 10: I decided to keep the naming like it was in Vyper (the language curve is written in)

foundry/test/executors/CurveExecutor.t.sol
```diff
@@ -0,0 +1,351 @@
...
+        internal
+        view
+        returns (int128, int128)
+    {
+        (int128 coinInIndex, int128 coinOutIndex,) =
+            metaRegistry.get_coin_indices(pool, tokenIn, tokenOut);
+        return (coinInIndex, coinOutIndex);
+    }
+
+    function dealAaveDai() internal {
```
dianacarvalho1: ah I see this function is useless now. I will delete it tomorrow!

foundry/test/Constants.sol
```diff
@@ -30,6 +30,7 @@ contract Constants is Test, BaseConstants {
     address UNPAUSER = makeAddr("unpauser");
 
     // Assets
+    address ETH_ADDR = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
dianacarvalho1, line 33: this worries me 😬 
I've tried to run the tests with ETH being the zero address and it works for all of them except the stETH pool 😬 this one it really needs to be this eeeeeEee address. @tamaralipows Do you think it's ok to just fix this in the encoder? Meaning, to see if it's this one stupid pool and if it's ETH then change the token address to EeeeeE

foundry/src/executors/CurveExecutor.sol
```diff
@@ -0,0 +1,133 @@
...
+        returns (uint256)
+    {
+        (
+            address tokenIn,
+            address tokenOut,
+            address pool,
+            uint8 poolType,
+            int128 i,
+            int128 j,
+            bool tokenApprovalNeeded
```
dianacarvalho1, line 60: notice that there isn't a receiver 😕 We will never be able to optimize the transfers from a curve pool. it will always have to go through the router

foundry/test/Constants.sol
dianacarvalho1: I ended up fixing this in the last commit of the encoding PR [here](https://github.com/propeller-heads/tycho-execution/pull/153/commits/7cc0e8f58699cd487ce39c08dd45d4731cf6fb8c)

tamaralipows: dead link :(

foundry/src/executors/CurveExecutor.sol
tamaralipows: no I think this is fine

tamaralipows (approved):
Thank you for this! It looks so clean now!

PR #114, opened 2025-03-13
feat: Add validation for cyclical trades

dianacarvalho1:
We only support cyclical trades when the repeated token is the first and the last token in the swap. We don't support cycles in between


src/encoding/evm/tycho_encoder.rs
```diff
@@ -436,4 +484,125 @@ mod tests {
...
+            },
+            Swap {
+                component: ProtocolComponent {
+                    id: "0x0000000000000000000000000000000000000000".to_string(),
+                    protocol_system: "uniswap_v2".to_string(),
+                    ..Default::default()
+                },
+                token_in: dai(),
+                token_out: weth(),
+                split: 0.5f64,
```
royvardhan: Shouldn't this be `0f64`, that would automatically take the rest of the DAI. Otherwise total split would be 1 (0.5+0.5) and it will trigger this [error](https://github.com/propeller-heads/tycho-execution/blob/f62a9d28c0683490d841439d6a0543370d238387/src/encoding/evm/strategy_encoder/strategy_validators.rs#L81)

dianacarvalho1: yes you are right! fixing it!

src/encoding/evm/tycho_encoder.rs
```diff
@@ -436,4 +484,125 @@ mod tests {
             )
         );
     }
+
+    #[test]
+    fn test_validate_cyclical_swap() {
```
tamaralipows, line 495: I'm intersted in a test case like 
```
                  - DAI
DAI - WETH  
                  - DAI
```
                        
How difficult would it be to add? I could also do it

src/encoding/evm/tycho_encoder.rs
```diff
@@ -92,6 +94,44 @@ impl EVMTychoEncoder {
                 }
             }
         }
+
```
tamaralipows, line 97: It's a bit unclear to me, since you are checking the first and last swap tokens and not the tokens of the solution itself - what do we support in the unwrapping/wrapping case? Could you add a comment about this?

like is WETH -> DAI -> WETH -(unwrap)-> ETH okay?

src/encoding/evm/tycho_encoder.rs
```diff
@@ -92,6 +94,44 @@ impl EVMTychoEncoder {
...
+        let mut split_tokens_already_considered = HashSet::new();
+        for (i, swap) in solution.swaps.iter().enumerate() {
+            // so we don't count the split tokens more than once
+            if swap.split != 0.0 {
+                if !split_tokens_already_considered.contains(&swap.token_in) {
+                    solution_tokens.push(swap.token_in.clone());
+                    split_tokens_already_considered.insert(swap.token_in.clone());
+                }
+            } else {
+                // it's the last swap of the split
```
tamaralipows: it could also simply be not a split, right?

src/encoding/evm/tycho_encoder.rs
dianacarvalho1: no no, I thought we decided that we wouldn't support those cases! I should add a check for that too actually! if it's a cyclical swap, no wrapping/unwrapping allowed

src/encoding/evm/tycho_encoder.rs
dianacarvalho1: the comment is about the next line.. not the previous one. will improve it

src/encoding/evm/tycho_encoder.rs
```diff
@@ -92,6 +94,44 @@ impl EVMTychoEncoder {
                 }
             }
         }
+
+        let mut solution_tokens = vec![];
+        let mut split_tokens_already_considered = HashSet::new();
```
tamaralipows, line 99: Feels like a weird name when it's also containing non-split tokens?

src/encoding/evm/tycho_encoder.rs
```diff
@@ -92,6 +94,44 @@ impl EVMTychoEncoder {
...
+            // so we don't count the split tokens more than once
+            if swap.split != 0.0 {
+                if !split_tokens_already_considered.contains(&swap.token_in) {
+                    solution_tokens.push(swap.token_in.clone());
+                    split_tokens_already_considered.insert(swap.token_in.clone());
+                }
+            } else {
+                // it's the last swap of the split
+                if split_tokens_already_considered.contains(&swap.token_in) {
+                    split_tokens_already_considered.remove(&swap.token_in);
```
tamaralipows: I don't really understand this part, sorry :( why are we removing it?

src/encoding/evm/tycho_encoder.rs
tamaralipows: yes im fine with not supporting those cases, just it's not super clear from the code or comments

tamaralipows: also - we should also get rid of the integration tests for those cases, what do you think?

src/encoding/evm/tycho_encoder.rs
dianacarvalho1: this passes fine!

src/encoding/evm/tycho_encoder.rs
dianacarvalho1: huumm no, it only contains split tokens

src/encoding/evm/tycho_encoder.rs
tamaralipows: Did you add this test case? I'm not seeing it

dianacarvalho1: yes. I collapsed it ehehe

src/encoding/evm/tycho_encoder.rs
```diff
@@ -436,4 +490,188 @@ mod tests {
...
+                "Cyclical swaps are only allowed if they are the first and last token of a solution".to_string()
+            )
+        );
+    }
+    #[test]
+    fn test_validate_cyclical_swap_split_native_action_fail() {
+        // This validation fails because there is a native action with a valid cyclical swap
+        //                    -> WETH
+        // ETH -> WETH -> DAI
+        //                    -> WETH
```
dianacarvalho1: @tamaralipows it's here! it gets to the native action check and fails, so it assumed that the cyclical swap was valid!

src/encoding/evm/tycho_encoder.rs
```diff
@@ -92,6 +94,50 @@ impl EVMTychoEncoder {
...
+                }
+            } else {
+                // it might be the last swap of the split or a regular swap
+                if !split_tokens_already_considered.contains(&swap.token_in) {
+                    solution_tokens.push(swap.token_in.clone());
+                }
+            }
+            if i == solution.swaps.len() - 1 {
+                solution_tokens.push(swap.token_out.clone());
+            }
```
tamaralipows, line 115: I'm a bit confused by this part. In this case, aren't you forgetting the token out of the second-last swap?
i.e. the token_in of the last swap? 
Shouldn't you be pushing both tokens in this case?

tamaralipows: Ahhhh I just realized it's an if, not an else. All good I think

src/encoding/evm/tycho_encoder.rs
tamaralipows: Ahhh hmm I think I'd like an explicit test for that, otherwise someone may change this not knowing it's actually testing two things at once (an "advanced" swap route), and focus only on the native action part with a simple route... Does that make sense?

tamaralipows (approved):
Thanks a lot! It's a lot clearer now! Just see that little nitpick about the tests

royvardhan (approved):
Looks good. Thank you 🙏🏻

PR #115, opened 2025-03-14
BREAKING CHANGE: rename split swap interfaces

tamaralipows:


foundry/src/TychoRouter.sol
```diff
@@ -145,7 +145,7 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
         bytes calldata swaps
     ) public payable whenNotPaused nonReentrant returns (uint256 amountOut) {
         IERC20(tokenIn).safeTransferFrom(msg.sender, address(this), amountIn);
-        return _swapChecked(
+        return _splitSwapChecked(
```
dianacarvalho1, line 148: should we keep the `checked`? I feel like it's not necessary

dianacarvalho1: aaff ok I accept it. aff we have too many methods with the same name 😩

dianacarvalho1: what bothers me is that it does more than checking as well 😕 it does (un)wrapping, transfers and fees.
could we give it a more general name? maybe one of them could be `execute_split_swap`? 😬 
or `split_swap_trade`? I dunno, I'm just saying stuff. I don't mind if we keep it like this either though, this is a tough problem tbh

tamaralipows: yeah, I'll keep it in mind, though not sure of an elegant alternative. don't want it to be too similar to the innermost `_splitSwap`

tamaralipows:
Closed due to bad branch name. Reopened at #116

PR #117, opened 2025-03-16
feat: Single swap methods

tamaralipows:
- Remove duplicate test case for single swp with permit2. This test came from the days when we could still have a 0 minimum amount.

### DIFF
--- a/.github/workflows/slither.yml
+++ b/.github/workflows/slither.yml
@@ -11,7 +11,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+      - name: Build the contracts
+        run: forge build --build-info
+        with:
+          target: 'foundry/'
       - uses: crytic/slither-action@v0.4.0
         with:
           target: 'foundry/'
+          ignore-compile: true
           slither-args: '--filter-paths foundry/lib/'
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -446,7 +446,7 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
 
         uint256 amountConsumed = initialBalance - currentBalance;
 
-        if (tokenIn != tokenOut && amountConsumed != amountIn) {
+        if (amountConsumed != amountIn) {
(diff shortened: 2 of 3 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -291,6 +402,76 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
+
+        amountOut =
+            _callExecutor(swap_.executor(), amountIn, swap_.protocolData());
+        uint256 currentBalance = tokenIn == address(0)
+            ? address(this).balance
+            : IERC20(tokenIn).balanceOf(address(this));
+
+        uint256 amountConsumed = initialBalance - currentBalance;
+
+        if (tokenIn != tokenOut && amountConsumed != amountIn) {
```
royvardhan: Can a single swap be a cyclic swap?

royvardhan (approved):

foundry/src/TychoRouter.sol
tamaralipows: I don't think so

royvardhan: Then we can remove this `tokenIn != tokenOut` check

.github/workflows/slither.yml
```diff
@@ -11,7 +11,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+      - name: Build the contracts
+        run: forge build --build-info
+        with:
+          target: 'foundry/'
```
dianacarvalho1, line 19: why is this needed?

foundry/src/TychoRouter.sol
```diff
@@ -222,6 +222,117 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
+            nTokens,
+            receiver,
+            swapData
+        );
+    }
+
+    /**
+     * @notice Executes a single swap operation.
+     *         This function enables optional ETH wrapping/unwrapping, and validates the output amount
+     *         against a user-specified minimum.
```
dianacarvalho1, line 276: I think here we should mention that it used Permit2 to transfer tokens

dianacarvalho1 (commented):
Thank you @tamaralipows ! 🙏🏼 tiny tiny requests

PR #120, opened 2025-03-18
feat: Add sequential swap methods

tamaralipows:
- Add some basic tests (will add more in next commits)
- Adapt sequential tests (which originally tested split swap)
- Adapt a forgotten single swap test
- Fix encoding/decoding of single and sequential swaps to not expect split/token info every time (that would consume gas unnecessarily)
- Better test organization (different file for each method)

Please read commit messages and/or review by commit.


foundry/lib/LibSwap.sol
```diff
@@ -2,36 +2,42 @@
...
 library LibSwap {
-    /// Returns the InToken index into an array of tokens
-    function tokenInIndex(
-        bytes calldata swap
-    ) internal pure returns (uint8 res) {
-        res = uint8(swap[0]);
+    /**
+     * @dev Returns arguments required to perform a single swap
+     */
+    function decodeSingleSwap(bytes calldata swap)
```
royvardhan, line 8: Can we merge this with the sequential decode function since it shares the same logic? I know you have may want to have a separate naming but we could just have one with 'decodeNonSplitSwap' naming. Do you agree?

foundry/src/TychoRouter.sol
```diff
@@ -472,6 +573,75 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
+        uint256 amountIn,
+        address tokenIn,
+        address tokenOut,
+        uint256 minAmountOut,
+        bool wrapEth,
+        bool unwrapEth,
+        address receiver,
+        bytes calldata swaps
+    ) internal returns (uint256 amountOut) {
+        if (receiver == address(0)) {
```
royvardhan, line 594: The receiver and the min amount out is checked in the _single, _split, _sequential internal function. Can we have one validation function and just call that internal function?

foundry/test/LibSwap.t.sol
```diff
@@ -7,7 +7,45 @@ import "../lib/LibSwap.sol";
...
-    function testSwap() public view {
+    function testSingleSwap() public view {
+        address executor = 0x1234567890123456789012345678901234567890;
+        bytes memory protocolData = abi.encodePacked(uint256(123));
+
+        bytes memory swap = abi.encodePacked(executor, protocolData);
+        this.assertSingleSwap(swap, executor, protocolData);
+    }
+
+    function assertSingleSwap(
```
royvardhan, line 18: If this assert function is just used as an helper, then this could be an internal function. Sorry if I missed anything.

foundry/test/LibSwap.t.sol
```diff
@@ -7,7 +7,45 @@ import "../lib/LibSwap.sol";
...
+
+    function testSequentialSwap() public view {
+        address executor = 0x1234567890123456789012345678901234567890;
+        bytes memory protocolData = abi.encodePacked(uint256(234));
+
+        bytes memory swap = abi.encodePacked(executor, protocolData);
+        this.assertSequentialSwap(swap, executor, protocolData);
+    }
+
+    function assertSequentialSwap(
```
royvardhan, line 37: Same ^

foundry/test/TychoRouterSplitSwap.t.sol
```diff
@@ -0,0 +1,683 @@
...
+            uint8(0),
+            uint24(0),
+            address(usv2Executor),
+            wethUsdcV2OneZeroData
+        );
+        tychoRouter.exposedSplitSwap(amountIn, 2, pleEncode(swaps));
+        assertEq(IERC20(USDC_ADDR).balanceOf(tychoRouterAddr), 99574171);
+    }
+
+    function testSplitOutputCyclicSwapInternalMethod() public {
```
royvardhan, line 611: Do we still need this? Didn't we remove the integration test for this case? Or am I daydreaming?

royvardhan (changes requested):

royvardhan:
These changes are so good for open source contributors reading experience for the repo. Thank you for this 🙏🏻

foundry/lib/LibSwap.sol
tamaralipows: I thought about that but I prefer to keep it separate tbh

foundry/test/TychoRouterSplitSwap.t.sol
tamaralipows: We removed the integration test only for cases of wrapping/unwrapping after deciding the swap should start and end on strictly the same token. This case is still valid

foundry/src/TychoRouter.sol
tamaralipows: How would we do this, considering that in between checking the initial and final balance, we need to call the executor in different ways?

**edit:** I was questioning whether this would add to the organization or complicate things further, because some validations need to be left in these methods (and unfortunately duplicated).

For example, the check that all input token has been spent - this one cannot be moved into a validation because it depends on the balance before calling the executor, and the balance after calling the executor. So really the only thing that can be validated in a separate method would be the min amount out and the receiver…

Also note that the receiver would not be able to be in the same validation method, since this is called before calling the executor, and the min amount is checked after calling the executor.

foundry/test/TychoRouterSplitSwap.t.sol
royvardhan: okay, got it!

foundry/test/LibSwap.t.sol
tamaralipows: No this isn't possible - they need to be public if we are calling them using `this.method()`

royvardhan (approved):

dianacarvalho1 (commented):
uuuufff soooo niceeeeee!! 🧹 thank you for all the cleaning, reorganising and docs!!

PR #122, opened 2025-03-19
fix: Slippage precision calculation

tamaralipows:
- The resolution wasn't high enough before, so 0.1% slippage would round to 0% slippage. We should instead use basis point resolution to overcome this.


royvardhan (approved):

dianacarvalho1 (commented):
oh noooooooo, great catch! 🙏🏼 thank yooou

PR #126, opened 2025-03-20
feat: Ekubo Integration

die-herdplatte:
In conjunction with https://github.com/propeller-heads/tycho-protocol-sdk/pull/172 and https://github.com/propeller-heads/tycho-simulation/pull/173


src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -258,6 +259,74 @@ impl SwapEncoder for BalancerV2SwapEncoder {
...
+        );
+
+        let extension: Address = get_static_attribute(&swap, "extension")?
+            .as_slice()
+            .try_into()
+            .map_err(|_| EncodingError::FatalError("extension should be an address".to_string()))?;
+
+        let mut encoded = vec![];
+
+        if encoding_context.group_token_in == swap.token_in {
```
die-herdplatte, line 305: What if the token_in appears at the start of a route and later on again?
There doesn't seem to be a way to detected this with the given `EncodingContext`.

die-herdplatte (commented):
Should Ekubo's callbacks be directly integrated into `TychoRouter.sol`?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -258,6 +259,74 @@ impl SwapEncoder for BalancerV2SwapEncoder {
...
+    }
+
+    fn encode_swap(
+        &self,
+        swap: Swap,
+        encoding_context: EncodingContext,
+    ) -> Result<Vec<u8>, EncodingError> {
+        if !swap.split.is_zero() {
+            return Err(EncodingError::InvalidInput("splits not implemented".to_string()));
+        }
```
tamaralipows: Our main `TychoRouter` contract should take care of this sufficiently - no special implementation is needed on the `EkuboSwapEncoder` side to support this.

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: Could you clarify why you need to detect this? We only support cyclic swaps that start and end on the same token.

tamaralipows (commented):
Thanks a lot for this PR!

Regarding this:

> Should Ekubo's callbacks be directly integrated into TychoRouter.sol?

This is discouraged if you can make the callback work with the `handleCallback` method in the executor - which seems to be the case here. No changes to the main `TychoRouter` necessary.

Could you please update your branch with the main branch - we have recently pushed a fix that should make CI run properly in external contribution PRs now.

Other than this - I would recommend to add an integration test on for the executor contract which uses data encoded by the swap encoder. Would it be possible for you to add one for the `EkuboExecutor`? See simple examples here:

`testSwapIntegration` in `BalancerV2ExecutorTest`
`testMultipleSwapIntegration` in `UniswapV4ExecutorTest`

src/encoding/evm/swap_encoder/swap_encoders.rs
die-herdplatte: The referenced line is a detection of whether the current call should encode the first swap of a swap group, [similar to how the v4 implementation does it](https://github.com/propeller-heads/tycho-execution/blob/59a5a558f5c6da905e64982566684571e1b5dd55/src/encoding/evm/swap_encoder/swap_encoders.rs#L167).
This seems to be the best way to detect that but becomes problematic when we have a swap group of e.g. ETH -> USDC -> ETH -> WBTC -> ETH.

tamaralipows: This swap scenario is not permitted with our current encoding - an error will be returned from `EVMTychoEncoder.validate_solution`. A token can only appear more than once if it is in the beginning and end of the swap.

Please let us know if you envision this causing any problems for your use case.

die-herdplatte: Gotcha!

> Please let us know if you envision this causing any problems for your use case.

Hm, it might not be a very pressing issue, but this restriction basically prevents solvers from choosing arbitrary routes inside Ekubo.

die-herdplatte:
> This is discouraged if you can make the callback work with the `handleCallback` method in the executor - which seems to be the case here. No changes to the main `TychoRouter` necessary.

Slightly related: Do I need the direct callback implementations on the executor itself (`locked` and `payCallback`).
I just implemented them because other executors did the same and it simplifies tests but shouldn't the executor only be callable by the router?

> Could you please update your branch with the main branch - we have recently pushed a fix that should make CI run properly in external contribution PRs now.

Done, but seems like there are still issues.

tamaralipows:
You're right to point out that the executor will only be callable through `delegatecall`s by the router, and thus those direct callbacks don't need to be implemented. We had included them just for clarity for the user and ease of testing - it's fine if you leave them here though.

One thing that I'd like to double check - does this work through the router? We need to make sure the callback data starts with the executor address in that case:
```
    function _handleCallback(bytes calldata data) internal {
        address executor = address(uint160(bytes20(data[data.length - 20:])));
  ```
  
**I'll take over this PR to double check this - thanks!**
  
  About the CI - this should be fixed soon, we are working on this internally.

die-herdplatte:
I just noticed that the executor doesn't use permits to pay Ekubo Core.
I can rectify this until tomorrow if you want.

tamaralipows:
> I just noticed that the executor doesn't use permits to pay Ekubo Core.
I can rectify this until tomorrow if you want.

Sure - that would be a big help!

die-herdplatte:
Nevermind, I see that the router uses the permit to transfer tokens to itself before calling the executor's swap function.
Then everything is fine from my side now.

tamaralipows (approved):
Thanks a lot for this! Any further changes will be taken care of in a separate PR.

PR #130, opened 2025-03-26
feat: SingleSwapStrategyEncoder

tamaralipows:
- Had to take the implementation of the swap header encoding out of the main EVMStrategyEncoder trait, since it will now be difference for single and split swap strategies.
- Integration tests will be added in separate task/PR.

### DIFF
--- a/src/encoding/evm/strategy_encoder/strategy_encoders.rs
+++ b/src/encoding/evm/strategy_encoder/strategy_encoders.rs
@@ -16,31 +16,13 @@ use crate::encoding::{
         swap_encoder::swap_encoder_registry::SwapEncoderRegistry,
         utils::{
             biguint_to_u256, bytes_to_address, encode_input, get_min_amount_for_solution,
-            get_token_position, percentage_to_uint24,
+            get_token_position, percentage_to_uint24, ple_encode,
         },
     },
     models::{Chain, EncodingContext, NativeAction, Solution},
     strategy_encoder::StrategyEncoder,
     swap_encoder::SwapEncoder,
 };
-
-/// Encodes a solution using a specific strategy for execution on the EVM-compatible network.
-pub trait EVMStrategyEncoder: StrategyEncoder {
-    /// Uses prefix-length encoding to efficient encode action data.
-    ///
-    /// Prefix-length encoding is a data encoding method where the beginning of a data segment
-    /// (the "prefix") contains information about the length of the following data.
-    fn ple_encode(&self, action_data_array: Vec<Vec<u8>>) -> Vec<u8> {
-        let mut encoded_action_data: Vec<u8> = Vec::new();
-
-        for action_data in action_data_array {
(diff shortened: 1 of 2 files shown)


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -57,6 +38,163 @@ pub trait EVMStrategyEncoder: StrategyEncoder {
     }
 }
 
+/// Represents the encoder for a swap strategy which supports single swaps.
```
dianacarvalho1, line 27: note for later: maybe we can add a section on the concepts in the docs for single, sequential and split swaps

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -57,6 +38,163 @@ pub trait EVMStrategyEncoder: StrategyEncoder {
...
+    /// a protocol.
+    fn encode_swap_header(&self, executor_address: Bytes, protocol_data: Vec<u8>) -> Vec<u8> {
+        let mut encoded = Vec::new();
+        encoded.extend(executor_address.to_vec());
+        encoded.extend(protocol_data);
+        encoded
+    }
+}
+
+impl EVMStrategyEncoder for SingleSwapStrategyEncoder {}
```
dianacarvalho1: I think we can get rid of the EVMStrategyEncoder now 🤔 it existed because of the swap header method. The only other method is the ple_encode, which we can put in utils?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -57,6 +38,163 @@ pub trait EVMStrategyEncoder: StrategyEncoder {
...
+                "singleSwap(uint256,address,address,uint256,bool,bool,uint256,address,bytes)"
+                    .to_string(),
+            )
+        };
+        Ok(Self { permit2, selector, swap_encoder_registry, router_address })
+    }
+
+    /// Encodes information necessary for performing a single swap against a given executor for
+    /// a protocol.
+    fn encode_swap_header(&self, executor_address: Bytes, protocol_data: Vec<u8>) -> Vec<u8> {
```
dianacarvalho1, line 67: I think we could add `encode_swap_header` to the StrategyEncoder trait

dianacarvalho1 (commented):
Thank yoouuu @tamaralipows 🙏🏼 I have one suggestion!

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: hmmm it's not used for the executor strategy no?

dianacarvalho1: ah yes.. true..

dianacarvalho1 (approved):
beeeryy niceeee ✨

PR #131, opened 2025-03-26
breaking: Remove router_address from Solution, set default

tamaralipows:
- The router address can be set when creating the `SplitSwapStrategy`, which now takes an `Option<Bytes>` as input during initialization (another breaking interface change)
- This change allows us to change our router address in the future with minimal effect on the users. Users should only pass the router address in the split swap initialization if they intend to use their own router for execution.
- This change also means that the router address does not need to be passed with the solution, even when using the executor strategy (which was pointless).
- I thought of having a `router_address()` method to set this in the encoder builder - but it seemed too messy, since we don't need the router address for execution for example. We would then potentially unnecessarily load and set the default router address when not needed. It is also not even used at the highest level `EVMTychoEncoder`, so it makes more sense for it to be directly associated with the swap strategy instead.
- Users will now not be able to encode for different routers without re-initializing the strategy. We assumed this use case to be very rare and not worth supporting.


src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -309,7 +331,7 @@ impl StrategyEncoder for ExecutorStrategyEncoder {
             let encoding_context = EncodingContext {
                 receiver: receiver.clone(),
                 exact_out: solution.exact_out,
-                router_address: router_address.clone(),
+                router_address: Bytes::from("0x0000000000000000000000000000000000000000"),
```
dianacarvalho1: maybe we could make it optional then?

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I think this looks good, it just bothers me that we had decided not to do this in the past because having default values/configs was too pythonic and not rust like 😐 
I have some other concerns:
1. We need to update the docs and the quickstart as well
3. We need to make a statement to alert the users of these breaking changes
4. Since we are deciding the router address for the users, every time we update the router address we need to be veeeeeeery explicit that if the user uses a new version of tycho-execution, the new router will be used. To make sure that we don't do things silently and the user is not aware of the changes

dianacarvalho1 (approved):

PR #138, opened 2025-03-27
feat: Support Ekubo callback in TychoRouter

tamaralipows:
**Only review the final commit**

- add integration test
- cannot directly call _handleCallback from the locked method of the tycho router because of bytes memory to bytes callback conversion
- Rename to EkuboPoolKey because of conflict with USV4 pool key

- **Bonus:** fix bug where input token to swap method must be ERC20 (we should also support ETH)

**NOTE:** This is not a final solution. This is a temporary solution since the Ekubo deployment is URGENT. We will solve this issue more elegantly soon by refactoring the callback functionality.

### DIFF
--- a/foundry/src/TychoRouter.sol
+++ b/foundry/src/TychoRouter.sol
@@ -143,7 +143,11 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
         address receiver,
         bytes calldata swaps
     ) public payable whenNotPaused nonReentrant returns (uint256 amountOut) {
-        IERC20(tokenIn).safeTransferFrom(msg.sender, address(this), amountIn);
+        if (address(tokenIn) != address(0)) {
+            IERC20(tokenIn).safeTransferFrom(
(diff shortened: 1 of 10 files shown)


foundry/src/TychoRouter.sol
```diff
@@ -529,4 +533,24 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
         _handleCallback(data);
         return "";
     }
+
+    function locked(uint256) external {
+        // TODO replace with real executor address once deployed
+        address executor = address(0x5991A2dF15A8F6A256D3Ec51E99254Cd3fb576A9);
```
dianacarvalho1: oh uau, this is the hack you mentioned right? 😬

dianacarvalho1: uuff this is not so nice. I understand this is quite urgent though right? 😬

dianacarvalho1 (commented):
😬

dianacarvalho1 (approved):
let's go then 😬

PR #143, opened 2025-03-31
feat: Refactor callback to use transient storage

dianacarvalho1:
With this, we don't need the univ3, pancakev3 and ekubo specific methods in the router contract. This should be flexible enough for most protocols that integrate with us

I tried to do the same for uniswap v4 but it doesn't work at all. The tests just fail with an EVM revert and no error message, even when the data passed is correct 🤔 

I _think_ this is safe enough actually

[9th April addition]

Support returning values from the callback
Rollback some of the Ekubo's Executor changes to a previous version to use the generic callback logic using transient storage (to do this I used [this PR](https://github.com/propeller-heads/tycho-execution/pull/146) as a baseline)


foundry/src/Dispatcher.sol
```diff
@@ -62,6 +62,10 @@ contract Dispatcher {
             revert Dispatcher__UnapprovedExecutor();
         }
 
+        assembly {
+            tstore(0, executor)
+        }
```
kayibal, line 67: Why assembly, newest solidity version doesn't require it anymore: https://docs.soliditylang.org/en/latest/contracts.html#transient-storage 

Also I think this won't work if you have nested callbacks which are useful sometimes for gas savings. E.g:

```
TychoRouter.swap -> PoolManager.swap
     PoolManager.swap -> TychoRouter.fallback
           TychoRouter.fallback -> UniswapV3Pool.swap
                UniswapV3Pool.swap -> TychoRouter.fallback
```

kayibal: So either we consciously do not support these in which case you'd want to error if there the current value is already set to sth or we support these in which case you'd need a mapping and an index:

```
uint256 transient callDepth
mapping(uint256 -> address) transient currentExecutor
```

tamaralipows: I vote we don't support this for now... we don't have plans to support nested callbacks in the foreseeable future.

dianacarvalho1: > So either we consciously do not support these in which case you'd want to error if there the current value is already set to sth

if this is a multiple hops swap, won't the value be already set as well? 🤔

dianacarvalho1: @kayibal the `transient` in solidity is only available in 0.8.28 and we are using 0.8.26. and I think we can't upgrade because we depend on uniswap v4 and they are using 0.8.26

tamaralipows (approved):
Thank you! From my side - all looks correct. Looking forward to hear from max when we audit this.

PR #148, opened 2025-04-02
feat: Refactor router tests 

dianacarvalho1:
- Created methods to deploy the router and the executors. Whenever we use rollFork we need to redeploy everything! Notice that the addresses will be different then!
- Created a test_executor_addresses.json to be used in the encoding tests, this way the calldata for the integration tests is already correct and we don't need to do any manual replacing (this was annoying). The addresses in this file match with the addresses used in the solidity tests
- Mock the ekubo router test


config/test_executor_addresses.json
```diff
@@ -0,0 +1,12 @@
...
+    "uniswap_v2": "0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f",
+    "sushiswap_v2": "0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f",
+    "pancakeswap_v2": "0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f",
+    "uniswap_v3": "0x2e234DAe75C793f67A35089C9d99245E1C58470b",
+    "pancakeswap_v3": "0x5991A2dF15A8F6A256D3Ec51E99254Cd3fb576A9",
+    "uniswap_v4": "0xF62849F9A0B5Bf2913b396098F7c7019b51A820a",
+    "vm:balancer_v2": "0xc7183455a4C133Ae270771860664b6B7ec320bB1",
+    "ekubo_v2": "0x2a07706473244BC757E10F2a9E86fB532828afe3"
+  }
+}
```
tamaralipows, line 12: Couldn't we use the same executor addresses that are actually deployed? Or is this too big of a change for this PR

dianacarvalho1: huumm yes, we could by using that `vm.etch` 🤔 do you think that is better?

tamaralipows: I think the less we have to maintain/update, the easier we will catch bugs, if it's not too much work to change that 🥺 I would vote we use that

dianacarvalho1: ok ok, agree. changing in a bit

dianacarvalho1: aaff it's just annoying because I can't read from this file in solidity 😬 I will have to hardcode the addresses in the tests as well

dianacarvalho1: so in the end we would still have to maintain the addresses in two different places 😕

dianacarvalho1: I think it's simpler to keep as it is tbh. it makes the solidity tests less complicated (without the etching)

tamaralipows: ok ok - sounds fine in that case!

tamaralipows: thank you for trying!

tamaralipows (approved):
tenk yu

PR #153, opened 2025-04-04
feat: Curve Encoder

dianacarvalho1:
- Add CurveEncoder
We don't know the pool tokens in the ProtocolComponent, so we can't infer the indexes correctly. Added a call to the MetaRegistry curve contract to get the correct token indexes. To do this, I had to move the get_client to utils. We could actually refactor the transaction logic into its' own struct and use it here and in the approval logic
It was assumed that all the pools will have a "factory" static attribute, even if empty

- Add integration test (Bonus fix of a shameful typo 😬)
- Fix Ekubo test (last commit). **This is really painful**. Because I deployed a new Executor, now all the contracts that are deployed afterwards will have different addresses 😬 So I had to change them everywhere. We should take time during cooldown to think of a better way to do this


foundry/test/TychoRouter.t.sol
```diff
@@ -1035,10 +1035,10 @@ contract TychoRouterTest is TychoRouterTestSetup {
...
         );
 
-        uint256 balancerAfter = IERC20(USDC_ADDR).balanceOf(ALICE);
+        uint256 balanceAfter = IERC20(USDC_ADDR).balanceOf(ALICE);
 
         assertTrue(success, "Call Failed");
-        assertGe(balancerAfter - balancerBefore, 26173932);
+        assertGe(balanceAfter - balanceBefore, 26173932);
```
tamaralipows, line 1040: 😂😂😂😂😂😂😂😂😂😂 no way!

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -328,13 +336,197 @@ impl SwapEncoder for EkuboSwapEncoder {
...
+                err
+            ))),
+        }
+    }
+}
+
+impl SwapEncoder for CurveSwapEncoder {
+    fn new(executor_address: String, chain: Chain) -> Result<Self, EncodingError> {
+        let native_token_curve_address =
+            Address::from_str("0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE").map_err(|_| {
```
tamaralipows: I guess this should be in some config somewhere since it depends on the chain?

tamaralipows: Can't we also use the `hex!` macro to avoid the error possibly being raised? Not sure

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -328,13 +336,197 @@ impl SwapEncoder for EkuboSwapEncoder {
...
+                EncodingError::FatalError("Invalid Curve native token address".to_string())
+            })?;
+        Ok(Self {
+            executor_address,
+            meta_registry_address: "0xF98B45FA17DE75FB1aD0e7aFD971b0ca00e379fC".to_string(),
```
tamaralipows: I also assume this won't be the same if curve gets deployed to other chains?

foundry/test/Constants.sol
```diff
@@ -30,7 +30,8 @@ contract Constants is Test, BaseConstants {
     address UNPAUSER = makeAddr("unpauser");
 
     // Assets
-    address ETH_ADDR = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
+    address ETH_ADDR_FOR_CURVE =
+        address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
tamaralipows, line 34: Also a bit confused, you said the zero address works sometimes?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -328,13 +336,197 @@ impl SwapEncoder for EkuboSwapEncoder {
...
+            },
+        }
+    }
+
+    fn get_coin_indexes(
+        &self,
+        pool_id: Address,
+        token_in: Address,
+        token_out: Address,
+    ) -> Result<(U8, U8), EncodingError> {
```
tamaralipows, line 428: To be honest I am quite concerned about making this request during encoding. How fast will this be?

Is there any reason we can't make this request in the executor itself? Is it just to save gas? How much gas does it actually save?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -328,13 +336,197 @@ impl SwapEncoder for EkuboSwapEncoder {
...
+                EncodingError::FatalError("Invalid Curve native token address".to_string())
+            })?;
+        Ok(Self {
+            executor_address,
+            meta_registry_address: "0xF98B45FA17DE75FB1aD0e7aFD971b0ca00e379fC".to_string(),
+            native_token_address: chain.native_token()?,
+            native_token_curve_address,
+        })
+    }
+    fn encode_swap(
```
tamaralipows:
```suggestion
    }
    
    fn encode_swap(
```

tamaralipows (commented):
Thank you! It mostly looks good - I just have one big question

foundry/test/TychoRouter.t.sol
dianacarvalho1: I KNOW 😂

foundry/test/Constants.sol
dianacarvalho1: yeah.. we have 2 pools that use ETH. one of the tests pass, the other one fails 😕 my solution was to use the this weird eeeEeeeE everywhere (because that works for everyone)

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: I think it should be very fast. We are also doing one request for the permit2 allowance/nonce and another for the token approvals 😬 but yeah, this whole code can be improved. I can test how much gas it would be to do this in the contract itself though 👀

tamaralipows: would love to know that! Thank you! otherwise it's a fine solution :)

config/protocol_specific_addresses.json
tamaralipows, line 1: Nice

src/encoding/evm/swap_encoder/swap_encoder_registry.rs
```diff
@@ -33,12 +36,27 @@ impl SwapEncoderRegistry {
...
             .get(&chain.name)
             .ok_or(EncodingError::FatalError("No executors found for chain".to_string()))?;
+
+        let protocol_specific_config: HashMap<String, HashMap<String, HashMap<String, String>>> =
+            serde_json::from_str(PROTOCOL_SPECIFIC_CONFIG)?;
+        let protocol_specific_config = protocol_specific_config
+            .get(&chain.name)
+            .ok_or(EncodingError::FatalError(
+                "No protocol specific config found for chain".to_string(),
+            ))?;
```
tamaralipows, line 49: Won't this fail for base and unichain atm since you didn't add those in there?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -328,13 +363,217 @@ impl SwapEncoder for EkuboSwapEncoder {
...
+            // TriPool
+            "0xbEbc44782C7dB0a1A60Cb6fe97d0b483032FF1C7" => Ok(U8::from(1)),
+            // STETHPool
+            "0xDC24316b9AE028F1497c275EB9192a3Ea0f67022" => Ok(U8::from(1)),
+            // TriCryptoPool
+            "0xD51a44d3FaE010294C616388b506AcdA1bfAAE46" => Ok(U8::from(3)),
+            // SUSDPool
+            "0xA5407eAE9Ba41422680e2e00537571bcC53efBfD" => Ok(U8::from(1)),
+            // FRAXUSDCPool
+            "0xDcEF968d416a41Cdac0ED8702fAC8128A64241A2" => Ok(U8::from(1)),
```
tamaralipows, line 398: I'm just assuming that if we ever need to support curve on another chain we will just need to be ok with a big refactor. Seems unreasonable to generalize for that now with all of these pools

tamaralipows (commented):
Thank you! I just have one main concern about the initialization

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: yeees I agree!

src/encoding/evm/swap_encoder/swap_encoder_registry.rs
dianacarvalho1: ah shit, you are right!!

dianacarvalho1: great catch!

dianacarvalho1: let me just add an empty dict there

tamaralipows (approved):
woohoo!

PR #155, opened 2025-04-07
feat: Refactor TychoEncoder into two: TychoRouterEncoder and TychoExecutorEncoder

dianacarvalho1:
We have a trait TychoEncoder and then two implementations: TychoRouterEncoder and TychoExecutorEncoder. This way we go a level above with the decision if it is a direct execution or if it should use the tycho router.

- Created two builders: one for each tycho encoder
- Delete ExecutorStrategyEncoder and move code straight into the TychoExecutorEncoder
- Add validate_solution to trait TychoEncoder
- Move group_swaps.rs a level up
- Update tests and usage cases

Doing this we get rid of all that weird stuff we were doing before in the builder


src/encoding/evm/tycho_encoders.rs
```diff
@@ -0,0 +1,1039 @@
...
+        }
+        Ok(())
+    }
+}
+
+/// Represents an encoder for one swap to be executed directly against an Executor. This is useful
+/// when you want to bypass the Tycho Router, use your own Router contract and just need the
+/// calldata for a particular swap. # Fields
+/// * `swap_encoder_registry`: Registry of swap encoders
+/// * `native_address`: Address of the chain's native token
```
tamaralipows: Wonky doc

tamaralipows (approved):
Thank you! I think this is exactly what we want! Looks straightforward, this diff is pretty wild though so I hope I didn't miss anything.

PR #156, opened 2025-04-08
feat: Select strategy in TychoRouterEncoder depending on the solution

dianacarvalho1:
- The tycho router address default is set at the EncoderBuilder level (not inside the strategies)
- Rename TychoCoreChain to TychoCommonChain
- Only take TychoCommonChain as an argument at the outermost level: EncoderBuilder. Everywhere else we use the execution Chain right away


tamaralipows (approved):
Thank you!!

PR #157, opened 2025-04-08
fix: Curve factory addresses are utf-8 encoded

dianacarvalho1:
I realised this when running the quickstart with only curve pools in tycho-simulation 0.82.2


tamaralipows (approved):
yikes

PR #158, opened 2025-04-09
feat: Optimize transfer to first pool in encoding

tamaralipows:
Review by commit.


src/encoding/evm/strategy_encoder/transfer_optimizations.rs
```diff
@@ -0,0 +1,95 @@
...
+};
+
+/// A trait that defines how the tokens will be transferred into the given pool given the solution.
+pub trait TransferOptimization {
+    /// Returns the transfer method that should be used for the given swap and solution.
+    ///
+    /// If the swap is for the in token of the solution and the protocol supports transferring
+    /// straight from the user, it will return `TransferType::Permit2Transfer` or
+    /// `TransferType::TransferFrom`.
+    fn get_transfer_method(&self, swap: Swap, given_token: Bytes, permit2: bool) -> TransferType {
```
dianacarvalho1: `get_transfer_type`

src/encoding/evm/strategy_encoder/transfer_optimizations.rs
```diff
@@ -0,0 +1,95 @@
...
+            token_out: dai(),
+            split: 0f64,
+        };
+        let strategy = MockStrategy {};
+        let transfer_method = strategy.get_transfer_method(swap.clone(), weth(), false);
+        assert_eq!(transfer_method, TransferType::TransferFrom);
+    }
+
+    #[test]
+    fn test_first_swap_transfer() {
```
dianacarvalho1: how could this be a first swap?? it would be at most another swap (not first no?)
Can you make a clearer example please? for example that the solution is DAI -> WETH -> USDC, so given token DAI and this current swap is WETH -> USDC, so it's the second hop

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -42,9 +46,19 @@ contract UniswapV4Executor is IExecutor, V4Router, ICallback {
...
             address callbackExecutor,
+            TransferType transferType,
             UniswapV4Executor.UniswapV4Pool[] memory pools
         ) = _decodeData(data);
 
+        // TODO move this into callback when we implement callback transfer type support
+        _transfer(
+            tokenIn,
+            msg.sender,
+            address(this), // irrelevant attribute
```
dianacarvalho1: can we add the why to the comment please?

src/encoding/evm/constants.rs
```diff
@@ -24,3 +24,14 @@ pub static IN_TRANSFER_OPTIMIZABLE_PROTOCOLS: LazyLock<HashSet<&'static str>> =
         set.insert("uniswap_v3");
         set
     });
+
+/// These protocols expect funds to be in the router at the time of swap.
+pub static PROTOCOLS_EXPECTING_FUNDS_IN_ROUTER: LazyLock<HashSet<&'static str>> =
```
dianacarvalho1, line 31: I'm starting to think that these hashsets should be in a config file (GROUPABLE_PROTOCOLS, IN_TRANSFER_OPTIMIZABLE_PROTOCOLS, PROTOCOLS_EXPECTING_FUNDS_IN_ROUTER) and we need to add really nice docs about them for integrators

dianacarvalho1 (commented):
oh boy this is brutal! giving a preliminary review only..

About your question
> TODO consider renaming TRANSFER_FROM and TRANSFER_PERMIT2 to include "pool" in the name

I would agree but using "protocol" instead. it's not always to the pool itself!

src/encoding/evm/strategy_encoder/transfer_optimizations.rs
tamaralipows: Just bad naming!

src/encoding/evm/constants.rs
tamaralipows: Will address this tomorrow!

foundry/src/TychoRouter.sol
```diff
@@ -449,23 +423,7 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
             _wrapETH(amountIn);
             tokenIn = address(_weth);
         }
-
-        uint256 initialBalance = tokenIn == address(0)
```
dianacarvalho1, line 453: you just removed the initial Balance checks entirely correct?

tamaralipows: correct. they don't make much sense anymore

dianacarvalho1: because we might bypass the router entirely right? makes sense makes sense

foundry/src/TychoRouter.sol
```diff
@@ -449,23 +423,7 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
...
-
         amountOut = _splitSwap(amountIn, nTokens, swaps);
-        uint256 currentBalance = tokenIn == address(0)
-            ? address(this).balance
-            : IERC20(tokenIn).balanceOf(address(this));
-
-        uint256 amountConsumed = initialBalance - currentBalance;
-
-        if (tokenIn != tokenOut && amountConsumed != amountIn) {
-            revert TychoRouter__AmountInDiffersFromConsumed(
```
dianacarvalho1, line 465: then we can delete this error as well TychoRouter__AmountInDiffersFromConsumed no?

foundry/src/executors/TokenTransfer.sol
```diff
@@ -16,9 +16,17 @@ contract TokenTransfer {
         // Assume funds are in the TychoRouter - transfer into the pool
         TRANSFER,
         // Assume funds are in msg.sender's wallet - transferFrom into the pool
-        TRANSFERFROM,
+        TRANSFER_FROM,
```
dianacarvalho1: reminder to rename this! and let's make it consistent and have "from" in both or in neither like
`TRANSFER_TO_PROTOCOL` and `TRANSFER_TO_ROUTER` for example

src/encoding/models.rs
```diff
@@ -96,6 +96,27 @@ pub struct Transaction {
...
+/// * `None`: No transfer is needed. Tokens are already in the pool.
+#[repr(u8)]
+#[derive(Clone, Debug, PartialEq)]
+pub enum TransferType {
+    Transfer = 0,
+    TransferFrom = 1,
+    Permit2Transfer = 2,
+    TransferToRouter = 3,
+    Permit2TransferToRouter = 4,
+    None = 5,
```
dianacarvalho1: make sure the naming matches with the ones in solidity!

dianacarvalho1 (commented):
Thank you @tamaralipows ! 🙏🏼 

One thing that bothers me is that in some executors we are doing things like
````
(data.length != 94)
````
and in others we are not checking at all 😕  could we standardize this?

foundry/src/executors/TokenTransfer.sol
tamaralipows: > and let's make it consistent and have "from" in both or in neither like
`TRANSFER_TO_PROTOCOL` and `TRANSFER_TO_ROUTER` for example

I think then I'll add the "FROM" to both of them, because `TRANSFER_TO_PROTOCOL` is different from `TRANSFER_FROM_TO_PROTOCOL`

src/encoding/evm/constants.rs
tamaralipows: We decided not to address this since it would involve either json parsing in the constants file (possible raising an error) or parsing wherever the constants are use, making it more difficult to document these constants.

foundry/src/TychoRouter.sol
tamaralipows: yes. we aim to bypass the router entirely actually!

dianacarvalho1 (approved):
peerfecctionnn ✨

PR #160, opened 2025-04-12
feat: No more fee taking

tamaralipows:
- This required an extra transfer into the router at the end of the swap sequence, costing an unnecessary 40k gas or more for certain protocols.


dianacarvalho1 (approved):
ai 🧹 ✨ 💆🏼‍♀️ 
Thank you 🙏🏼

PR #162, opened 2025-04-15
feat: Use TokenTransfer optimization helper in Ekubo

tamaralipows:


foundry/src/executors/EkuboExecutor.sol
```diff
@@ -169,9 +186,11 @@ contract EkuboExecutor is IExecutor, ILocker, IPayer, ICallback {
                 mstore(free, shl(224, 0x0c11dedd))
                 mstore(add(free, 4), token)
                 mstore(add(free, 36), shl(128, amount))
+                mstore(add(free, 52), shl(96, sender))
+                mstore(add(free, 72), shl(248, transferType))
```
tamaralipows, line 190: This was the key part, we needed to assembly-magic the transfer type into the calldata for the `pay` method [of the core contract](https://etherscan.io/address/0xe0e0e08a6a4b9dc7bd67bcb7aade5cf48157d444#code), which doesnt accept bytes as its input, but apparently that doesn't matter. It then does a `calldatacopy` back to our method, where we can decode it easily

foundry/src/executors/EkuboExecutor.sol
```diff
@@ -11,21 +11,28 @@ import {SafeTransferLib} from "@solady/utils/SafeTransferLib.sol";
...
+    TokenTransfer
+{
     error EkuboExecutor__InvalidDataLength();
     error EkuboExecutor__CoreOnly();
     error EkuboExecutor__UnknownCallback();
 
     ICore immutable core;
 
-    uint256 constant POOL_DATA_OFFSET = 56;
+    uint256 constant POOL_DATA_OFFSET = 77;
```
dianacarvalho1, line 29: why did it go from 55 to 77? aren't we just adding the TransferType?

dianacarvalho1: aaah I see we are passing the msg.sender too (+ 20) ok ok

foundry/src/executors/EkuboExecutor.sol
```diff
@@ -169,9 +186,11 @@ contract EkuboExecutor is IExecutor, ILocker, IPayer, ICallback {
                 mstore(free, shl(224, 0x0c11dedd))
                 mstore(add(free, 4), token)
                 mstore(add(free, 36), shl(128, amount))
+                mstore(add(free, 52), shl(96, sender))
+                mstore(add(free, 72), shl(248, transferType))
 
-                // if it failed, pass through revert
-                if iszero(call(gas(), target, 0, free, 52, 0, 0)) {
+                // 4 (selector) + 32 (token) + 16 (amount) + 20 (recipient) + 1 (transferType) = 73
```
dianacarvalho1: isn't it the sender here?

foundry/src/executors/EkuboExecutor.sol
dianacarvalho1: oh uau, that's obscure hein?

dianacarvalho1 (approved):
Thank you sooooo much @tamaralipows 🙏🏼 🙏🏼  This is amaaaaziingg 👏🏼👏🏼👏🏼 Only have 1 small doc comment

PR #163, opened 2025-04-15
feat: Support out transfer straight to the receiver

dianacarvalho1:
- The out transfer is now a responsibility of the Executors -> remove this from router methods
- Also adding a check that the receiver got the full amount out
- In encoding, if it is the last swap (for split swaps also), pass the receiver as the trade receiver and not the router address
- Adapt curve and uniswap v4 to support out transfer (and define receiver)
- If the trade is cyclical, we can't do the balance check for the correctness of the amountOut transfers
- In encoding, if it is a wrap trade, the transfer should be just a normal Transfer (not TransferFrom nor a Permit2Transfer)


foundry/src/TychoRouter.sol
```diff
@@ -408,9 +411,6 @@ contract TychoRouter is AccessControl, Dispatcher, Pausable, ReentrancyGuard {
         address receiver,
         bytes calldata swaps
     ) internal returns (uint256 amountOut) {
-        if (receiver == address(0)) {
-            revert TychoRouter__AddressZero();
-        }
```
tamaralipows: Why did we get rid of this check? We still need the receiver to not be zero in order to perform a check

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -286,16 +289,23 @@ impl StrategyEncoder for SequentialSwapStrategyEncoder {
...
+                .get_swap_encoder(&protocol)
                 .ok_or_else(|| {
                     EncodingError::InvalidInput(format!(
                         "Swap encoder not found for protocol: {}",
-                        grouped_swap.protocol_system
+                        protocol
                     ))
                 })?;
 
+            let receiver = if i == grouped_swaps.len() - 1 && !unwrap {
```
tamaralipows: I think this could use a comment, like if it's the final swap and we are unwrapping, receiver should be the router

tamaralipows: Also, perhaps we should call this `swap_receiver`? To not be confused with the receiver of the entire solution?

tamaralipows (commented):
Thank you, I think this is it! I just have a main concern about the address zero check

foundry/src/TychoRouter.sol
dianacarvalho1: yes, true. now we need it! sorry sorry, too many iterations happened 😅 I will bring it back

tamaralipows (approved):
Thank you!

PR #164, opened 2025-04-16
feat: Support in between swaps transfer optimizations 

dianacarvalho1:
Added tests cases for:
1. uniswap v3 --(straight to the pool)--> uniswap v2 
2. uniswap v3 --(to the router)--(to the pool)--> uniswap v3
3. uniswap v3 --(to the router)--(to the pool)--> curve
4. balancer v2 --(straight to the pool)--> uniswap v2


src/encoding/evm/constants.rs
```diff
@@ -23,18 +23,21 @@ pub static IN_TRANSFER_OPTIMIZABLE_PROTOCOLS: LazyLock<HashSet<&'static str>> =
...
         set.insert("ekubo_v2");
         set
     });
 
-/// These protocols expect funds to be in the router at the time of swap.
-pub static PROTOCOLS_EXPECTING_FUNDS_IN_ROUTER: LazyLock<HashSet<&'static str>> =
+// These protocols do not support chained swaps from the same protocol. This is the case for uniswap
+// v3 because of the callback logic. The only way for this to work it would be to call the second
+// swap during the callback of the first swap. This is currently not supported.
+pub static PROTOCOLS_NOT_OPTIMIZED_CHAINED_SWAPS: LazyLock<HashSet<&'static str>> =
```
dianacarvalho1: not a super fan of this name 😬

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -300,12 +303,31 @@ impl StrategyEncoder for SequentialSwapStrategyEncoder {
                     ))
                 })?;
```
dianacarvalho1, line 305: the meat starts here

src/encoding/evm/strategy_encoder/transfer_optimizations.rs
```diff
@@ -16,37 +17,40 @@ pub trait TransferOptimization {
...
 
-        // In the case of wrapping, check if the swap's token in is the wrapped token to
-        // determine if it's the first swap. Otherwise, compare to the given token.
         let is_first_swap = swap.token_in == given_token;
 
         if swap.token_in == native_token {
             // Funds are already in router. All protocols currently take care of native transfers.
             TransferType::None
         } else if (swap.token_in == wrapped_token) && wrap {
+            // Wrapping already happened in the router so we can just use a normal transfer.
```
dianacarvalho1, line 31: more meat here

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -1148,164 +1172,466 @@ mod tests {
...
 
-    #[test]
-    fn test_sequential_swap_strategy_encoder_complex_route() {
-        // Note: This test does not assert anything. It is only used to obtain integration test
-        // data for our router solidity test.
-        //
-        // Performs a split swap from WETH to USDC though WBTC and DAI using USV2 pools
-        //
-        //   WETH ───(USV2)──> WBTC ───(USV2)──> USDC
+    mod sequential {
```
tamaralipows, line 1172: Love the organization ❤️

src/encoding/evm/strategy_encoder/strategy_encoders.rs
```diff
@@ -300,12 +303,31 @@ impl StrategyEncoder for SequentialSwapStrategyEncoder {
                     ))
                 })?;
 
-            // if it is the last swap and there isn't an unwrap at the end, we can set the receiver
-            // to the final user
-            let swap_receiver = if i == grouped_swaps.len() - 1 && !unwrap {
-                solution.receiver.clone()
+            let in_between_swap_optimization = next_in_between_swap_optimization;
+            let next_swap = grouped_swaps.get(i + 1);
```
tamaralipows, line 307: This is quite a lot of meat, what do you think about moving this into the `transfer_optimizations` file? Not sure if it makes sense or if we would need to input too make things to that method tbh.

src/encoding/evm/constants.rs
```diff
@@ -23,18 +23,21 @@ pub static IN_TRANSFER_OPTIMIZABLE_PROTOCOLS: LazyLock<HashSet<&'static str>> =
     LazyLock::new(|| {
         let mut set = HashSet::new();
         set.insert("uniswap_v2");
+        set.insert("sushiswap_v2");
+        set.insert("pancakeswap_v2");
         set.insert("uniswap_v3");
+        set.insert("pancakeswap_v3");
```
tamaralipows: I know this isn't your doing but maybe can you add to the docstring here that any protocols not in this set will expect the funds to be in the router?

src/encoding/evm/constants.rs
```diff
@@ -23,18 +23,21 @@ pub static IN_TRANSFER_OPTIMIZABLE_PROTOCOLS: LazyLock<HashSet<&'static str>> =
...
         set.insert("ekubo_v2");
         set
     });
 
-/// These protocols expect funds to be in the router at the time of swap.
-pub static PROTOCOLS_EXPECTING_FUNDS_IN_ROUTER: LazyLock<HashSet<&'static str>> =
+// These protocols do not support chained swaps from the same protocol. This is the case for uniswap
```
tamaralipows: Don't think this docstring makes sense - it's not just from the same protocol that this cant be optimized? Also USV2 to USV3 can't really be optimized

src/encoding/evm/constants.rs
tamaralipows: It's a bit unclear to me here - does this mean that the receiver of these protocols can't be set?  That it's the router by default and that's that?

Does curve let you set the receiver? I thought not. Or, now it doesn't matter since there is an out transfer at the end? very confused, maybe we should have a call

src/encoding/evm/strategy_encoder/transfer_optimizations.rs
```diff
@@ -16,37 +17,40 @@ pub trait TransferOptimization {
...
-                // Transfer from swapper to pool using permit2.
-                TransferType::TransferPermit2ToProtocol
-            } else {
-                // Transfer from swapper to pool.
-                TransferType::TransferFromToProtocol
-            }
-        } else if is_first_swap && funds_expected_in_router {
-            if permit2 {
+            // first swap
+        } else if is_first_swap {
```
tamaralipows: kindof redundant comment

tamaralipows (commented):
Hey hey, this was a lot of work, thank you! I'm confused about a lot though, maybe we can have a quick call when I start so you can walk me through this?

src/encoding/evm/strategy_encoder/strategy_encoders.rs
dianacarvalho1: yeah I thought about that too. but this is needed literally only in the sequential strategy 😕 so I thought it was an overkill.

src/encoding/evm/constants.rs
dianacarvalho1: really?? 🤔 oh I didn't realise that. but yes, I understand why. you are right. let me improve this. literally uniswap v3 can't get funds send to it before hand. true true

src/encoding/evm/constants.rs
dianacarvalho1: > does this mean that the receiver of these protocols can't be set? That it's the router by default and that's that?

it has to be the router yes.

dianacarvalho1: > Does curve let you set the receiver?

now it does. but it does an extra transfer in the executor

src/encoding/evm/strategy_encoder/strategy_encoders.rs
tamaralipows: Hmm I think a good reason for a separate method is not just that it can be reused in other places, but also just for pure organization and testing

dianacarvalho1: I understand.. but tbh I can't see a better way to do this now tbh 😕 maybe cooldown task to go back to think about this?

tamaralipows (approved):
Big baby. Thank you!

PR #165, opened 2025-04-16
[DO NOT MERGE] feat: (WIP) Use permit2 signature transfer

tamaralipows:
- This should be more fitted to our use case, since we are recreating the signature at every swap transaction. This should be more gas efficient, since no slots need to be changes on the permit2 contract itself.

Yes I have deleted every file to not have to deal with compilation issues. Of course these should be put back before ever merging.


tamaralipows:
Not pursuing this atm.

PR #166, opened 2025-04-17
feat: Do not use V4Router for uniswap v4

dianacarvalho1:
Do our own implementation. I took inspiration from their [Router](https://github.com/Uniswap/v4-periphery/blob/main/src/V4Router.sol) 
In the end this is much cleaner than I expected and I was able to do a few improvements:
- we don't need to use actions. using function selectors and delegate call instead
- we don't need to convert into V4 Router types
- we don't need to check balances to get the amount out, we can just use the returned value

Support using the TransferType in uniswap v4


foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -199,15 +182,255 @@ contract UniswapV4Executor is IExecutor, V4Router, ICallback, TokenTransfer {
...
+    /**
+     * @notice Handles the unlock callback from the pool manager. This is used for swaps against the executor directly (bypassing the router).
+     */
+    function unlockCallback(bytes calldata data)
+        external
+        poolManagerOnly
+        returns (bytes memory)
+    {
+        return _unlockCallback(data);
+    }
```
tamaralipows, line 196: I would say this isn't even useful for swaps against the executor, since this needs to be called through delegatecall, the router calling this would need their own callback implemented. We have also decided not to support this, so I'd remove it. I understand it is useful for testing, but I think our router level integration tests cover this anyway. What do you think?

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -102,51 +111,22 @@ contract UniswapV4Executor is IExecutor, V4Router, ICallback, TokenTransfer {
...
-        calculatedAmount = tokenOutBalanceAfter - tokenOutBalanceBefore;
-
-        return calculatedAmount;
-    }
-
-    // necessary to convert bytes memory to bytes calldata
-    function executeActions(bytes memory unlockData) public {
-        // slither-disable-next-line unused-return
-        poolManager.unlock(unlockData);
+        return amountOut;
```
tamaralipows, line 129: I am so surprised at how much simpler this method is! It's amazing

foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -199,15 +182,255 @@ contract UniswapV4Executor is IExecutor, V4Router, ICallback, TokenTransfer {
...
+        }
 
-    function _pay(Currency token, address, uint256 amount) internal override {
-        IERC20(Currency.unwrap(token)).safeTransfer(
-            address(poolManager), amount
+        if (executor == address(0)) {
+            executor = address(this);
+        }
+        // slither-disable-next-line low-level-calls
+        (bool success, bytes memory returnData) = executor.delegatecall(data);
```
tamaralipows, line 218: I would just add a comment here on which method exactly this is expected to call. I understand this is the swap data with the swap method selectors that we have encoded earlier, right? It's a bit obscure like this though...

tamaralipows (approved):
Thank you! It's quite a lot more code added but it feels nice to have this now in all one contract and not rely on the V4 router!

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: > I would say this isn't even useful for swaps against the executor, since this needs to be called through delegatecall,

this is not true :/ this is the default callback function that the pool manager expects to call. so if we swap against the executor directly it goes here and then in the `_unlockCallback` that then makes the delegate call to `swapExactInputSingle` or `swapExactInput`.

I would like to keep it tbh. We only don't support this for univ3 (because we can't) and I think the tests at the executor level are more thorough than the ones in the router..

tamaralipows: > this is not true :/ this is the default callback function that the pool manager expects to call. so if we swap against the executor directly it goes here and then in the _unlockCallback that then makes the delegate call to swapExactInputSingle or swapExactInput.

What do you mean if we swap against the executor directly?
If we do this via a `delegatecall` (which is the only way calling our executors is supported) then the pool manager looks for the method in the TychoRouter (or whoever is calling's) contract, not in here...
If you want to just keep it for testing though then that's a different story

dianacarvalho1: I mean that the executor works on it's own (and we use this for testing but it could be used for real as well)

tamaralipows: Slack discussion: We have decided to support direct-against-executor swaps for all executors. Will attempt to make this work with USV3 today

tamaralipows (approved):
Thank you!

PR #168, opened 2025-04-18
fix: Configurable fee on USV2 executor.

tamaralipows:
- This fee is not the same depending on the fork. For example, Pancake V2 uses a 0.25% fee, while the USV2 fee is 0.3%. We were wrongly hard-coding this fee to 0.3%.

### DIFF
--- a/foundry/src/executors/UniswapV2Executor.sol
+++ b/foundry/src/executors/UniswapV2Executor.sol
@@ -10,6 +10,7 @@ error UniswapV2Executor__InvalidDataLength();
 error UniswapV2Executor__InvalidTarget();
 error UniswapV2Executor__InvalidFactory();
 error UniswapV2Executor__InvalidInitCode();
+error UniswapV2Executor__InvalidFee();
 
 contract UniswapV2Executor is IExecutor, TokenTransfer {
     using SafeERC20 for IERC20;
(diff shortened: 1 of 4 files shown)


foundry/src/executors/UniswapV2Executor.sol
```diff
@@ -26,6 +28,10 @@ contract UniswapV2Executor is IExecutor {
         }
         factory = _factory;
         initCode = _initCode;
+        if (_feeBps > 10000) {
```
tamaralipows: This seems like too lenient of a restriction but I don't know where to arbitrarily cut this off otherwise.

dianacarvalho1: huum I agree this is too lenient 😕 but then why have a restriction at all?
If we want a check then I would check if they are indeed one of the two values we expect (30 and 25)

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 one comment!

foundry/src/executors/UniswapV2Executor.sol
tamaralipows: Just worried that another fork will come along with a different value and then we would need to change this yet again :/ I guess I can check that it's not greater than 30? can't imagine a fork being more expensive than the OG USV2

dianacarvalho1: yes I understand.. but then this check is a bit redundant 😕 but yes I agree with this assumption 

> I guess I can check that it's not greater than 30? can't imagine a fork being more expensive than the OG USV2

dianacarvalho1 (approved):
Thank you! 🙏🏼

PR #169, opened 2025-04-21
chore: Improve tests

dianacarvalho1:
In both encoding and contracts:
- Make sure that the tests names/encoder match what is happening (we had test_split_... that was only a single swap).
- Made modules inside the encodings tests
  - In swap_encoders: there is one module per protocol
  - In strategy_encoders: there is one module per strategy and an extra protocol_integration module. Inside the sequential module there is another module called chained_swaps
- In contracts, each strategy has its own file. Integration tests per strategy should also be here. Renamed TychoRouterIntegration to TychoRouterProtocolIntegration. Only protocol integration tests should be in this file


tamaralipows (approved):
Beautiful! This is a lot cleaner now, thank you!

PR #172, opened 2025-04-22
feat: Make USV3 callback work with direct executor call

tamaralipows:
- We have decided to now support calling our executors normally (without using delegatecalls) since they now support specifying token transfers.
- This was disabled for UniswapV3 in the past also because of data decoding issues. This seems to be the solution, though.


dianacarvalho1 (commented):
oh uau shocked that this simple thing actually worked 👀  veeery nice! Thank you 🙏🏼

dianacarvalho1 (approved):

PR #173, opened 2025-04-22
test: 5-hop Multi protocol integration test

tamaralipows:
- Main goal: make sure all is fine with token transfers between protocols. ✅ 
- You really only need to review `testMultiProtocolIntegration` and `test_multi_protocol`. The rest of the diff is because of the last point down there.
- Needed to add ekubo and uniswap v4 to callback-limited protocols.
- I had to bump the fork block in all of our integration tests: The way it was before meant that certain integration tests were using certain executor addresses, and others were using different ones, because of the redeployment. This was a pain to account for on the rust side. Instead, all tests now use an Ekubo-compatible fork block. Values needed to be updated because of price changes between blocks. This was a **huge pain** but I think will save us pain in the future.

I put USV2 first not because I care about USV2 (it's already thoroughly tested), but because I wanted Balancer to be second after a simple non-complicated pool where receiver can be easily specified. I was suspicious about transferring directly to Balancer before the swap. Doesn't seem like an issue after all though.
```
DAI ─(USV2)─> WETH ─(bal)─> WBTC ─(curve)─> USDT ─(ekubo)─> ETH ─(USV4)─> USDC
```


dianacarvalho1 (approved):
aaaaaffff shiiittt I understand what happened here 😬 this is painful indeeedd! Ai I wish we could just easily make all these tests better 😕 soon I hope 🤞🏼 
Thank you so much for this 🙏🏼

PR #177, opened 2025-04-23
feat: Gas optimizations feature

dianacarvalho1:
* Router endpoints:
    * Added new endpoints: singleSwap and sequentialSwap
    * Renamed old swap to splitSwap
    * For each type there is a permit2 and a “normal” endpoint. Like singleSwap and singleSwapPermit2
        * The permit2 methods uses permit2 to transfer the funds while the other methods uses normal token transfers with tranferFrom or assume that the tokens are already in the router

* Callback logic:
    * Refactor callback to use transient storage to store the executor address. This way we don’t need hardcoded callback methods in the TychoRouter (only uniswap v4 is still done this way)
    * Support returning data from the callbacks
* Fee taking ability has been removed
* Executors are now responsible for all tokens transfers (in and out). Add TokenTransfer helper contract for this.
    * Now we pass a TransferType value to indicate which type of transfer (if any) should the executor perform for the token in.
    * The token out transfer is defined by the value of the receiver address. It can be the TychoRouter, the next pool or the user (as a final transfer). This is all decided at encoding.
    * In the TokenTransfer _transfer method, we are doing transferFrom sender, which should only ever contain the original msg.sender, and cannot be altered by outside encoding
* Security checks:
    * Added a check that the receiver got the full amount out in the main methods
    * Removed check that all tokens in the router have been used (since the tokens may not pass through the router on the way in anymore)
* Executors
    * Uniswap v4 executor was refactored to not use the V4Router anymore
    * Curve executor now performs the token out transfer
    * Ekubo executor was refactored to handle another receiver (that not msg sender) and the callback with transient storage


tamaralipows (approved):
😭 thank you

PR #180, opened 2025-04-24
fix: Remove tload from executor

tamaralipows:
- Store the executor address when deploying instead.
- We would like to keep all instances of tload and tstore within the callback mechanism of our main TychoRouter contract for security reasons and to prevent any unexpected behaviour
- This way it's easy to reason that UniswapV4Executor will only ever execute a delegatecall to itself. Before it could in theory execute a delegatecall to any address. One had to look at all occurences of tstore(0, x) to ensure the address was constrained.


dianacarvalho1 (approved):
Veeery nice! I like this! ✨  thank you!

PR #182, opened 2025-04-26
feat: Clear transient storage after swap call

tamaralipows:
- This was originally remaining if no callback was performed, possible resulting in unexpected behaviour and an increased attack surface.
- Also specify nonzero slot for transient storage in order to reduce the risk of dangerous slot collision.


dianacarvalho1 (approved):
this is a great idea! Thank you 🙏🏼

PR #184, opened 2025-04-29
chore: Write encoding rust calldata to file and read in solidity test

dianacarvalho1:
This way we can automatically replace the calldata when something changes. We don't need to manually replace the string ourselves.

It's not a perfect solution but I already think it's much better than what we had before! it should also make things easier for integrators! 

TODO (after review):
- Mention this test logic in the docs


tamaralipows (approved):
Words REALLY cannot describe how HAPPY this makes me

PR #192, opened 2025-05-05
feat: Choose single strategy if it's a grouped swap of the same protocol

dianacarvalho1:
Add and update tests


foundry/test/assets/calldata.txt
```diff
@@ -3,26 +3,26 @@ test_single_encoding_strategy_ekubo:20144a07000000000000000000000000000000000000
...
...00000000cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000000000000000000000000000000000003b9aca000000000000000000000000000000000000000000000000000000000068402a3900000000000000000000000000000000000000000000000000000000000000000000000000000000000000003ede3eca2a72b3aecc820e955b36f38437d01395000000000000000000000000000000000000000000000000000000006818a44100000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000026000000000000000000000000000000000000000000000000000000000000000419facefe3089c4d8177dc12ace73e275369649112be2bead25fcd035707a50de67aed574499636214e89fd88bf2907096140c552cecf3a93305d37b8e922166e61c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000086f62849f9a0b5bf2913b396098f7c7019b51a820aa0b86991c6218b36c1d19d4a2e9eb0ce3606eb486982508145454ce325ddbe47a25d4ec3d23119330002cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc20000000000000000000000000000000000000000000bb800003c6982508145454ce325ddbe47a25d4ec3d23119330061a80001f40000000000000000000000000000000000000000000000000000
```
dianacarvalho1, line 6: this is the only diff that is actually different! It was using the sequential swap selector and now it is using the single swap.
The other diffs are permit2 changes

src/encoding/evm/tycho_encoders.rs
```diff
@@ -77,7 +78,20 @@ impl TychoEncoder for TychoRouterEncoder {
         let mut transactions: Vec<Transaction> = Vec::new();
         for solution in solutions.iter() {
             self.validate_solution(solution)?;
-            let (contract_interaction, target_address) = if solution.swaps.len() == 1 {
+
+            let protocols: HashSet<String> = solution
```
dianacarvalho1, line 82: I did this instead of actually moving the grouping of swaps. I think this was easier

louise-poole (approved):

PR #193, opened 2025-05-08
chore: Add script to remove executor addresses from Tycho Router

dianacarvalho1:
🧹 🧹 🧹  also removed the old executors in ethereum from TychoRouter


tamaralipows (approved):
THANK YOU!

PR #194, opened 2025-05-09
feat: Remove special handling of the Univ4 callback

dianacarvalho1:
The problem was that the pool manager was expecting an ABI encoded result to be returned and we were not returning that (we were returning just a result) 

✨ ✨ ✨ Special thanks to Max for figuring this out ✨ ✨ ✨

### DIFF
--- a/foundry/src/executors/UniswapV4Executor.sol
+++ b/foundry/src/executors/UniswapV4Executor.sol
@@ -184,8 +184,11 @@ contract UniswapV4Executor is
         external
         returns (bytes memory)
     {
-        verifyCallback(data);
-        return _unlockCallback(data);
+        bytes calldata stripped = data[68:];
+        verifyCallback(stripped);
+        // Our general callback logic returns a not ABI encoded result.
+        // However, the pool manager expects the result to be ABI encoded. That is why we need to encode it here again.
+        return abi.encode(_unlockCallback(stripped));
     }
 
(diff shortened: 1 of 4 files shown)


foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -184,8 +184,11 @@ contract UniswapV4Executor is
         external
         returns (bytes memory)
     {
-        verifyCallback(data);
-        return _unlockCallback(data);
+        bytes calldata stripped = data[68:];
```
tamaralipows, line 187: I'm curious if instead of stripping manually we could do sth like
```
bytes memory decodedData = abi.decode(data, (bytes));
```
?

dianacarvalho1: good question 🤔 I need to try! 🔍

dianacarvalho1: oh no no no no. aboooorttt. to do that we need to transfor the bytes from calldata to memory and then we get our old friend:
`Invalid type for argument in function call. Invalid implicit conversion from bytes memory to bytes calldata requested.`

tamaralipows: oh f-----

tamaralipows (approved):
Thank youuuu!!

PR #195, opened 2025-05-13
feat: Arbitrage balance check and explicitly handle the TransferType.NONE case

dianacarvalho1:


foundry/src/executors/TokenTransfer.sol
```diff
@@ -45,7 +45,9 @@ contract TokenTransfer {
         uint256 amount,
         TransferType transferType
     ) internal {
-        if (transferType == TransferType.TRANSFER_TO_PROTOCOL) {
+        if (transferType == TransferType.NONE) {
+            return;
+        } else if (transferType == TransferType.TRANSFER_TO_PROTOCOL) {
```
tamaralipows, line 51: Max also suggested to revert if "else"

> I recommend handling the NONE case explicitely in _transfer with a comment explaining it and reverting with InvalidTransferType in the else branch of _transfer .

dianacarvalho1: AH true! forgot about that! thank you

tamaralipows (approved):
Thank youuu

PR #198, opened 2025-05-14
feat: Restrict transferFrom

dianacarvalho1:
The issue is that the executors have "access to the user funds" unnecessarily from the TokenTransfer generalisation. The transfer types
```
TRANSFER_FROM_TO_PROTOCOL,
TRANSFER_PERMIT2_TO_PROTOCOL,
TRANSFER_FROM_TO_ROUTER,
TRANSFER_PERMIT2_TO_ROUTER,
```
are actually only meant to be used once per solution! This proposal is:
- Create a RestrictTransferFrom contract that is allowed to transfer only up to the amount in
  - In the beginning of the swap, we set in transient storage the values necessary for the in transfer 
  - Then we transfer inside the executors 
- The actions were simplified into:
````
    enum TransferType {
        TransferFrom,
        Transfer,
        None
    }
````

### CODE REVIEW

File: foundry/src/executors/UniswapV3Executor.sol [hunk 1]
```diff
@@ -98,20 +104,24 @@ contract UniswapV3Executor is IExecutor, ICallback, TokenTransfer {
 
         address tokenIn = address(bytes20(msgData[132:152]));
 
-        // Transfer type does not exist
-        if (uint8(msgData[175]) > uint8(TransferType.NONE)) {
-            revert UniswapV3Executor__InvalidTransferType(uint8(msgData[175]));
-        }
-
-        TransferType transferType = TransferType(uint8(msgData[175]));
+        bool inTransferNeeded = bool(msgData[175]);
```
tamaralipows: This name is a bit confusing - this means `inTransferFromNeeded`, right? Trying to also distinguish in the case where we are wrapping in our router. In this case, an "in transfer" of WETH to the pool is needed, but not an "in transferFrom"

foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -51,7 +51,8 @@ contract UniswapV3Executor is IExecutor, ICallback, TokenTransfer {
             address receiver,
             address target,
             bool zeroForOne,
-            TransferType transferType
+            bool inTransferNeeded,
+            bool inBetweenSwapsTransferNeeded
```
tamaralipows: I think it's not necessary to leak info here about first swap or in swap, etc... we can simply call it transferFrom / transfer needed (like mentioned in other comment)

dianacarvalho1: huuumm not really. we need to know which one it is. they are different types of transfers. I don think we need both here

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -282,7 +284,7 @@ impl SwapEncoder for BalancerV2SwapEncoder {
             component_id,
             bytes_to_address(&encoding_context.receiver)?,
             approval_needed,
-            (encoding_context.transfer_type as u8).to_be_bytes(),
+            encoding_context.transfer,
```
tamaralipows: I would say this isn't necessary for balancer or curve since there will never be a transfer into the protocol ever ever ever

foundry/src/executors/UniswapV3Executor.sol
tamaralipows: what do you mean? I am talking about naming it `transferFromRequired` and `transferRequired` instead of having it called in transfer and in between swaps transfer. See latest commit for what I mean

dianacarvalho1: aaaaaahhhhhhhh I missunderstood then, sorry. I'm fine with renaming

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: oooohhhhh 🤔 you are absolutely right

foundry/src/OneTransferFromOnly.sol
```diff
@@ -0,0 +1,94 @@
...
+import "@interfaces/IExecutor.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@permit2/src/interfaces/IAllowanceTransfer.sol";
+
+error OneTransferFromOnly__AddressZero();
+error OneTransferFromOnly__MultipleTransferFrom();
+
+/**
+ * @title OneTransferFromOnly - Restrict to one transferFrom on approved params per swap
+ * @author PropellerHeads Devs
```
dianacarvalho1: oh.. I don't think we are using this `author` anywhere else no? 🤔

src/encoding/evm/strategy_encoder/transfer_optimizations.rs
```diff
@@ -155,146 +149,68 @@ mod tests {
...
-        let optimization = TransferOptimization::new(eth(), weth(), true, false, router_address());
-        let transfer_method = optimization.get_transfer_type(swap.clone(), weth(), false, false);
-        assert_eq!(transfer_method, TransferType::TransferPermit2ToProtocol);
+    fn zero_address() -> String {
+        Bytes::from([0u8; 20]).to_string()
     }
 
-    #[test]
-    fn test_first_swap_transfer_from() {
+    #[rstest]
```
dianacarvalho1, line 146: I collapsed these tests into cases

dianacarvalho1 (commented):
Great job us!! 👏🏼 👏🏼  I like this actually

foundry/src/OneTransferFromOnly.sol
tamaralipows: I copy pasted this from the dispatcher :/

dianacarvalho1: ooooohhhh

dianacarvalho1: it's the only place we have it though? let's remove it?

tamaralipows (approved):
🙏

foundry/src/executors/CurveExecutor.sol
```diff
@@ -64,25 +64,16 @@ contract CurveExecutor is IExecutor, TokenTransfer {
...
-            // Thus, this call will only ever be used to transfer funds from the user into the router.
-            address(this),
-            amountIn,
-            transferType
-        );
-
-        if (tokenApprovalNeeded && tokenIn != nativeToken) {
+        if (approvalNeeded && tokenIn != nativeToken) {
             // slither-disable-next-line unused-return
             IERC20(tokenIn).forceApprove(address(pool), type(uint256).max);
```
dianacarvalho1, line 74: oh uau we had a bug here hein? 😬

PR #202, opened 2025-05-19
feat(curve): Get coin indexes from static attributes

dianacarvalho1:
This way we don't need extra RPC calls to get these values ✨


tamaralipows (approved):
Very pretty!

PR #203, opened 2025-05-21
feat: Separate encoding swaps from encoding txs

dianacarvalho1:
This way the user is responsible for encoding the Tycho Router method inputs that are used as guardrails in execution.

Interface changes:
- Create EncodedSolution
- StrategyEncoder
  - don't need to know have permit2 or token_in_already_in_router as attributes anymore
  - encode_strategy returns EncodedSolution now (no method encoding done here now)
- TychoEncoder
  - add encode_solution() method. This is the recommended method for users
  - needs to have permit2, token_in_already_in_router and router_address as attributes
  - permit creation is made in the router now
  - Remove encode_full_calldata from the TychoEncoder trait
  - Make the TychoExecutorEncoder use encode_solutions instead of encode_full_calldata

Also:
- create encoding_utils.rs
- update all tests
- Remove slippage and expected_amount from Solution


src/encoding/evm/tycho_encoders.rs
```diff
@@ -26,13 +30,18 @@ use crate::encoding::{
...
 #[derive(Clone)]
 pub struct TychoRouterEncoder {
     single_swap_strategy: SingleSwapStrategyEncoder,
     sequential_swap_strategy: SequentialSwapStrategyEncoder,
     split_swap_strategy: SplitSwapStrategyEncoder,
     native_address: Bytes,
     wrapped_address: Bytes,
+    router_address: Bytes,
+    token_in_already_in_router: bool,
+    permit2: Option<Permit2>,
```
dianacarvalho1, line 43: should we make these two an enum? 👀

src/encoding/tycho_encoder.rs
```diff
@@ -1,20 +1,75 @@
...
+    ///
+    /// # Warning
+    /// This implementation uses default logic to construct the outer calldata (e.g., for setting
+    /// `minAmountOut`). This might not be optimal or safe for production use.
+    ///
+    /// To ensure correctness, **users should implement their own encoding pipeline** using
+    /// [`encode_solutions`].
+    ///
+    /// # Returns
+    /// A vector of fully constructed [`Transaction`]s that can be submitted to a node or bundler.
```
tamaralipows, line 63: Just a note to rename this to `encode_full_calldata`

src/encoding/evm/tycho_encoders.rs
tamaralipows: hmmmmm maybe. I don't have strong feelings tbh

tamaralipows (approved):
Thank you! I think this does the trick!

src/encoding/evm/tycho_encoders.rs
```diff
@@ -45,78 +54,111 @@ impl TychoRouterEncoder {
...
                 swap_encoder_registry.clone(),
-                swapper_pk.clone(),
+                permit2.is_some(),
                 router_address.clone(),
                 token_in_already_in_router,
             )?,
             split_swap_strategy: SplitSwapStrategyEncoder::new(
                 chain,
                 swap_encoder_registry,
-                None,
```
dianacarvalho1, line 66: this was a big bad bug that we never noticed 😬

src/bin/tycho-encode.rs
```diff
@@ -99,11 +100,20 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        "value": format!("0x{}", hex::encode(transactions[0].value.to_bytes_be())),
-        "data": format!("0x{}", hex::encode(&transactions[0].data)),
+        "swaps": format!("0x{}", hex::encode(&encoded_solutions[0].swaps)),
+        "interacting_with": format!("0x{}", hex::encode(&encoded_solutions[0].interacting_with)),
+        "selector": format!("{}",&encoded_solutions[0].selector),
+        "n_tokens": format!("{}", &encoded_solutions[0].n_tokens),
+        "permit": encoded_solutions[0].permit
+            .as_ref()
+            .map(|permit| format!("0x{}", hex::encode(permit.abi_encode())))
+            .unwrap_or_else(String::new),
```
tamaralipows, line 112: uff nice

src/encoding/evm/tycho_encoders.rs
```diff
@@ -39,7 +39,6 @@ pub struct TychoRouterEncoder {
     native_address: Bytes,
     wrapped_address: Bytes,
     router_address: Bytes,
-    #[allow(dead_code)]
     token_in_already_in_router: bool,
```
tamaralipows, line 42: Can we change this to match the solidity naming?

tamaralipows (approved):
Just a minor comment but rest looks good!

src/encoding/evm/tycho_encoders.rs
dianacarvalho1: no

PR #205, opened 2025-05-22
fix: Small misc improvements from audit

dianacarvalho1:
- In RestrictTransferFrom:
  - Compare tokenIn with tokenIn from storage
  - Correct docstrings
  - Recompute storage slots with new names
  - Rename transferFromNeeded to isTransferFromAllowed
  - Don't track amount spent but subtract from amount allowed
- In TychoRouter: Rename transferFromNeeded to isTransferFromAllowed


tamaralipows (approved):
Thank you!

PR #207, opened 2025-05-23
feat: Don't sign permit2 objects

dianacarvalho1:
We don't want to be responsible for holding private keys -> the user is the one that should do this outside of tycho-execution

Done:
- Remove signature from EncodedSolution
- Introduce UserTransferType and pass that everywhere instead of is_permit2_active and token_in_already_in_router
- Remove signing from permit2. Added it to the encoding_utils.rs only
- Mark encode_full_calldata as deprecated
- Backwards compatibility: still accept a signer for the encode_full_calldata case
- Update all tests


src/encoding/evm/encoding_utils.rs
```diff
@@ -9,7 +15,8 @@ use crate::encoding::{
         approvals::permit2::PermitSingle,
         utils::{biguint_to_u256, bytes_to_address},
     },
-    models::{EncodedSolution, NativeAction, Solution, Transaction},
+    models,
+    models::{EncodedSolution, NativeAction, Solution, Transaction, UserTransferType},
 };
 
 /// Encodes the input data for a function call to the given function selector.
```
dianacarvalho1: should we add these functions that the user should not rely on all in one folder and put a huge disclaimer at the top? It feels weird to have them here with actual real encoding utils functions we use in other places

tamaralipows: Yeah - you are so right on this. Is there any way to make the fn not public and thus not accessible by the user?

src/encoding/evm/encoding_utils.rs
```diff
@@ -87,10 +95,12 @@ pub fn encode_input(selector: &str, mut encoded_args: Vec<u8>) -> Vec<u8> {
 /// - Returns `EncodingError::FatalError` if the selector is unsupported or required fields (e.g.,
 ///   permit or signature) are missing.
 pub fn encode_tycho_router_call(
+    chain_id: u64,
     encoded_solution: EncodedSolution,
     solution: &Solution,
-    token_in_already_in_router: bool,
+    user_transfer_type: UserTransferType,
     native_address: Bytes,
+    signer: Option<PrivateKeySigner>,
```
tamaralipows, line 82: Should we then perhaps raise an error if we pass a signer and the `user_transfer_type` is not `permit2`? or does that get too complicated?

src/encoding/models.rs
```diff
@@ -95,15 +121,13 @@ pub struct Transaction {
...
 /// * `permit`: Optional permit for the swap (if permit2 is enabled).
-/// * `signature`: Optional signature for the swap (if permit2 is enabled).
 #[derive(Clone, Debug)]
 pub struct EncodedSolution {
     pub swaps: Vec<u8>,
     pub interacting_with: Bytes,
     pub selector: String,
     pub n_tokens: usize,
     pub permit: Option<PermitSingle>,
-    pub signature: Option<Vec<u8>>,
```
tamaralipows, line 106: Do you think we should still leave the PermitSingle object in here? Couldn't that in theory be manipulated in an attack?

tamaralipows (commented):
Hey hey, I like this change, thanks! Only one "big" concern

src/encoding/models.rs
dianacarvalho1: huuumm the user can validate the PermitSingle object if they want? 😕

src/encoding/evm/encoding_utils.rs
dianacarvalho1: huum do you think that is a big deal?? it wouldn't do anything 🤷🏼‍♀️

src/encoding/evm/encoding_utils.rs
dianacarvalho1: huuummm there should be no? 🤔

src/encoding/models.rs
tamaralipows: Couldn't we have made the same argument then for the min amount? 🤔

dianacarvalho1: huuumm not before no. because we were encoding it already in bytes. there was no easy way to check the min amount out

tamaralipows: Hmmm okay. I'm a bit uncomfortable tbh but I won't insist

tamaralipows (approved):
Thank you!

src/encoding/models.rs
dianacarvalho1: I understand.. but I think here it is a bit different. and we are already putting a lot on the user :/ they can easily check that the amounts/tokens are correct on their side though. in fact, that should be the first step of their final encoding logic (it's still easier than making the object I think)

tamaralipows (approved):
Thank you (for real this time)

PR #211, opened 2025-06-04
feat: Implement BalancerV3Executor

dianacarvalho1:
This was easier than expected 👀 ✨ 

The vault interface is pretty simple and only allows for 1 single swap at a time! 
In contrast to Balancer v2, we are now in charge of the token transfers so we can perform optimizations here 💅🏼

### DIFF
--- a/foundry/src/executors/BalancerV3Executor.sol
+++ b/foundry/src/executors/BalancerV3Executor.sol
@@ -12,42 +12,18 @@ import {
     VaultSwapParams
 } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
 import {RestrictTransferFrom} from "../RestrictTransferFrom.sol";
+import {ICallback} from "../../interfaces/ICallback.sol";
 
 error BalancerV3Executor__InvalidDataLength();
 error BalancerV3Executor__SenderIsNotVault(address sender);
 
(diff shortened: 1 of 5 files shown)


foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -0,0 +1,142 @@
...
+                        tokenIn: tokenIn,
+                        tokenOut: tokenOut,
+                        amountGiven: givenAmount,
+                        transferType: transferType,
+                        receiver: receiver
+                    })
+                )
+            ),
+            (uint256)
+        );
```
tamaralipows: Hmmm, is this a callback that you are encoding here when calling `VAULT.unlock`? In that case, shouldn't the executor inherit from `ICallback`?

tamaralipows (commented):
All looks good, just concerned about the callback interface. Thanks!

foundry/src/executors/BalancerV3Executor.sol
dianacarvalho1: 🥴 you are absolutely right.. I was so focused on balancer specific stuff that I forgot about our own interfaces 🤦🏼‍♀️ so sorryyy.. coming right up!

tamaralipows (approved):
Thank yoooou!

PR #212, opened 2025-06-04
feat: Bebop PMM RFQ 

pedrobergamini:
# Bebop PMM RFQ Protocol Integration

## Description

This PR adds support for Bebop's Private Market Maker Request-for-Quote (PMM RFQ) solution to the Tycho Execution SDK. This is an **execution-only** integration that enables swaps through Bebop's off-chain quote system with on-chain settlement.

#### BebopExecutor.sol
- Implements `IExecutor` interface for Bebop PMM RFQ settlement
- Supports two order types after simplification:
  - **Single**: One-to-one token swaps with a single maker
  - **Aggregate**: Many-to-one swaps with multiple makers (single input → single output)
- Key features:
  - Native ETH handling via `address(0)` convention
  - Partial fill support via `filledTakerAmount` parameter
  - Gas-optimized settlement using single `filledTakerAmount` for all order types
  - Custom error handling for better debugging

#### BebopSwapEncoder (Rust)
- Implements `SwapEncoder` trait for encoding Bebop swap data
- Registered as `rfq:bebop` in the encoder registry


foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,474 @@
...
+import {
+    IERC20,
+    SafeERC20
+} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+/// @dev Bebop settlement interface for PMM RFQ swaps
+interface IBebopSettlement {
+    struct Single {
+        uint256 expiry;
```
dianacarvalho1: can we use camelCase for all these attributes please?

dianacarvalho1: oh ok.. I see this is how it is in their settlement 😬

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,474 @@
...
+            address tokenOut,
+            TransferType transferType,
+            OrderType orderType,
+            bytes memory quoteData,
+            uint8 signatureType,
+            bytes memory signature,
+            bool approvalNeeded
+        ) = _decodeData(data);
+
+        _transfer(address(this), transferType, tokenIn, givenAmount);
```
dianacarvalho1: this transfer is for the in transfer of tokens.. so the receiver should be the bebop settlement contract, not `this`

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,474 @@
...
+        } else {
+            // For ERC20 input, call settlement
+            try IBebopSettlement(bebopSettlement).swapSingle(
+                order, sig, amountIn
+            ) {
+                // Success, calculate amountOut from balance difference
+            } catch {
+                revert BebopExecutor__SettlementFailed();
+            }
+        }
```
dianacarvalho1: I think this could be collapsed into something like:
```

        uint256 value = tokenIn == address(0) ? amountIn : 0;
        try IBebopSettlement(bebopSettlement).swapSingle{value: value}(
            order, sig, amountIn
        ) {
        } catch {
            revert BebopExecutor__SettlementFailed();
        }

```

dianacarvalho1: Can you explain why you want to catch the error though? The revert errors from the Bebop settlement are not nice?

foundry/test/TychoRouterSequentialSwap.t.sol
```diff
@@ -494,4 +494,27 @@ contract TychoRouterSequentialSwapTest is TychoRouterTestSetup {
...
     }
+
+    function testBebopUSV2Integration() public {
+        // Performs a sequential swap from USDC to WETH though Bebop and USV2 pools
+        //
+        //   USDC ──(bebop)──> WETH ───(USV2)──> USDC
+        deal(USDC_ADDR, ALICE, 1000 * 10 ** 6);
+        uint256 balanceBefore = IERC20(USDC_ADDR).balanceOf(ALICE);
+
+        // Approve permit2
```
dianacarvalho1: you are not approving permit2 here

foundry/test/TychoRouterSequentialSwap.t.sol
```diff
@@ -494,4 +494,27 @@ contract TychoRouterSequentialSwapTest is TychoRouterTestSetup {
...
+        IERC20(USDC_ADDR).approve(tychoRouterAddr, type(uint256).max);
+        bytes memory callData = loadCallDataFromFile("test_encode_bebop_single");
+        (bool success,) = tychoRouterAddr.call(callData);
+
+        vm.stopPrank();
+
+        uint256 balanceAfter = IERC20(USDC_ADDR).balanceOf(ALICE);
+
+        assertTrue(success, "Call Failed");
+        // Expected amount would need to be determined from actual integration test
```
dianacarvalho1: why? 🤔

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -121,20 +123,22 @@ contract TychoRouterTestSetup is Constants, Permit2TestHelper, TestUtils {
...
+        address[] memory executors = new address[](9);
         executors[0] = address(usv2Executor);
         executors[1] = address(usv3Executor);
         executors[2] = address(pancakev3Executor);
         executors[3] = address(usv4Executor);
         executors[4] = address(balancerv2Executor);
-        executors[5] = address(ekuboExecutor);
-        executors[6] = address(curveExecutor);
-        executors[7] = address(maverickv2Executor);
+        executors[5] = address(bebopExecutor);
```
dianacarvalho1: can bebop be the 8th please? this changes the addresses of the other executors. I think that this is why some of the tests are failing

foundry/test/executors/BebopExecutor.t.sol
```diff
@@ -0,0 +1,695 @@
...
+            bytes memory signature,
+            bool approvalNeeded
+        )
+    {
+        return _decodeData(data);
+    }
+}
+
+/// @notice Mock Bebop settlement contract for testing
+contract MockBebopSettlement is Test, Constants {
```
dianacarvalho1: I understand you decided to mock this because of the signatures and etc.. but I would prefer if we tested with the real settlement contract. Could we try to replicate a bebop swap that actually happened in a certain block?

src/encoding/evm/swap_encoder/builder.rs
```diff
@@ -81,6 +81,9 @@ impl SwapEncoderBuilder {
                 self.chain,
                 self.config,
             )?)),
+            "vm:bebop" => {
```
dianacarvalho1: hum this is not a vm protocol.. maybe we can do `rfq:bebop` instead

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -569,6 +569,123 @@ impl SwapEncoder for MaverickV2SwapEncoder {
...
+            )?;
+        } else {
+            approval_needed = true;
+        }
+
+        // Validate component ID
+        Self::validate_component_id(&swap.component.id)?;
+
+        // Get quote data and signature from static attributes
+        let quote_data = get_static_attribute(&swap, "quote_data")?;
```
dianacarvalho1: uff this feels very wrong to me 😬 but I do realise that there isn't any other way of passing this data to encoding. But here we are basically using the static attributes of the ProtocolComponent for bebop specific data. and this is not static at all

dianacarvalho1: I realise that we possibly need to extend our interfaces! I will discuss this with the team and get back to you!

dianacarvalho1 (commented):
Thank you so much for this @pedrobergamini 🙏🏼 It looks great so far! I have a few suggestions and concerns.
We should know from the quote data how much the amount out actually is. maybe we can double check that the calculated amount out (from the balances) matches the one from the quote?

src/encoding/evm/swap_encoder/builder.rs
dianacarvalho1: also I think we need to add this to the `IN_TRANSFER_REQUIRED_PROTOCOLS` no?

src/encoding/evm/constants.rs
```diff
@@ -48,5 +49,6 @@ pub static CALLBACK_CONSTRAINED_PROTOCOLS: LazyLock<HashSet<&'static str>> = Laz
     set.insert("uniswap_v4");
     set.insert("ekubo_v2");
     set.insert("vm:balancer_v3");
+    set.insert("rfq:bebop");
```
dianacarvalho1: it is not really callback constrained is it? 🤔 But I see the issue.. 
If we wanted to have something like in your test:
univ3 -> bebop
the receiver of the first swap would have to be the settlement contract and not the "pool" and right now we don't support that 😕

dianacarvalho1 (commented):
Thank you! 🙏🏼 This is looking veeery gooood! We are almost there! 
I still have a few outstanding comments from last review though. The one that worries me the most is the one related to testing. It concerns me that we are mocking the bebop settlement every time we test.
Another note, I mentioned on telegram already that we need to be careful with the givenAmount. The givenAmount might not be exactly the amount quoted. I think we need to pass the quoted amount in in the data as well, so that in the executor we can check:
1. If the givenAmount is higher, we can only pass the quoted amount
2. If the givenAmount is lower, it's a partial order (not sure if anything needs to change there)
Can you add tests for these cases as well please?

Ah also make sure to do the [slither](https://docs.propellerheads.xyz/tycho/for-dexs/protocol-integration/contributing-guidelines#contract-analysis) analysis and address the suggestions/silence the warnings.

src/encoding/evm/constants.rs
pedrobergamini: Exactly, which approach would you suggest as we have this constraint?

dianacarvalho1: yeah 😕 for now, can you just leave a comment there explaining this please?

dianacarvalho1 (approved):

greendianasaur (approved):
Thank you so much for this @pedrobergamini  🙏 
Awesome work here ✨ We are ready to deploy and test this 🚀

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,267 @@
...
+        external
+        payable
+        virtual
+        override
+        returns (uint256 calculatedAmount)
+    {
+        calculatedAmount = _swap(givenAmount, data);
+    }
+
+    function _swap(uint256 givenAmount, bytes calldata data)
```
dianacarvalho1: no need for these two functions here I think

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,267 @@
...
+    /// @param givenAmount The amount of tokens available from the router for this swap
+    /// @param filledTakerAmount The requested fill amount (guaranteed to be non-zero by encoder)
+    /// @return actualFilledTakerAmount The amount that will actually be filled
+    function _getActualFilledTakerAmount(
+        uint256 givenAmount,
+        uint256 filledTakerAmount
+    ) internal pure returns (uint256 actualFilledTakerAmount) {
+        actualFilledTakerAmount =
+            filledTakerAmount > givenAmount ? givenAmount : filledTakerAmount;
+    }
```
dianacarvalho1: I think we don't need a full method for a one liner

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,267 @@
...
+        bytes memory bebopCalldata,
+        uint256 givenAmount,
+        uint256 originalFilledTakerAmount
+    ) public pure returns (bytes memory) {
+        bytes4 selector = _getSelector(bebopCalldata);
+
+        // The position of filledTakerAmount differs between swapSingle and swapAggregate
+        // due to how Solidity encodes structs:
+        // - swapSingle: Single struct is encoded inline (no offset), so filledTakerAmount is at position 388
+        // - swapAggregate: Aggregate struct uses offset (has arrays), so filledTakerAmount is at position 68
```
dianacarvalho1: is this always true?? 🤔 I think they return this information inside of the Bebop quote in `partialFillOffset`

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,267 @@
...
+
+        // Approve Bebop settlement to spend tokens if needed
+        if (approvalNeeded && tokenIn != address(0)) {
+            // slither-disable-next-line unused-return
+            IERC20(tokenIn).forceApprove(bebopSettlement, type(uint256).max);
+        }
+
+        // Bebop orders specify the receiver, so we need to check the receiver's balance
+        // We'll use the executor's balance since Bebop should send tokens here for the router to collect
+        uint256 balanceBefore = _balanceOf(tokenOut, address(this));
```
dianacarvalho1: this is not necessarily true. If this is a single swap for example, the receiver might be the user already. I think we need to pass the receiver's value in the executor calldata 🤔

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -631,6 +631,255 @@ impl SwapEncoder for BalancerV3SwapEncoder {
...
+            if filled_amount != U256::ZERO {
+                filled_amount
+            } else {
+                // Extract taker_amount from the order based on the function selector
+                let selector = &bebop_calldata[0..4];
+
+                // swapSingle selector: 0x4dcebcba
+                // swapAggregate selector: 0xa2f74893
+                const SWAP_SINGLE_SELECTOR: [u8; 4] = [0x4d, 0xce, 0xbc, 0xba];
+                const SWAP_AGGREGATE_SELECTOR: [u8; 4] = [0xa2, 0xf7, 0x48, 0x93];
```
dianacarvalho1: Can you explain why this is necessary? I think they expose this value in `partialFillOffset` already..

dianacarvalho1 (commented):
Thank you for these changes @pedrobergamini 🙏🏼 It looks much simpler already! 
I reviewed only the executor and the encoder now though (not the tests) and I have one big question

foundry/src/executors/BebopExecutor.sol
pedrobergamini: Agreed

foundry/src/executors/BebopExecutor.sol
pedrobergamini: I may need the internal `_swap` function for the `BebopExecutionHarness`'s swap function override, as I need to call `super` and avoid rewriting the entire function. I'm getting back to the tests now and will re-unify the logic in case requirements change

foundry/src/executors/BebopExecutor.sol
pedrobergamini: Will refactor to use `partialFillOffset` instead so the implementation is cleaner, good reminder, thanks!

foundry/src/executors/BebopExecutor.sol
pedrobergamini: Yep agreed

src/encoding/evm/swap_encoder/swap_encoders.rs
pedrobergamini: Agreed, we should definitely use the offset from the quote response instead. I'll update the encoding pipleine so `partialFillOffset` is passed to the executor. Will also simplify significantly

foundry/src/executors/BebopExecutor.sol
pedrobergamini: This is probably the reason why tests are breaking 😓 thanks!

src/encoding/evm/swap_encoder/swap_encoders.rs
pedrobergamini: Actually, I think I misunderstood your question earlier. If you're asking why we need to perform those checks, it's because when we're handling a swapAggregate case where filledTakerAmount == 0, we need to iterate over each order.taker_amount to determine the total filled taker amount that needs to be passed to the executor.

Basically, the idea is that we never pass 0 to the executor. While Bebop's settlement contracts interpret 0 as a full fill, passing 0 would require the executor contract to compare the givenAmount with the order's original total filled taker amount to determine the actual amount to be filled, resulting in significant extra gas that is avoidable, makes sense?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -634,6 +634,303 @@ impl SwapEncoder for BalancerV3SwapEncoder {
...
+            )));
+        }
+
+        // Extract the original filledTakerAmount from the order and use the context receiver
+        let (original_filled_taker_amount, receiver) = {
+            let filled_taker_amount = U256::from_be_slice(
+                &bebop_calldata[filled_taker_amount_pos..filled_taker_amount_pos + 32],
+            );
+
+            // Extract receiver and taker_amount from the order based on the function selector
```
dianacarvalho1: why don't we pass this in the user data instead of doing this very complex logic? 😬

pedrobergamini: I updated to get receiver from the user data, but I think we still need to extract the total taker_amount here rather than relying on the user data due to our constraint of not passing `filled_taker_amount` as 0 to the executor

dianacarvalho1: why? the bebop quote response has that value already outside of the calldata for us I think 🤔

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,278 @@
...
+import "../RestrictTransferFrom.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+import {
+    IERC20,
+    SafeERC20
+} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+/// @dev Bebop settlement interface for PMM RFQ swaps
+interface IBebopSettlement {
```
dianacarvalho1: I think this is only needed in the Harness now, can we move it there please?

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,278 @@
...
+    /// @notice Function selectors for Bebop settlement methods
+    bytes4 public constant SWAP_SINGLE_SELECTOR = 0x4dcebcba;
+    bytes4 public constant SWAP_AGGREGATE_SELECTOR = 0xa2f74893;
+
+    /// @notice Bebop-specific errors
+    error BebopExecutor__InvalidDataLength();
+    error BebopExecutor__InvalidInput();
+    error BebopExecutor__InvalidSignatureLength();
+    error BebopExecutor__InvalidSignatureType();
+    error BebopExecutor__ZeroAddress();
```
dianacarvalho1: some of these errors are not used anymore

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,278 @@
...
+        );
+
+        // For ERC20 inputs, don't move tokens here. Bebop settlement pulls from the taker directly.
+        // The router/harness will ensure the taker has the funds and has granted allowance to settlement.
+        if (tokenIn != address(0)) {
+            // no-op
+        }
+        // For ETH (tokenIn == address(0)), don't transfer
+        // The harness gives ETH to the taker, and Bebop settlement expects it to stay there
+
```
dianacarvalho1: ?

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,278 @@
...
+        // Approve Bebop settlement to spend tokens if needed
+        if (approvalNeeded && tokenIn != address(0)) {
+            // slither-disable-next-line unused-return
+            IERC20(tokenIn).forceApprove(bebopSettlement, type(uint256).max);
+        }
+
+        // Check the receiver's balance before the swap
+        uint256 balanceBefore = _balanceOf(tokenOut, receiver);
+
+        // Execute the swap with the forwarded calldata
```
dianacarvalho1: the comment is misplaced

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,278 @@
...
+        tokenOut = address(bytes20(data[20:40]));
+        transferType = TransferType(uint8(data[40]));
+
+        // Get bebop calldata length and validate
+        uint32 bebopCalldataLength = uint32(bytes4(data[41:45]));
+        if (data.length != 99 + bebopCalldataLength) {
+            revert BebopExecutor__InvalidDataLength();
+        }
+
+        // Extract bebop calldata
```
dianacarvalho1: these comments are a bit redundant imo 😅

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,278 @@
...
+            // Calculate the actual position and store the new value
+            let actualPos := add(dataPtr, filledTakerAmountPos)
+            mstore(actualPos, newFilledTakerAmount)
+        }
+
+        return bebopCalldata;
+    }
+
+    /// @dev Helper function to extract selector from bytes
+    function _getSelector(bytes memory data) internal pure returns (bytes4) {
```
dianacarvalho1: this is needed in the harness as well I think

foundry/test/TychoRouterProtocolIntegration.t.sol
```diff
@@ -26,4 +27,26 @@ contract TychoRouterTestProtocolIntegration is TychoRouterTestSetup {
...
+        bytes memory callData =
+            loadCallDataFromFile("test_single_encoding_strategy_usv4_eth_in");
+        (bool success,) = tychoRouterAddr.call{value: 1 ether}(callData);
+
+        vm.stopPrank();
+
+        uint256 balanceAfter = IERC20(PEPE_ADDR).balanceOf(ALICE);
+
+        assertTrue(success, "Call Failed");
+        assertEq(balanceAfter - balanceBefore, 235610487387677804636755778);
```
dianacarvalho1: oh can you delete this please? I think I messed up in the merge with main 😅

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -188,6 +193,25 @@ contract TychoRouterTestSetup is Constants, Permit2TestHelper, TestUtils {
         );
     }
 
+    function encodeBebopSwap(
```
dianacarvalho1: I think this is not used anywhere now

foundry/test/protocols/Bebop.t.sol
```diff
@@ -0,0 +1,1195 @@
...
+        uint256 originalAmountIn = 200000000; // 200 USDC
+
+        // Create the executor params
+        bytes memory params = abi.encodePacked(
+            USDC_ADDR,
+            ONDO_ADDR,
+            uint8(RestrictTransferFrom.TransferType.Transfer),
+            uint32(bebopCalldata.length),
+            bebopCalldata,
+            uint8(12), // partialFillOffset for swapSingle (388 = 4 + 12*32)
```
dianacarvalho1: what does the comment mean? how does that relate to 12? 4 is the selector and then 12 entries until the value we are interested in?

foundry/test/protocols/Bebop.t.sol
```diff
@@ -0,0 +1,1195 @@
...
+            originalAmountIn,
+            uint8(1), // approvalNeeded: true
```
dianacarvalho1: I think this can be just `true`

foundry/test/protocols/Bebop.t.sol
```diff
@@ -0,0 +1,1195 @@
...
+            uint8(RestrictTransferFrom.TransferType.Transfer)
+        );
+        // Missing rest of the data
+
+        vm.expectRevert(BebopExecutor.BebopExecutor__InvalidDataLength.selector);
+        bebopExecutor.decodeParams(tooShortParams);
+    }
+
+    // Integration tests
+    function testSwapSingleIntegration() public {
```
dianacarvalho1: what is the difference between this test and `testSingleOrder`? They seem pretty similar..

foundry/test/protocols/Bebop.t.sol
```diff
@@ -0,0 +1,1195 @@
...
+
+        // Should be 250e6 (the givenAmount, since it's less than both originalFilledTakerAmount and originalAmountIn)
+        assertEq(
+            newFilledTakerAmount,
+            250e6,
+            "Modified filledTakerAmount should match givenAmount"
+        );
+    }
+
+    function testModifyFilledTakerAmount_AggregateOrder() public {
```
dianacarvalho1: this test and `testModifyFilledTakerAmount_SingleOrder` are testing the same now no? 🤔

foundry/test/protocols/BebopExecutionHarness.t.sol
```diff
@@ -0,0 +1,252 @@
...
+
+        // Trust the encoder-provided receiver when present; if it's zero, fall back to
+        // decoding the taker from the Bebop order so we still impersonate correctly
+        bytes4 sel = _getSelector(bebopCalldata);
+        console.log("[BebopHarness] selector computed");
+        console.logBytes4(sel);
+        console.log("[BebopHarness] bebopCalldata len=%s", bebopCalldata.length);
+        address takerAddress = receiver;
+        address outputReceiver = receiver;
+        if (takerAddress == address(0)) {
```
dianacarvalho1: Do we really need this if else? why? 🤔 the take should always be defined

foundry/test/protocols/BebopExecutionHarness.t.sol
```diff
@@ -0,0 +1,252 @@
...
+                outputReceiver = order.receiver;
+            }
+        }
+        console.log("[BebopHarness] taker=%s", takerAddress);
+
+        // Make sure taker has the input assets and approvals when needed
+        // If the encoder gave us a zero original amount, pull it from the calldata so we can
+        // still set the correct fill
+        uint256 effectiveOriginal = originalFilledTakerAmount;
+        if (effectiveOriginal == 0) {
```
dianacarvalho1: why do we need this as well?

foundry/test/protocols/BebopExecutionHarness.t.sol
```diff
@@ -0,0 +1,252 @@
...
+
+        uint256 afterBal = _balanceOf(tokenOut, outputReceiver);
+        calculatedAmount = afterBal - beforeBal;
+        console.log(
+            "[BebopHarness] afterBal=%s calculatedAmount=%s",
+            afterBal,
+            calculatedAmount
+        );
+
+        // no-op; keep function end balanced
```
dianacarvalho1: lost comment

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -634,6 +634,303 @@ impl SwapEncoder for BalancerV3SwapEncoder {
...
+        // Encode packed data for the executor
+        // Format: token_in | token_out | transfer_type | bebop_calldata_length |
+        //         bebop_calldata | partial_fill_offset | original_filled_taker_amount |
+        //         approval_needed | receiver
+        let args = (
+            token_in,
+            token_out,
+            (encoding_context.transfer_type as u8).to_be_bytes(),
+            (bebop_calldata.len() as u32).to_be_bytes(),
+            &bebop_calldata[..],
```
dianacarvalho1: can we simplify this? We can put the bebop calldata at the end of everything. Then in the executor we only need to decode it like `bebopCalldata = data[95:];`

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1697,4 +1994,356 @@ mod tests {
...
+            )
+            .unwrap();
+
+            let encoded_swap = encoder
+                .encode_swap(&swap, &encoding_context)
+                .unwrap();
+            let hex_swap = encode(&encoded_swap);
+
+            // Verify the encoding contains the expected tokens
+            assert!(hex_swap.contains("a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48")); // USDC
```
dianacarvalho1: can we verify that the position of the encoded data is also correct please?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1697,4 +1994,356 @@ mod tests {
...
+            // bebop_calldata | partialFillOffset | original_filled_taker_amount | approval_needed |
+            // receiver Looking at the hex output, we can see that partialFillOffset
+            // (0c) is followed by the original filledTakerAmount
+            assert!(
+                hex_swap
+                    .contains("0c000000000000000000000000000000000000000000000000000000000bebc200"),
+                "partialFillOffset byte (0c) should be followed by original filledTakerAmount"
+            );
+
+            write_calldata_to_file("test_encode_bebop_single", hex_swap.as_str());
```
dianacarvalho1: I think this is not used anywhere in the executor tests.. maybe we can use it for a `testDecodeIntegration`?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1697,4 +1994,356 @@ mod tests {
...
+            )
+            .unwrap();
+
+            let encoded_swap = encoder
+                .encode_swap(&swap, &encoding_context)
+                .unwrap();
+            let hex_swap = encode(&encoded_swap);
+
+            // Verify the encoding contains the expected tokens
+            assert!(hex_swap.contains("c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2")); // WETH
```
dianacarvalho1: same here

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1697,4 +1994,356 @@ mod tests {
...
+            );
+
+            // Verify the partialFillOffset byte (02 = 2) appears in the right place
+            let expected_pattern = format!("02{}", filled_amount_hex);
+            assert!(
+                hex_swap.contains(&expected_pattern),
+                "partialFillOffset byte (02) should be followed by original filledTakerAmount"
+            );
+
+            write_calldata_to_file("test_encode_bebop_aggregate", hex_swap.as_str());
```
dianacarvalho1: and here I think we can just remove this line (don't forget to delete the calldata from the file as well)

dianacarvalho1 (commented):
Thank you so much for this @pedrobergamini 🙏🏼 🙏🏼 We are at the finish line now 🤞🏼 🤞🏼 almost there

foundry/test/protocols/Bebop.t.sol
dianacarvalho1: @pedrobergamini can you answer this please? I don't see a big difference and I would like to remove them

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1697,4 +2045,363 @@ mod tests {
             write_calldata_to_file("test_encode_maverick_v2", hex_swap.as_str());
         }
     }
+
+    mod bebop {
+        use super::*;
+        use crate::encoding::{evm::utils::write_calldata_to_file, models::TransferType};
+
+        #[test]
+        fn test_encode_bebop_single() {
```
dianacarvalho1, line 1729: TODO for me: simplify this test

foundry/src/executors/BebopExecutor.sol
```diff
@@ -0,0 +1,182 @@
...
+        // If the new filledTakerAmount is the same as the original, return the original calldata
+        if (newFilledTakerAmount == originalFilledTakerAmount) {
+            return bebopCalldata;
+        }
+
+        // Use assembly to modify the filledTakerAmount at the correct position
+        // slither-disable-next-line assembly
+        assembly {
+            // Get pointer to the data portion of the bytes array
+            let dataPtr := add(bebopCalldata, 0x20)
```
tamaralipows, line 149: How does this work?

foundry/test/protocols/Bebop.t.sol
```diff
@@ -0,0 +1,474 @@
...
+    IERC20 WBTC = IERC20(WBTC_ADDR);
+    IERC20 ONDO = IERC20(ONDO_ADDR);
+    IERC20 USDT = IERC20(USDT_ADDR);
+
+    function setUp() public {
+        // Fork will be created in individual tests to allow different fork blocks
+    }
+
+    function testDecodeData() public {
+        vm.createSelectFork(vm.rpcUrl("mainnet"), 22667985);
```
tamaralipows, line 49: vm.rollFork doesn't work?

foundry/test/protocols/Bebop.t.sol
```diff
@@ -0,0 +1,474 @@
...
+    IERC20 WETH = IERC20(WETH_ADDR);
+    IERC20 USDC = IERC20(USDC_ADDR);
+    IERC20 DAI = IERC20(DAI_ADDR);
+    IERC20 WBTC = IERC20(WBTC_ADDR);
+    IERC20 ONDO = IERC20(ONDO_ADDR);
+    IERC20 USDT = IERC20(USDT_ADDR);
+
+    function setUp() public {
+        // Fork will be created in individual tests to allow different fork blocks
+    }
```
tamaralipows: We can remove the setup altogether then no?

tamaralipows (approved):
Thank you! Not much to say here, it looks great! Minor questions, nothing blocking. Great job!

foundry/src/executors/BebopExecutor.sol
pedrobergamini: This line sets dataPtr to the start of the bebopCalldata array in memory (skips first 32 bytes that store the array's length), then we calculate the actual position in the array of the filledTakerAmount value to be overriden by newFilledTakerAmount

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -634,6 +634,117 @@ impl SwapEncoder for BalancerV3SwapEncoder {
...
+        })?;
+
+        if user_data.len() < 37 {
+            return Err(EncodingError::InvalidInput(
+                "User data too short to contain offset and Bebop calldata".to_string(),
+            ));
+        }
+
+        let partial_fill_offset = user_data[0];
+        let original_filled_taker_amount = U256::from_be_slice(&user_data[1..33]);
```
pedrobergamini, line 720: Quick question, does the other part of the tycho rfq system inject this value or check if it's correct? My only concern is that if the user passes an incorrect value here it could lead to issues

pedrobergamini (commented):
Looks good to me! Thanks @dianacarvalho1, just added a quick question.

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: yeah I understand your concern. So we will have two modes of running this:
1. The one that you were doing: just using encoding and the user is responsible for passing this themselves
2. Use Tycho as a whole and then we will request the quote here and pass these values automatically (this is coming in my next PR)

foundry/test/protocols/Bebop.t.sol
dianacarvalho1: that is only when you already created the fork I think. Since we are not doing that at setup, rollFork doesn't work

PR #214, opened 2025-06-05
feat: Add BalancerV3Encoder (+ integration tests)

dianacarvalho1:


foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -82,6 +82,24 @@ contract BalancerV3Executor is IExecutor, RestrictTransferFrom, ICallback {
         return abi.encode(amountCalculated);
     }
 
+    function handleCallback(bytes calldata data)
+        external
+        returns (bytes memory result)
+    {
+        verifyCallback(data);
+        bytes memory result = _swapCallback(data[68:]);
```
dianacarvalho1: this is to remove the selector + dataOffset + dataLength

foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -82,6 +82,24 @@ contract BalancerV3Executor is IExecutor, RestrictTransferFrom, ICallback {
...
+        // However, the Vault expects the result to be ABI encoded. That is why we need to encode it here again.
+        return abi.encode(result);
+    }
+
+    function swapCallback(bytes calldata data)
+        external
+        returns (bytes memory result)
+    {
+        return _swapCallback(data);
+    }
```
dianacarvalho1, line 105: This is the exposed method if we want to swap directly with the executor (no router)

foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -94,7 +112,7 @@ contract BalancerV3Executor is IExecutor, RestrictTransferFrom, ICallback {
             address receiver
         )
     {
-        if (data.length != 113) {
+        if (data.length < 113) {
```
dianacarvalho1: when we do the whole router option, we might get padding at the end, so it won't be exactly 113. I thought it was better to just check < than to remove the padding manually

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1506,59 +1551,120 @@ mod tests {
...
+                    "01",
+                    // receiver
+                    "1d96f2f6bef1202e4ce1ff6dad0c2cb002861d3e",
+                ))
+            );
+            write_calldata_to_file("test_encode_balancer_v3", hex_swap.as_str());
+        }
+    }
+
+    mod maverick_v2 {
```
dianacarvalho1, line 1611: it was outside of a mod. I didn't like it 😅

dianacarvalho1:
All three swap balancer v3 tests (solidity test, executor integration and router integration) use different pools and different tokens! This was just to test a couple of different cases..

foundry/src/executors/BalancerV3Executor.sol
tamaralipows: Confused - why do we not get this problem for other protocols?

foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -82,6 +82,24 @@ contract BalancerV3Executor is IExecutor, RestrictTransferFrom, ICallback {
         return abi.encode(amountCalculated);
     }
 
+    function handleCallback(bytes calldata data)
+        external
+        returns (bytes memory result)
+    {
+        verifyCallback(data);
+        result = _swapCallback(data[68:]);
```
tamaralipows: Let's add a comment here? About what we are skipping

foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -82,6 +82,24 @@ contract BalancerV3Executor is IExecutor, RestrictTransferFrom, ICallback {
...
+        // Our general callback logic returns a not ABI encoded result.
+        // However, the Vault expects the result to be ABI encoded. That is why we need to encode it here again.
+        return abi.encode(result);
```
tamaralipows, line 97: I'm confused about what you mean about the vault expecting an ABI encoded result, what does this mean? Which error would you get otherwise? This just encodes it as a dynamic bytes array?

tamaralipows (commented):
All looks good but I think I'm missing some context here about the abi encoding situation, maybe let's have a call?

foundry/src/executors/BalancerV3Executor.sol
dianacarvalho1: this is to remove the selector + dataOffset + dataLength

foundry/src/executors/BalancerV3Executor.sol
dianacarvalho1: because we are slicing nicely.. ooookk I can slice nicely xD

foundry/src/executors/BalancerV3Executor.sol
dianacarvalho1: Ok let me explain.. The Vault expects a return value that is ABI encoded. In our callback logic, in the [dispatcher](https://github.com/propeller-heads/tycho-execution/blob/main/foundry/src/Dispatcher.sol#L129), we decided that the result would **not** be abi encoded to save on gas so we are decoding it there. That's why we need to encode twice here 😬

foundry/src/executors/BalancerV3Executor.sol
tamaralipows: Yes I know - I'm asking to add a comment haha

foundry/src/executors/BalancerV3Executor.sol
tamaralipows: Instead of doing this check here, we could check that the data is one of two lengths at a higher level? One length for direct through execution, and one length through router? Not sure, what do you think would be safest? To me it doesnt make sense to slice and then check length after we explicitly slice, because how could the length even change between these two points

dianacarvalho1: yes yes I'm adding the explicit slice now!

tamaralipows: But does that mean you are removing the length check?

dianacarvalho1: no no, see this commit [fix(BalancerV3Executor): Slice callback data nicely](https://github.com/propeller-heads/tycho-execution/pull/214/commits/941879d5d6657c1a3b5e7f77b2b8a4ab76e2f574)

tamaralipows: This was what this comment was about lol

> To me it doesnt make sense to slice and then check length after we explicitly slice, because how could the length even change between these two points

Are we doing this in other executors? Thats so werid

dianacarvalho1: yes we are xD

dianacarvalho1: or no?? no actually. because I repurposed this from the overall decode data method.. you are right.. I will do this at a higher level only once!

tamaralipows: Thank you!

tamaralipows (approved):
Very nice, thank you!

PR #223, opened 2025-06-23
test: Refactor tests

greendianasaur:
- Move encoding integration tests to integration test folder (in the rust project)
  - Removed them all from src/encoding/evm/tycho_encoders.rs
  - To move them out of the crate I had to use the encoder builder instead of the tycho encoder directly. This is actually a nice change because it is how our users use us anyway. Everything else in the tests is copy pasted.
- Move protocol integration tests be inside the protocol file. This way we can change the block without any problems and it is easier for integrators


tests/common/encoding.rs
```diff
@@ -0,0 +1,278 @@
...
+/// - `native_address`: The address used to represent the native token
+///
+/// # Returns
+/// A `Result<Transaction, EncodingError>` that either contains the full transaction data (to,
+/// value, data), or an error if the inputs are invalid.
+///
+/// # Errors
+/// - Returns `EncodingError::FatalError` if the function signature is unsupported or required
+///   fields (e.g., permit or signature) are missing.
+pub fn encode_tycho_router_call(
```
greendianasaur, line 73: I copied this outside of the crate because:
1. The one in encoding_utils is not public outside of the crate
2. Once we deprecate the one in utils, we can delete that function there and keep this one literally only for testing. This makes it even more evident that this should be implemented by the user

louise-poole (approved):

greendianasaur:
not sure what happened with CI here, made a new PR with the same commits here https://github.com/propeller-heads/tycho-execution/pull/225

PR #224, opened 2025-06-23
fix: Remove unnecessary clones

greendianasaur:
Bonus: Exclude foundry files from the rust crate (this should improve the build time)


louise-poole (approved):
❤️

PR #232, opened 2025-06-30
feat: Plan for holding tokens in the router

greendianasaur:
The goal is: to make the Router safe to hold tokens. 

For this I introduced two new transient storage values that hold uint256:
- `transferUnlocked`: the amount for this token that is unlocked
    - this is necessary because of split swaps. We might need to unlock/lock the same token multiple times, so we need to keep track of the amounts.
- `preSwapTokenBalance`: the balance of the router for this token before the swap
- `_UNLOCKED_TOKENS`: the number of tokens that are currently unlocked.

The slots for each are computed using the token address.

We have multiple scenarios where transfers happen:
- Token in transfer: the user's token in
    - normal allowances/permit2: this is under the router control.
        - before transferring we check the router balance for this token
        - after transferring and if the receiver was the router (it might have been a pool directly) => we unlock the router for this token and this amount. Unlocking means adding the amount to the current `transferUnlocked` variable, setting `preSwapTokenBalance` to the balance (if it's 0 previously) and adding 1 to the current _UNLOCKED_TOKENS.
    - user transfers tokens before calling the swap method
        - for this case, the user will have to call `unlock` on the Router themselves before transferring the funds. This way we know the router's balance prior to the transfer
- In between swaps token transfers: Sometimes between swaps (or in the first swap if it's a protocol that doesn't support optimizations) we also perform token transfers from the router funds. This happens in two scenarios:
    - At the end of the current swap, if the receiver is the router, then we need to unlock the router for token out so it can be used in the next swap (see Uniswap v2 Executor for example)
    -  When we call `_transfer` with `TransferType.Transfer`
        - In this case,  we need to call `_is_unlocked` before performing the actual transfer to make sure we won't use funds we are not supposed to. After the transfer we need to call `_maybe_lock`. `_maybe_lock` decreases the `transferUnlocked` with the current amount and if the unlocked amount is 0 then it will completely lock this token by setting preSwapTokenBalance to 0 and decrease one from _UNLOCKED_TOKENS.
        - For cases like curve and balancer v2, where the pool/vault takes the tokens directly from the router itself, we need to perform the `_is_unlocked` and `_maybe_lock` checks manually (see Curve Executor for example)

### DIFF
--- a/foundry/src/BetterTychoRouter.sol
+++ b/foundry/src/BetterTychoRouter.sol
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.26;
+
+import "../lib/IWETH.sol";
+import "../lib/bytes/LibPrefixLengthEncodedByteArray.sol";
+
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import "@openzeppelin/contracts/utils/Pausable.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@permit2/src/interfaces/IAllowanceTransfer.sol";
+import "./Dispatcher.sol";
+import {LibSwap} from "../lib/LibSwap.sol";
+import {RestrictTransferFrom} from "./RestrictTransferFrom.sol";
+
+error TychoRouter__AddressZero();
+error TychoRouter__EmptySwaps();
+error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
+error TychoRouter__AmountOutNotFullyReceived(
+    uint256 amountIn, uint256 amountConsumed
+);
+error TychoRouter__MessageValueMismatch(uint256 value, uint256 amount);
+error TychoRouter__InvalidDataLength();
+error TychoRouter__UndefinedMinAmountOut();
+
+contract TychoRouter is
+    AccessControl,
+    Dispatcher,
+    Pausable,
+    ReentrancyGuard,
+    RestrictTransferFrom
+{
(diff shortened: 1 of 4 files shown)


foundry/src/BetterTychoRouter.sol
```diff
@@ -0,0 +1,103 @@
...
+error TychoRouter__EmptySwaps();
+error TychoRouter__NegativeSlippage(uint256 amount, uint256 minAmount);
+error TychoRouter__AmountOutNotFullyReceived(
+    uint256 amountIn, uint256 amountConsumed
+);
+error TychoRouter__MessageValueMismatch(uint256 value, uint256 amount);
+error TychoRouter__InvalidDataLength();
+error TychoRouter__UndefinedMinAmountOut();
+
+contract TychoRouter is
```
greendianasaur, line 28: very simplified TychoRouter just for the purpose of this plan

foundry/src/RestrictTransferFrom.sol
```diff
@@ -115,23 +142,91 @@ contract RestrictTransferFrom {
...
             if (isPermit2) {
                 // Permit2.permit is already called from the TychoRouter
                 permit2.transferFrom(sender, receiver, uint160(amount), tokenIn);
             } else {
                 // slither-disable-next-line arbitrary-send-erc20
                 IERC20(tokenIn).safeTransferFrom(sender, receiver, amount);
             }
+            if (receiver == address(this)) {
+                _unlock(tokenIn, routerBalance);
+            }
```
kayibal: At what point does the router then become locked again here?

greendianasaur: in the `_maybe_lock` method

tamaralipows:
Quoted from slack:

> I’ve gone through the audit and the PR and I can’t really imagine any deep reason why we chose to do it that way, other than the fact that we weren’t planning to hold tokens per user. I think this was designed with the idea that only we could withdraw the funds, and so every single swap would need to be careful not to spend funds that didn’t belong to them, without any per-user system, so we were quite limited in that regard. For that reason I think it would be a lot easier to start fresh here.

PR #236, opened 2025-07-08
feat: Upgrade scripts to submit to Safe wallet

dianacarvalho1:
Also add revoke-role.js script


foundry/scripts/set-roles.js
```diff
@@ -27,13 +33,21 @@ async function main() {
     };
 
     // Iterate through roles and grant them to the corresponding addresses
+    let nonceOffset = 0
```
tamaralipows: What is this? The wallet nonce? Is it fine for it to always be 0?

tamaralipows (approved):
Thank you for this!! Just a teeny tiny question. Looks wonderful!

foundry/scripts/set-roles.js
dianacarvalho1: no, this is the nonce offset. Inside of the `proposeTransaction` function, we are getting the correct nonce. We then add the offset to it.
This is relevant because if we have a transaction waiting to be signed with nonce 10, when we propose a new tx, the wallet nonce will be 10 too (because the other one hasn't been executed for real) and this will create conflicting txs and we can't execute both. One will override the other

tamaralipows (approved):
Thank you again!

PR #237, opened 2025-07-08
test: Add encoding test for a uniswap X integration test

dianacarvalho1:


tests/uniswap_x_integration_tests.rs
```diff
@@ -0,0 +1,117 @@
...
+            bytes_to_address(&router_address()).unwrap(),
+        )
+        .unwrap();
+
+    let token_out_approval_needed = token_approvals_manager
+        .approval_needed(bytes_to_address(&usdc).unwrap(), filler_address, unix_reactor)
+        .unwrap();
+
+    let full_calldata =
+        (token_in_approval_needed, token_out_approval_needed, tycho_calldata).abi_encode_packed();
```
dianacarvalho1, line 119: I just assumed this. @tamaralipows if you think of a better way, we can change this easily!

tamaralipows: I think this is fine! I don't see a need to make it more complicated (or put extra effort to simplifying) honestly, its very understandable.

tamaralipows (commented):
Had a look here - all is good! Trying to finish up my approvals PR in the meantime to test this :)

tests/uniswap_x_integration_tests.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    let hex_calldata = encode(&full_calldata);
+    write_calldata_to_file("test_sequential_swap_unix", hex_calldata.as_str());
```
tamaralipows: I know we have discussed it, but I should have written it down - we aren't going to be responsible for the encoding of the `SignedOrder` to call `execute`? We are only responsible for the callback data?

I'm asking because I'm wondering if we can just transfer directly to the tycho router during the encoding of the signed order, since this is how their transfer works in the reactor:

```
    /// @inheritdoc BaseReactor
    function _transferInputTokens(ResolvedOrder memory order, address to) internal override {
        permit2.permitWitnessTransferFrom(
            order.toPermit(),
            order.transferDetails(to), <-------- is there a reason this needs to be our filler contract?
            order.info.swapper,
            order.hash,
            V2DutchOrderLib.PERMIT2_ORDER_TYPE,
            order.sig
        );
    }
```

dianacarvalho1: huuummm 🤔 isn't that only for [direct fillers](https://docs.uniswap.org/contracts/uniswapx/guides/mainnet/createfiller#1-defining-a-filler-execution-strategy)?

tamaralipows: What makes you say that? I am following the callstack of `executeWithCallback` in the [reactor contract](https://etherscan.io/address/0x00000011F84B9aa48e5f8aA8B9897600006289Be#code) and this is how they transfer input tokens 🤔 maybe I'm missing something?

dianacarvalho1: aaahhh sorry I misunderstood 🤔 ah you think that could already be the first pool/tycho router??

dianacarvalho1: I honestly thought that was hardcoded on their side to be the filler

tamaralipows: AFAIK it's not hardcoded

dianacarvalho1: i think it is [here](https://github.com/Uniswap/UniswapX/blob/4c01ff07cb16df7ad7f5c2b8e9253005c1259275/src/reactors/BaseReactor.sol#L99)

foundry/test/assets/calldata.txt
```diff
@@ -33,3 +33,4 @@ test_encode_balancer_v3:7bc3485026ac48b6cf9baf0a377477fff5703af8c71ea051a5f82c67
...
+test_sequential_swap_unix:0101e21dd0d300000000000000000000000000000000000000000000006c6b935b8bbd4000000000000000000000000000006b175474e89094c44da98b954eedeac495271d0f000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec700000000000000000000000000000000000000000000000000000000769cfd8000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc20000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000000bf00692e234dae75c793f67a35089c9d99245e1c58470b6b175474e89094c44da98b954eedeac495271d0fa0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000643041cbd36888becc7bbcbc0045e3b1f144466f5f5777d92f208679db4b9778590fa3cab3ac9e2168010000525615deb798bb3e4dfa0139dfa1b3d433cc23b72fa0b86991c6218b36c1d19d4a2e9eb0ce3606eb483041cbd36888becc7bbcbc0045e3b1f144466f5fcd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2010200
```
tamaralipows: unix sounds like the operating system lol. changing this to usx

foundry/test/uniswap_x/UniswapXFiller.t.sol
```diff
@@ -119,7 +116,45 @@ contract UniswapXFillerTest is Test, TychoRouterTestSetup {
         assertGe(IERC20(DAI_ADDR).balanceOf(BOB), amountOut);
     }
 
-    function testExecuteIntegration() public {
+    function testCallbackIntegration() public {
```
dianacarvalho1: aaaahhhh can we do a full test though?? starting in the execute like [this test](https://github.com/propeller-heads/tycho-execution/pull/229/files#diff-a6270dc8c2cec81d38840df2c2260c71e5bc181a9ecd37f249c9d7142b603194R23)

tamaralipows: oh :/ That's what I asked [here](https://github.com/propeller-heads/tycho-execution/pull/237#discussion_r2193502060), I thought we would give an encoding example of that...

> I know we have discussed it, but I should have written it down - we aren't going to be responsible for the encoding of the SignedOrder to call execute? We are only responsible for the callback data?

I guess this can be done on the solidity side? I'll put it last on my list for today

dianacarvalho1 (commented):
Thank yoouu 🙏🏼 I would approve if I could 😅

tamaralipows (approved):

PR #238, opened 2025-07-08
feat: Handle approvals in UniswapXFiller

tamaralipows:


foundry/src/uniswap_x/UniswapXFiller.sol
```diff
@@ -58,12 +59,34 @@ contract UniswapXFiller is AccessControl, IReactorCallback {
...
+                tychoRouter, order.input.maxAmount
+            );
+        } else {
+            // The filler will manually transfer into the TychoRouter
+            // Note: We are using the balance of our filler instead of the order
+            // input amount to avoid having to do a decay calculation in the filler.
+            IERC20 inputToken = IERC20(order.input.token);
+            inputToken.transfer(
+                tychoRouter, inputToken.balanceOf(address(this))
+            );
```
dianacarvalho1: why do we need this case? 
I would just assume that if the `tokenInApprovelNeeded` is false, it's because it was already done before and we are good to go. the router will still take the input tokens from the filler

tamaralipows: I think we need to have a conversation about the flow of tokens here lol, because I'm confused. Let's do it when you're back from lunch

foundry/src/uniswap_x/UniswapXFiller.sol
```diff
@@ -58,12 +67,22 @@ contract UniswapXFiller is AccessControl, IReactorCallback {
...
+            bool tokenInApprovalNeeded,
+            bool tokenOutApprovalNeeded,
+            bytes memory tychoCalldata
+        ) = abi.decode(callbackData, (bool, bool, bytes));
+
+        // The TychoRouter will take the input tokens from the filler
+        if (tokenInApprovalNeeded) {
+            // Native ETH input is not supported by UniswapX
+            IERC20(order.input.token).forceApprove(
+                tychoRouter, order.input.maxAmount
```
dianacarvalho1: weren't we going to force it to `type(uint256).max`?

foundry/src/uniswap_x/UniswapXFiller.sol
```diff
@@ -58,12 +67,22 @@ contract UniswapXFiller is AccessControl, IReactorCallback {
 
         ResolvedOrder memory order = resolvedOrders[0];
 
-        // TODO properly handle native in and out tokens
-        uint256 ethValue = 0;
+        (
+            bool tokenInApprovalNeeded,
+            bool tokenOutApprovalNeeded,
+            bytes memory tychoCalldata
+        ) = abi.decode(callbackData, (bool, bool, bytes));
```
dianacarvalho1: this is assumi abi encoding (not packed). In the encoding I assumed it would be packed 😕

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼  I have one concern!

foundry/src/uniswap_x/UniswapXFiller.sol
tamaralipows: booo that makes it a bit less readable. I'll make it packed then

foundry/src/uniswap_x/UniswapXFiller.sol
tamaralipows: oooops good catch

dianacarvalho1 (approved):
thank you thank you 🙏🏼

PR #241, opened 2025-07-11
feat: UniswapX encoding example

tamaralipows:


examples/uniswapx-encoding-example/main.rs
```diff
@@ -0,0 +1,172 @@
...
+    evm::{
+        approvals::protocol_approvals_manager::ProtocolApprovalsManager,
+        encoder_builders::TychoRouterEncoderBuilder,
+        utils::{biguint_to_u256, bytes_to_address},
+    },
+    models::{Solution, Swap, UserTransferType},
+};
+
+/// Encodes the input data for a function call to the given function selector.
+pub fn encode_input(selector: &str, mut encoded_args: Vec<u8>) -> Vec<u8> {
```
dianacarvalho1, line 20: huuummm this raises another question for me (not in the scope of this PR though). Should we change the basic example to encode the full thing as well? 🤔

examples/uniswapx-encoding-example/main.rs
```diff
@@ -0,0 +1,172 @@
...
+                .collect::<Vec<u8>>()
+    {
+        encoded_args = encoded_args[32..].to_vec();
+    }
+    call_data.extend(encoded_args);
+    call_data
+}
+
+fn main() {
+    let router_address = Bytes::from_str("0xfD0b31d2E955fA55e3fa641Fe90e08b677188d35")
```
dianacarvalho1, line 42: I think we shouldn't pass this address

examples/uniswapx-encoding-example/main.rs
```diff
@@ -0,0 +1,172 @@
...
+        .expect("Failed to create router address");
+
+    // Initialize the encoder
+    let encoder = TychoRouterEncoderBuilder::new()
+        .chain(tycho_common::models::Chain::Ethereum)
+        .user_transfer_type(UserTransferType::TransferFrom)
+        .executors_file_path("config/test_executor_addresses.json".to_string())
```
dianacarvalho1: and neither this

examples/uniswapx-encoding-example/main.rs
```diff
@@ -0,0 +1,172 @@
...
+    let token_out_approval_needed = token_approvals_manager
+        .approval_needed(bytes_to_address(&usdc).unwrap(), filler_address, usx_reactor)
+        .unwrap();
+
+    let full_calldata =
+        (token_in_approval_needed, token_out_approval_needed, tycho_calldata).abi_encode_packed();
+
+    let hex_calldata = encode(&full_calldata);
+
+    println!(" ====== Simple swap WETH -> USDC ======");
```
dianacarvalho1: it's DAI to USDT

dianacarvalho1 (commented):
Thank you! Tiny tiny comments only

examples/uniswapx-encoding-example/main.rs
tamaralipows: I only did it here because the callback data doesn't make any sense without it (this is the TychoRouter call). I am technically not encoding the full thing, since I'm excluding the execution calldata. So I think it's fine to leave the basic example as is

examples/uniswapx-encoding-example/main.rs
tamaralipows: What address should we pass? Take it as input? We need this for the approvals

dianacarvalho1: aaah shit, good point. we can't get it from the encoder right? 😬

tamaralipows: No, because the struct isn't known at compile time - the builder returns a `Box<dyn TychoEncoder>` :(  I would need to add another method to the trait to return the router address. Do you think it's worth it?

dianacarvalho1: oh actually no. because in the filler contract we are hardcoding the router address. So they need to know it and make sure it matches here!! AG! sorry for the confusion 😅

dianacarvalho1 (approved):
yes me like! 👍🏼 thank you!

PR #242, opened 2025-07-15
feat: Uniswap X deployment script

dianacarvalho1:
Deployed one [here](https://etherscan.io/address/0xfC944C02C0064c829B52CC8A57EAf64ed2aeb103#code) to test the script!

### DIFF
--- a/foundry/scripts/deploy-uniswap-x-filler.js
+++ b/foundry/scripts/deploy-uniswap-x-filler.js
@@ -1,27 +1,28 @@
 require('dotenv').config();
 const {ethers} = require("hardhat");
 const hre = require("hardhat");
+const path = require("path");
+const fs = require("fs");
 
 async function main() {
     const network = hre.network.name;
-    let tychoRouter;
     let uniswapXReactor;
     let nativeToken;
     if (network === "ethereum") {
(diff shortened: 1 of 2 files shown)


foundry/scripts/deploy-uniswap-x-filler.js
```diff
@@ -0,0 +1,62 @@
...
+const {ethers} = require("hardhat");
+const hre = require("hardhat");
+
+async function main() {
+    const network = hre.network.name;
+    let tychoRouter;
+    let uniswapXReactor;
+    let nativeToken;
+    if (network === "ethereum") {
+        tychoRouter = "0xfD0b31d2E955fA55e3fa641Fe90e08b677188d35";
```
tamaralipows: Can't we actually read this from the file (the tycho router)? Similar to how we read the executors that we want to set? So it's less maintenance

tamaralipows (commented):
Thank you! Just one thing, let me know what you think

foundry/scripts/deploy-uniswap-x-filler.js
dianacarvalho1: good idea!

tamaralipows (approved):
Thank you!

PR #245, opened 2025-07-22
feat: Add protocol state to Swap object

tamaralipows:
- This consists of lots of necessary battles with lifetimes.


src/encoding/evm/group_swaps.rs
```diff
@@ -210,13 +223,15 @@ mod tests {
...
         };
-        let grouped_swaps = group_swaps(&vec![
+        let swaps = vec![
             swap_wbtc_weth.clone(),
             swap_weth_usdc.clone(),
             swap_weth_dai.clone(),
             swap_dai_usdc.clone(),
-        ]);
+        ];
+        let grouped_swaps = group_swaps(&swaps);
```
dianacarvalho1, line 234: why are these changes necessary?

src/encoding/models.rs
```diff
@@ -84,17 +86,32 @@ pub struct Swap {
     pub split: f64,
     /// Optional user data to be passed to encoding.
     pub user_data: Option<Bytes>,
+    /// Optional protocol state used to perform the swap.
+    #[serde(skip)]
```
dianacarvalho1, line 90: why serde skip?

src/encoding/models.rs
```diff
@@ -84,17 +86,32 @@ pub struct Swap {
...
+}
+
+impl<'a> PartialEq for Swap<'a> {
+    fn eq(&self, other: &Self) -> bool {
+        self.component == other.component &&
+            self.token_in == other.token_in &&
+            self.token_out == other.token_out &&
+            self.split == other.split &&
+            self.user_data == other.user_data
+        // Skip protocol_state comparison since trait objects don't implement PartialEq
```
dianacarvalho1, line 114: if the component matches I think it's fair to say that the ProtocolSim will match too

dianacarvalho1 (approved):
oh uau, this was nicer than I anticipated! 🙏🏼 thank you. only have questions to understand better what is happening

src/encoding/models.rs
tamaralipows: To avoid having to derive Serialize in the Swap object, which would be unnecessary. Let me know if you disagree

```
error[E0277]: the trait bound `dyn ProtocolSim: Serialize` is not satisfied
    --> src/encoding/models.rs:76:37
     |
76   |   #[derive(Clone, Debug, Deserialize, Serialize)]
     |                                       ^^^^^^^^^ unsatisfied trait bound
     ```

File: src/encoding/models.rs (hunk 3 above)
Comment (tamaralipows) [reply src/encoding/models.rs:114]: Another reason for not explicitly checking. Wouldn't like to have to implement `PartialEq` for `ProtocolSim`

File: src/encoding/evm/group_swaps.rs (hunk 1 above)
Comment (tamaralipows) [reply src/encoding/evm/group_swaps.rs:234]:
```
error[E0716]: temporary value dropped while borrowed
   --> src/encoding/evm/group_swaps.rs:143:42
    |
143 |           let grouped_swaps = group_swaps(&vec![swap_weth_wbtc.clone(), swap_wbtc_usdc.clone(), swap_usdc_dai.clone()]);
    |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - temporary value is freed at the end of this statement
    |                                            |
    |                                            creates a temporary value which is freed while still in use
144 |
145 | /         assert_eq!(
146 | |             grouped_swaps,
147 | |             vec![
148 | |                 SwapGroup {
...   |
163 | |         );
    | |_________- borrow later used here
    |
    = note: this error originates in the macro `vec` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider using a `let` binding to create a longer lived value
    |
143 ~         let binding = vec![swap_weth_wbtc.clone(), swap_wbtc_usdc.clone(), swap_usdc_dai.clone()];
144 ~         let grouped_swaps = group_swaps(&binding);
```

PR #248, opened 2025-08-06
fix: Encode sequential swaps with unwrapping WETH correctly

dianacarvalho1:
This is a bug reported by Wraxyn


tamaralipows (approved):
Thank you! This was also the solution I imagined

PR #249, opened 2025-08-11
feat: add hashflow executor

adrianbenavides:


foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+// Additional parameters not related to the quote
+struct SwapParams {
+    bool needsApproval;
+    address recipient;
+}
+
+contract HashflowExecutor is IExecutor, RestrictTransferFrom {
+    using SafeERC20 for IERC20;
+
+    address public immutable hashflowRouter; // TODO: should this be a constant?
```
dianacarvalho1: yes that is ok

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+        // Verify that the input token amount matches with the quote data
+        if (givenAmount != quote.baseTokenAmount) {
+            revert HashflowExecutor__AmountMismatch();
+        }
+
+        // Handle token approval if needed
+        if (params.needsApproval) {
+            // TODO: side effects or previous state checks?
+            // slither-disable-next-line unused-return
+            IERC20(quote.baseToken).approve(hashflowRouter, givenAmount);
```
dianacarvalho1: here it should be forceApprove. There are some weird tokens (like USDT) where normal approve doesn't work

dianacarvalho1: also I think we can just approve the max amount possible so we can try to save a bit on gas. you can do
```
IERC20(quote.baseToken).forceApprove(hashflowRouter, type(uint256).max)
```

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // Prepare the quote for the Hashflow router
+        IHashflowRouter.RFQTQuote memory rfqtQuote = IHashflowRouter.RFQTQuote({
```
dianacarvalho1: the quote is already of the correct type no? 🤔

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+
+        uint256 balanceAfter = IERC20(quote.quoteToken).balanceOf(address(this));
+        calculatedAmount = balanceAfter - balanceBefore;
+    }
+
+    function _decodeData(bytes calldata data)
+        internal
+        pure
+        returns (
+            //TODO: are these variables being initialized correctly?
```
dianacarvalho1: yes, this is correct

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+        ptr += 32;
+
+        quote.txid = bytes32(data[ptr:ptr + 32]);
+        ptr += 32;
+
+        // Decode signature
+        uint256 sigLen = uint256(bytes32(data[ptr:ptr + 32]));
+        ptr += 32;
+
+        quote.signature = data[ptr:ptr + sigLen];
```
dianacarvalho1: We don't need to pass the signature length. Since it is the last param we can just do:
`quote.signature = data[ptr:];`

foundry/test/protocols/Hashflow.t.sol
```diff
@@ -0,0 +1,140 @@
...
+        bytes memory quoteData = createSampleEncodedData();
+
+        deal(WETH_ADDR, address(executor), amountIn);
+
+        vm.expectRevert(HashflowExecutor__AmountMismatch.selector);
+        executor.swap(amountIn, quoteData);
+    }
+
+    // Encoded data sample for testing
+    function createSampleEncodedData() internal view returns (bytes memory) {
```
dianacarvalho1: This is useful for the decoding test but for the swap test this will never work out 😬

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    // TODO: is recipient needed here?
+    function tradeRFQT(RFQTQuote calldata quote, address recipient) external;
+}
+
+// Additional parameters not related to the quote
+struct SwapParams {
+    bool needsApproval;
+    address recipient;
+}
```
dianacarvalho1: I think we don't need to have a struct for these extra params.. we can just pass them as individual variables

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+        payable
+        returns (uint256 calculatedAmount)
+    {
+        // Decode swap data
+        (IHashflowRouter.RFQTQuote memory quote, SwapParams memory params) =
+            _decodeData(data);
+
+        // Verify that the input token amount matches with the quote data
+        if (givenAmount != quote.baseTokenAmount) {
+            revert HashflowExecutor__AmountMismatch();
```
dianacarvalho1: This is not necessarily true. But let's get the other things out of the way before we address this. Just leave this comment unresolved so we don't forget please 🙏🏼

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+            quoteExpiry: quote.quoteExpiry,
+            nonce: quote.nonce,
+            txid: quote.txid,
+            signature: quote.signature
+        });
+
+        uint256 balanceBefore =
+            IERC20(quote.quoteToken).balanceOf(address(this));
+
+        IHashflowRouter(hashflowRouter).tradeRFQT(rfqtQuote, params.recipient);
```
dianacarvalho1: according to the contract's ABI there is no recipient here

dianacarvalho1 (commented):
Thank you @adrianbenavides ! Great job here already 👏🏼  you have a skeleton for the Executor and working tests! Very good! ✨ If you have more specific questions that I didn't catch, please let me know

foundry/src/executors/HashflowExecutor.sol
adrianbenavides: right, I think that's a leftover from a previous version, thanks!

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,167 @@
...
+        ptr += 32;
+
+        quote.quoteExpiry = uint256(bytes32(data[ptr:ptr + 32]));
+        ptr += 32;
+
+        quote.nonce = uint256(bytes32(data[ptr:ptr + 32]));
+        ptr += 32;
+
+        quote.txid = bytes32(data[ptr:ptr + 32]);
+        ptr += 32;
```
tamaralipows: Interesting. We don't use this pattern in our other executors (storing the pointer instead of hardcoding the indices). I can see the advantage of doing it this way! I'm curious if it costs more gas (I'm not sure), but we can easily double check later

foundry/test/protocols/Hashflow.t.sol
adrianbenavides: This should be fixed now!

foundry/src/executors/HashflowExecutor.sol
adrianbenavides: > I'm curious if it costs more gas (I'm not sure)

Hardcoded indices costs slightly less (from `300521` down to `300159`). I'll use the common pattern just to not introduce more noise.

foundry/test/protocols/Hashflow.t.sol
```diff
@@ -0,0 +1,138 @@
...
+
+    function testDecodeParamsInvalidDataLength() public {
+        bytes memory invalidData = new bytes(10);
+        vm.expectRevert(HashflowExecutor__InvalidDataLength.selector);
+        executor.decodeData(invalidData);
+    }
+
+    function testSwapOk() public {
+        uint256 amountIn = 1;
+        address trader = address(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f);
```
dianacarvalho1: this is the `address(executor)` right?

dianacarvalho1 (commented):
Awesoomeeee 👏🏼 👏🏼  very very good! I have only 1 more comment and for the rest let's have a quick call!

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        // Handle token approval if needed
+        if (approvalNeeded) {
+            // slither-disable-next-line unused-return
+            IERC20(quote.baseToken).forceApprove(
+                HASHFLOW_ROUTER, type(uint256).max
+            );
+        }
+
+        uint256 balanceBefore =
```
dianacarvalho1: before this we need something like
````
 _transfer(address(this), transferType, address(tokenIn), givenAmount);
````

File: foundry/src/executors/HashflowExecutor.sol [hunk 13]
```diff
@@ -0,0 +1,125 @@
...
+            // ERC20
+            if (approvalNeeded) {
+                baseTokenECR20.forceApprove(HASHFLOW_ROUTER, type(uint256).max);
+            }
+            _transfer(
+                address(this),
+                transferType,
+                address(quote.baseToken),
+                givenAmount
+            );
```
tamaralipows: Is transferType always gonna be None here, since the hashflow router performs the transfer? In that case the call to transfer is unnecessary. @dianacarvalho1 can you confirm please?

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,125 @@
...
+            IHashflowRouter.RFQTQuote memory quote,
+            bool approvalNeeded,
+            TransferType transferType
+        ) = _decodeData(data);
+
+        // Slippage checks
+        if (givenAmount > quote.baseTokenAmount) {
+            // Do not transfer more than the quote's maximum permitted amount.
+            givenAmount = quote.baseTokenAmount;
+        }
```
tamaralipows, line 57: Interesting, do the extra tokens from positive slippage just live in our router then? I guess so

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,125 @@
...
+        pure
+        returns (
+            IHashflowRouter.RFQTQuote memory quote,
+            bool approvalNeeded,
+            TransferType transferType
+        )
+    {
+        if (data.length < 282) {
+            revert HashflowExecutor__InvalidDataLength();
+        }
```
tamaralipows: I think the signature is a fixed length, right? AFAIK the bebop contracts will revert otherwise... So this could perhaps be an inequality check instead to make it safer?

foundry/test/protocols/Hashflow.t.sol
```diff
@@ -0,0 +1,276 @@
...
+    using SafeERC20 for IERC20;
+
+    HashflowExecutorExposed executor;
+    uint256 forkBlock;
+
+    IERC20 WETH = IERC20(WETH_ADDR);
+    IERC20 USDC = IERC20(USDC_ADDR);
+
+    function setUp() public {
+        forkBlock = 23125321; // Using expiry date: 1755006017, Native
```
tamaralipows, line 216: I'm curious if using the same forkBlock `23124977` would work? Since it's still below this expiry date

tamaralipows (commented):
This is great, a huuuuge congratulations to you for figuring this out @adrianbenavides! I've put a few questions here for my own understanding, if you don't know the answer then perhaps @dianacarvalho1 could answer them too. Thank you again!

foundry/src/executors/HashflowExecutor.sol
dianacarvalho1: No no, it will not always be None. It is None now for the sake of these tests. In the router integration tests, it shouldn't be None

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,125 @@
...
+
+        // Slippage checks
+        if (givenAmount > quote.baseTokenAmount) {
+            // Do not transfer more than the quote's maximum permitted amount.
+            givenAmount = quote.baseTokenAmount;
+        }
+        quote.effectiveBaseTokenAmount = givenAmount;
+
+        IERC20 baseTokenECR20 = IERC20(quote.baseToken);
+        IERC20 quoteTokenECR20 = IERC20(quote.quoteToken);
```
dianacarvalho1: it might happen that one of these tokens is ETH. So then this will give weird things..

dianacarvalho1: I would suggesting writing your own `_balanceOf` method that takes care if the token is ETH or an ERC20 token. (there are some methods like this in the repo already)

foundry/src/executors/HashflowExecutor.sol
```diff
@@ -0,0 +1,125 @@
...
+
+        // Do the swap
+        if (quote.baseToken == address(0)) {
```
dianacarvalho1: I think this if-else condition could be simplified.. could we save the eth_value in an extra variable like:
````
uint256 ethValue=0;
if (quote.baseToken == address(0)){
    ethValue=quote.effectiveBaseTokenAmount
}
_transfer(
    address(this),
    transferType,
    address(quote.baseToken),
    givenAmount
);
uint256 balanceBefore = quoteTokenECR20.balanceOf(address(this));
IHashflowRouter(HASHFLOW_ROUTER).tradeRFQT{
    value: ethValue
}(quote);
uint256 balanceAfter = quoteTokenECR20.balanceOf(address(this));
calculatedAmount = balanceAfter - balanceBefore;
````

File: foundry/test/protocols/Hashflow.t.sol [hunk 19]
```diff
@@ -0,0 +1,276 @@
...
+        address trader = address(executor);
+        IHashflowRouter.RFQTQuote memory quote = rfqtQuote();
+        uint256 amountIn = quote.baseTokenAmount;
+        bytes memory encodedQuote = encodeRfqtQuoteWithDefaults(quote);
+
+        deal(USDC_ADDR, address(executor), amountIn);
+        uint256 balanceBefore = WETH.balanceOf(trader);
+
+        vm.prank(trader);
+        uint256 amountOut = executor.swap(amountIn, encodedQuote);
```
dianacarvalho1, line 135: can we assert that amount out is the value we expect and got in the quote?

foundry/test/protocols/Hashflow.t.sol
```diff
@@ -0,0 +1,276 @@
...
+                uint256(
+                    0x12500006400064000000174813b960ffffffffffffff00293fdb4569fe760000
+                )
+            ),
+            signature: hex"5b26977fecaf794c3d6900b9523b9632b5c62623f92732347dc9f24d8b5c4d611f5d733bbe82b594b6b47ab8aa1923c9f6b8aa66ef822ce412a767200f1520e11b"
+        });
+    }
+}
+
+contract HashflowExecutorNativeTest is Constants, HashflowUtils {
```
dianacarvalho1, line 206: huummm i think we don't need separate contracts.. What I meant in slack was to do the setup inside of each test case, like
```
function testSwapNoSlippage() public {
        forkBlock = 23125321; // Using expiry date: 1755006017, Native
        vm.createSelectFork("mainnet", forkBlock);
        executor = new HashflowExecutorExposed(PERMIT2_ADDRESS);
... your normal test
```

dianacarvalho1 (commented):
Thank you @adrianbenavides ! 👏🏼 This looks really really good already! Great job! I only have a few more comments

foundry/src/executors/HashflowExecutor.sol
tamaralipows: Ahhhhh because we will sometimes need to send to our own router! Got it! I was only thinking about the hashflow router

foundry/src/executors/HashflowExecutor.sol
tamaralipows: I think you can still wrap the eth address in IERC20 without erroring but yeah it doesn't make a lot of sense to put it outside the if block

foundry/src/executors/HashflowExecutor.sol
adrianbenavides: I couldn't tell from the [hashflow ABI](https://github.com/hashflownetwork/x-protocol/blob/main/evm/abi/IHashflowRouter.json#L805). Do you know if the signatures follow a particular "standard" format?

foundry/test/protocols/Hashflow.t.sol
adrianbenavides: I tried but it won't work because of the signature (the expiry is part of the quote data).

foundry/test/protocols/Hashflow.t.sol
adrianbenavides: I tried it, but I couldn't make it work. I had to "make the ETH token persistent" (following a compiler warning), but even then the test was failing. I thought it might be because I was creating two forks?

foundry/src/executors/HashflowExecutor.sol
adrianbenavides: Oh I see, so if `{value: 0}` then it's ignored / has no effect?

foundry/src/executors/HashflowExecutor.sol
adrianbenavides: Got it, it looks way cleaner now after the improvements you suggested. Thanks!

foundry/test/protocols/Hashflow.t.sol
dianacarvalho1: huumm that is strange.. did you remove it from the setup method?

adrianbenavides: Nope, but if I do then I need to do the setup manually in all the tests only because one test needs a special setup. That felt like it was asking for a different approach 😆

dianacarvalho1 (approved):
Veeery good! Thank you 🙏🏼 This looks good to me! ✨

foundry/src/executors/HashflowExecutor.sol
tamaralipows: As far as I know it uses an ECDSA signature which is 65 bytes. Could you double check if it works to consider that in the length?

tamaralipows (approved):
Looks good! Just had one comment about the data length, but if you can't make it work don't let it block you

foundry/src/executors/HashflowExecutor.sol
adrianbenavides: That worked, yes. Passing a signature with a different length makes hashflow return a `ECDSA: invalid signature length` error.

PR #251, opened 2025-08-14
feat: Support Bebop quote through IndicativelyPriced state

dianacarvalho1:
Done: 
- I used the runtime like we were doing before (we can reevaluate during cooldown if this is the best way to go or not)
- Made a new feature gate (test-utils) for MockRFQState 
  - I wanted to use this in the integration tests to make sure that the receivers of a sequential trade were being set correctly


src/encoding/evm/mod.rs
```diff
@@ -5,5 +5,7 @@ mod encoding_utils;
 mod group_swaps;
 pub mod strategy_encoder;
 mod swap_encoder;
+#[cfg(feature = "test-utils")]
+pub mod testing_utils;
```
tamaralipows, line 9: Is it not possible to put this in the tests dir? Or is that bad practice? I don't know honestly

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -703,25 +734,98 @@ impl SwapEncoder for BebopSwapEncoder {
...
+                    swap.estimated_amount_in
+                        .clone()
+                        .ok_or(EncodingError::FatalError(
+                            "Estimated amount in is mandatory for a Bebop swap".to_string(),
+                        ))?;
+                // Bebop uses another address for the native token than the zero address
+                let bebop_native_address = Bytes::from_str(&self.native_token_bebop_address)
+                    .map_err(|_| {
+                        EncodingError::FatalError(
+                            "Invalid Bebop native token curve address".to_string(),
```
tamaralipows: what's a curve address?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -669,7 +687,21 @@ impl SwapEncoder for BebopSwapEncoder {
                 "Missing bebop settlement address in config".to_string(),
             ))?
             .to_string();
-        Ok(Self { executor_address, settlement_address })
+        let native_token_bebop_address = config
+            .get("native_token_address")
+            .ok_or(EncodingError::FatalError(
+                "Missing native token bebop address in config".to_string(),
+            ))?
+            .to_string();
```
tamaralipows, line 695: Should we perhaps already save this as bytes to avoid conversion below? Or do we need it somewhere as string

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -703,25 +734,98 @@ impl SwapEncoder for BebopSwapEncoder {
...
+                    receiver: encoding_context.receiver.clone(),
+                };
+                let signed_quote = block_in_place(|| {
+                    self.runtime_handle.block_on(async {
+                        indicatively_priced_state
+                            .request_signed_quote(params)
+                            .await
+                    })
+                })
+                .map_err(|e| EncodingError::FatalError(format!("Failed to get Bebop quote {e}")))?;
```
tamaralipows: Is this going to be a fatal error even if there is an internet issue?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -703,25 +734,98 @@ impl SwapEncoder for BebopSwapEncoder {
...
+                let partial_fill_offset = signed_quote
+                    .quote_attributes
+                    .get("partial_fill_offset")
+                    .ok_or(EncodingError::FatalError(
+                        "Bebop quote must have a partial_fill_offset attribute".to_string(),
+                    ))?;
+                let original_filled_taker_amount = biguint_to_u256(&signed_quote.amount_out);
+                (
+                    // we are only interested in the last byte to get a u8
+                    partial_fill_offset[partial_fill_offset.len() - 1],
```
tamaralipows, line 798: Does it return more than one byte normally in the response? Why? If we only need the last

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -703,25 +734,98 @@ impl SwapEncoder for BebopSwapEncoder {
...
-        }
-
-        let partial_fill_offset = user_data[0];
-        let original_filled_taker_amount = U256::from_be_slice(&user_data[1..33]);
-
-        // The calldata should be for either swapSingle or swapAggregate
-        let bebop_calldata = user_data[33..].to_vec();
-
+        let (partial_fill_offset, original_filled_taker_amount, bebop_calldata) =
+            if let Some(state) = swap.protocol_state {
```
tamaralipows, line 742: Could you explain (maybe in a comment/doc but also in this thread) when we can expect there to be a state, and when we rely on user data instead? I'm forgetting

tests/optimized_transfers_integration_tests.rs
```diff
@@ -608,3 +609,92 @@ fn test_uniswap_v3_balancer_v3() {
     let hex_calldata = encode(&calldata);
     write_calldata_to_file("test_uniswap_v3_balancer_v3", hex_calldata.as_str());
 }
+
+#[test]
+fn test_uniswap_v3_bebop() {
+    // Note: This test does not assert anything. It is only used to obtain
+    // integration test data for our router solidity test.
+    //
+    // Performs a sequential swap from WETH to ONDO through USDC using USV3 and
```
tamaralipows:
```suggestion
    // Performs a sequential swap from WETH to WBTC through USDC using USV3 and
```

tamaralipows (commented):
Thank you!! Just a few questions mostly

src/encoding/evm/mod.rs
dianacarvalho1: you mean to put this in the integration tests directory? and then use it here?? 🤔 I'm not even sure that is possible tbh

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: upsi.. copy pasting is the greatest evil to dev

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: it is left as string so we don't need to change the signature of the `new()` method 😬

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: huuuummmm 🤔 yes because then the encoding for this whole solution would fail

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: I'm keeping things coherent with how we did in the bebop client.. we are abi encoding there basically

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: yeah.. I'm also not sure how to make this super duper clear

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: I don't understand, do you mean the struct field? I think it's fine to change that, no?

src/encoding/evm/mod.rs
tamaralipows: i probably have python brain

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: But that is considered recoverable (i.e. try again and it may pass) no? Or am I missing something

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: ewww. okay

src/encoding/evm/mod.rs
dianacarvalho1: nop, I just tested. the main crate can not access anything inside of the integration test 😬

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: no, I mean the swap encoder new. But I double checked now and what I said it's not true. sorry I was confused! I will change this!

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: remember that execution is not a running program that stays forever.. it is a library that you call. So for this particular encoding it is a fatal error. I don't think it's a recoverable error because the swap calldata can no be recovered somehow

tamaralipows: Hmm what do you mean by the last sentence?
Are you implying that we should never have recoverable error in execution?
Because we are returning it in the approvals manager for example, and another time when getting permit2 existing allowances

tamaralipows: I don't think it being a library that you call discounts the need for a recoverable error.

FatalError = something wrong in our code. SOS.
RecoverableError = try again in a bit?

dianacarvalho1: huuummmm but in this case in particular this might fail for many reasons.. it could be a connection error, it could be that Bebop does not support this tokens anymore, etc etc 😬

adrianbenavides: Would it make sense if we just bubble up the SimulationError returned by the request_signed_quote call? It can be mapped 1:1 into an EncodingError.

tamaralipows: >  it could be that Bebop does not support this tokens anymore

This would rather then be an `InvalidInput` error - so still not fatal :/ 

@adrianbenavides has a good point about bubbling up the SimulationError! I totally missed that point! We do have this information

dianacarvalho1: > This would rather then be an InvalidInput error - so still not fatal :/

not necessarily. Because the quote for this pool state (with this certain token pair) will never work then 😬

dianacarvalho1: I did [this](https://github.com/propeller-heads/tycho-execution/pull/251/commits/987956eb0bea2a46115d160ea05db395ce17077a). What do you think @tamaralipows @adrianbenavides ?

tamaralipows: Perfect!!!

tamaralipows (approved):
Thank you!

PR #252, opened 2025-08-18
feat: implement SwapEncoder for Hashflow

adrianbenavides:
Implement the `SwapEncoder` trait for Hashflow -- [task ref](https://propeller-heads.atlassian.net/browse/ENG-4750).

Related PR, fixing the quote parsing in the simulation side: https://github.com/propeller-heads/tycho-simulation/pull/304


foundry/test/TychoRouterSequentialSwap.t.sol
```diff
@@ -547,3 +547,37 @@ contract TychoRouterSequentialSwapTestForBebop is TychoRouterTestSetup {
...
+        bytes memory callData = loadCallDataFromFile("test_uniswap_v3_hashflow");
+        (bool success,) = tychoRouterAddr.call(callData);
+
+        vm.stopPrank();
+
+        uint256 balanceAfter = IERC20(WBTC_ADDR).balanceOf(ALICE);
+
+        assertTrue(success, "Call Failed");
+        assertGt(balanceAfter - balanceBefore, 0);
+        assertLe(balanceAfter - balanceBefore, expectedAmountOut);
```
adrianbenavides: Not 100% if this is correct. In this test, the amount out is slightly less than the expected.

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -850,6 +846,148 @@ impl SwapEncoder for BebopSwapEncoder {
...
+        chain: Chain,
+        config: Option<HashMap<String, String>>,
+    ) -> Result<Self, EncodingError> {
+        let config = config.ok_or(EncodingError::FatalError(
+            "Missing hashflow specific addresses in config".to_string(),
+        ))?;
+        let hashflow_router_address = config
+            .get("hashflow_router_address")
+            .ok_or(EncodingError::FatalError(
+                "Missing bebop settlement address in config".to_string(),
```
dianacarvalho1: copy 🍝 gone wrong

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -850,6 +846,148 @@ impl SwapEncoder for BebopSwapEncoder {
...
+                EncodingError::FatalError("protocol_state is required for Hashflow".to_string())
+            })?;
+        let amount_in = swap
+            .estimated_amount_in
+            .as_ref()
+            .ok_or(EncodingError::FatalError(
+                "Estimated amount in is mandatory for a Hashflow swap".to_string(),
+            ))?
+            .clone();
+        let sender = encoding_context
```
dianacarvalho1, line 924: we are actually checking for the presence of the router address twice: once for the approval and another time here. (If it's the same in Bebop feel free to change it as well 😅 )

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -2037,4 +2175,164 @@ mod tests {
...
+
+        fn hashflow_config() -> Option<HashMap<String, String>> {
+            Some(HashMap::from([(
+                "hashflow_router_address".to_string(),
+                "0x55084eE0fEf03f14a305cd24286359A35D735151".to_string(),
+            )]))
+        }
+
+        #[test]
+        fn test_encode_hashflow_single_with_user_data() {
```
dianacarvalho1: how is this working?? Hashflow shouldn't work with user data.. and we are also not passing any user data here.. I'm confused about this test 🤔

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -2037,4 +2175,164 @@ mod tests {
...
+            .unwrap();
+            encoder
+                .encode_swap(&swap, &encoding_context)
+                .expect_err("Should returned an error if the swap has no protocol state");
+        }
+
+        #[test]
+        fn test_encode_hashflow_single_with_protocol_state() {
+            // 3000 USDC -> 1 WETH using a mocked RFQ state to get a quote
+            let partial_fill_offset = 12u64;
```
dianacarvalho1: this is bebop specific

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -2037,4 +2175,164 @@ mod tests {
...
+                ),
+                (
+                    "trader".to_string(),
+                    Bytes::from_str("0xcd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2").unwrap(),
+                ),
+                (
+                    // Passing the tycho router address here has no effect
+                    "effective_trader".to_string(),
+                    Bytes::from_str("0xcd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2").unwrap(),
+                ),
```
dianacarvalho1: please remove it then!

src/encoding/evm/testing_utils.rs
```diff
@@ -17,7 +17,7 @@ use tycho_common::{
 #[derive(Debug)]
 pub struct MockRFQState {
     pub quote_amount_out: BigUint,
-    pub quote_calldata: Bytes,
+    pub quote_data: Vec<(String, Bytes)>,
```
dianacarvalho1: we might as well make it a HashMap already then..
This change breaks the Bebop tests though

src/encoding/evm/testing_utils.rs
```diff
@@ -17,7 +17,7 @@ use tycho_common::{
...
     pub quote_partial_fill_offset: u64,
```
dianacarvalho1: you could assume that the quote_partial_fill_offset would be inside of quote_data as well for Bebop

foundry/test/TychoRouterSequentialSwap.t.sol
dianacarvalho1: why?? does the Univ3 pool outputs less tokens?

foundry/test/TychoRouterSingleSwap.t.sol
```diff
@@ -426,3 +426,36 @@ contract TychoRouterSingleSwapTest is TychoRouterTestSetup {
         assertEq(balanceAfter - balanceBefore, 2018817438608734439722);
     }
 }
+
+contract TychoRouterSingleSwapTestForHashflow is TychoRouterTestSetup {
```
dianacarvalho1: Please move this inside of the Hashflow test file

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -2,27 +2,28 @@
...
 
 // Executors
-import {BalancerV2Executor} from "../src/executors/BalancerV2Executor.sol";
-import {BalancerV3Executor} from "../src/executors/BalancerV3Executor.sol";
-import {BebopExecutor} from "../src/executors/BebopExecutor.sol";
-import {CurveExecutor} from "../src/executors/CurveExecutor.sol";
-import {EkuboExecutor} from "../src/executors/EkuboExecutor.sol";
-import {MaverickV2Executor} from "../src/executors/MaverickV2Executor.sol";
-import {UniswapV2Executor} from "../src/executors/UniswapV2Executor.sol";
+import "../src/executors/HashflowExecutor.sol";
```
dianacarvalho1: oh no.. automatic imports messed things up and now the comments don't make sense

dianacarvalho1 (commented):
Thank you so much @adrianbenavides ! 🙏🏼 this looks reeeaally good! Only have small comments/requests

foundry/test/TychoRouterSequentialSwap.t.sol
tamaralipows: I would double check this too

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -850,6 +846,148 @@ impl SwapEncoder for BebopSwapEncoder {
...
+            } else {
+                approval_needed = ProtocolApprovalsManager::new()?.approval_needed(
+                    bytes_to_address(&swap.token_in)?,
+                    tycho_router_address,
+                    hashflow_router_address,
+                )?;
+            }
+        } else {
+            approval_needed = true;
+        }
```
tamaralipows: I think this (whether the router address is passed) should be checked earlier and maybe error if not passed? Same as Diana's comment below I guess. I would think we don't want to continue if it's not passed.

tamaralipows:
This is a huge complex PR, thank you so much for taking this on as one of your first tasks, and so cleanly too! I have no major comments that Diana didn't already bring up!

src/encoding/evm/swap_encoder/swap_encoders.rs
adrianbenavides: In the last line, the call to `encode_swap` is expected to fail. So the test is checking that the function fails if no protocol state is passed.

src/encoding/evm/swap_encoder/swap_encoders.rs
adrianbenavides: but it's needed here. If you request a quote passing only the `trader` field, the returned quote will have `effective_trader=trader`. With that comment I'm only saying that passing the _tycho router_ address instead of the user's address has no effect.

src/encoding/evm/testing_utils.rs
adrianbenavides: > we might as well make it a HashMap already then..

sure, I'll do that.

> This change breaks the Bebop tests though

I run the `TychoRouterSequentialSwapTestForBebop` test and it worked fine. Were you referring to this one?

src/encoding/evm/swap_encoder/swap_encoders.rs
adrianbenavides: done!

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: aaahhh then please rename it to reflect that 😅

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: aaah it's needed for this particular quote you mean. can you explain that in the comment then please?

src/encoding/evm/testing_utils.rs
dianacarvalho1: huuumm no, I mean all the bebop encoding rust tests that use MockRFQState should fail..

foundry/test/TychoRouterSequentialSwap.t.sol
adrianbenavides: Fixed now, and with an extra assert. I've added more context in the test description. Thanks @dianacarvalho1 for the help navigating the test logs!

src/encoding/evm/testing_utils.rs
adrianbenavides: Why do you think they will fail? All tests are passing without modifying anything, not sure if I'm missing something.

src/encoding/evm/swap_encoder/swap_encoders.rs
adrianbenavides: This field is now removed from the calldata

tests/optimized_transfers_integration_tests.rs
```diff
@@ -698,3 +710,233 @@ fn test_uniswap_v3_bebop() {
     let hex_calldata = encode(&calldata);
     write_calldata_to_file("test_uniswap_v3_bebop", hex_calldata.as_str());
 }
+
+#[test]
+fn test_hashflow() {
```
dianacarvalho1: this test should be moved to `protocol_integration_tests.rs` please!

src/encoding/evm/testing_utils.rs
dianacarvalho1: ah I see that you already adapted the bebop tests to this! Sorry I think I was looking at an older version! all good

tests/optimized_transfers_integration_tests.rs
adrianbenavides: should I rename it to `test_single_encoding_strategy_hashflow`?

dianacarvalho1: yeah that would be good!

adrianbenavides: done!

Cargo.toml
```diff
@@ -37,7 +37,8 @@ tokio = { version = "1.38.0", features = ["full"] }
 chrono = "0.4.39"
 clap = { version = "4.5.3", features = ["derive"] }
 once_cell = "1.20.2"
-tycho-common = ">0.81.5"
+tycho-common = ">=0.81.6"
+
```
dianacarvalho1, line 41: unnecessary blank line

foundry/test/assets/calldata.txt
```diff
@@ -38,3 +38,6 @@ test_single_encoding_strategy_bebop_aggregate:5c4b639c00000000000000000000000000
...
...000000002260fac5e5542a773aa44fbcfedf7c193bc2c599000000000000000000000000000000000000000000000000000000000038aebf00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc200000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000015b15cf58144ef33af1e14b5208015d11f9143e27b90001478eca1b93865dca0b9f325935eb123c8a4af011bee3211ab312a8d065c4fef0247448e17a8da000cd09f75e2bf2a4d11f3ab23f1389fcc1621c0cc2a0b86991c6218b36c1d19d4a2e9eb0ce3606eb482260fac5e5542a773aa44fbcfedf7c193bc2c5990000000000000000000000000000000000000000000000000000000100c84f11000000000000000000000000000000000000000000000000000000000038aebf0000000000000000000000000000000000000000000000000000000068a47cd800000000000000000000000000000000000000000000000000000198c286fecb125000064000640000001747eb8c38ffffffffffffff0029642016edb36d00006ddb3b21fe8509e274ddf46c55209cdbf30360944abbca6569ed6b26740d052f419964dcb5a3bdb98b4ed1fb3642a2760b8312118599a962251f7a8f73fe4fbe1c0000000000
```
dianacarvalho1, line 42: please delete this entry now that you renamed the test!

dianacarvalho1 (approved):
veeery nice! 👏🏼 👏🏼  Thank you @adrianbenavides ! only have tiny comments now

adrianbenavides:
I addressed the last two comments at https://github.com/propeller-heads/tycho-execution/pull/255

PR #253, opened 2025-08-18
feat: implement SwapEncoder for Balancer V3 boosted pool

zach030:


src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -603,6 +604,23 @@ pub struct BalancerV3SwapEncoder {
...
+    None,
+    Wrapped,
+    Underlying,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+struct MappingToken {
+    pub addresses: Vec<Vec<u8>>,
+    pub token_type: TokenType,
+}
```
dianacarvalho1: can you give an example on how does this look like please? I don't understand how we have multiple addresses and only one token type

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -621,11 +639,81 @@ impl SwapEncoder for BalancerV3SwapEncoder {
...
+            }
+        };
+
+        // - if token_in exists in mapping_tokens && token_type == Underlying --> wrap_in token_in:
+        //   ERC20 --> pool: ERC4626
+        let (wrap_in, wrap_in_index) = matches(&swap.token_in, TokenType::Underlying);
+
+        // - if token_in exists in mapping_tokens && token_type == Wrapped --> unwrap_in token_in:
+        //   ERC4626 --> pool: ERC20
+        let (unwrap_in, _) = matches(&swap.token_in, TokenType::Wrapped);
```
dianacarvalho1: why? we might want to just trade the wrapped token version.. I think we only need the wrap_in and unwrap_out. Can you explain why we would need the other cases please?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -621,11 +639,81 @@ impl SwapEncoder for BalancerV3SwapEncoder {
...
+            get_wrapped_token(unwrap_out, wrap_out, unwrap_out_index, &swap.token_out);
+
         let args = (
             bytes_to_address(&swap.token_in)?,
             bytes_to_address(&swap.token_out)?,
             pool,
-            (encoding_context.transfer_type as u8).to_be_bytes(),
+            [packed_byte],
+            bytes_to_address(&wrapped_token_in)?,
+            bytes_to_address(&wrapped_token_out)?,
```
dianacarvalho1: I think this is not necessary.. we need only to pass the wrapped version of the tokens.. we already know if we need to wrap or unwrap based on the flags..

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: Does this give you mapping the between wrapped and underlying token?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -621,11 +639,81 @@ impl SwapEncoder for BalancerV3SwapEncoder {
             EncodingError::FatalError("Invalid pool address for Balancer v3".to_string())
         })?;
 
+        let mapping_tokens_bytes = get_static_attribute(swap, "mapping_tokens")?;
```
dianacarvalho1: this is not a mandatory attribute.. the pool might not have it and it shouldn't fail

foundry/src/executors/BalancerV3Executor.sol
```diff
@@ -60,28 +63,123 @@ contract BalancerV3Executor is IExecutor, RestrictTransferFrom, ICallback {
...
                 userData: ""
             })
         );
 
+        amountCalculated = swapAmountCalculated;
+
+        // todo need to delete?
+        if (amountIn == 0 && swapAmountIn > 0) {
+            amountIn = swapAmountIn;
+        }
```
dianacarvalho1: what is this for?

dianacarvalho1 (commented):
Thank you for this @zach030 🙏🏼 I have a few questions/suggestions!

src/encoding/evm/swap_encoder/swap_encoders.rs
zach030: > can you give an example on how does this look like please? I don't understand how we have multiple addresses and only one token type

In the liquidityBuffer design, each wrapped_token maps to one underlying_token, but one underlying_token can have multiple wrapped_tokens.

For example:

```rust
// Example 1: waEthUSDT → USDT
let waEthUSDT_mapping = MappingToken {
    token_type: TokenType::Underlying,
    addresses: vec![hex::decode("dAC17F958D2ee523a2206206994597C13D831ec7").unwrap()],
};

// Example 2: USDC → multiple wrapped tokens
let usdc_mapping = MappingToken {
    token_type: TokenType::Wrapped,
    addresses: vec![
        hex::decode("BEEF01735c132Ada46AA9aA4c54623cAA92A64CB").unwrap(), // steakUSDC
        hex::decode("D4fa2D31b7968E448877f69A96DE69f5de8cD23E").unwrap(), // waEthUSDC
    ],
};

```

So basically, wrapped → underlying is one-to-one, but underlying → wrapped can be one-to-many.

zach030: > Does this give you mapping the between wrapped and underlying token?

Yes, the length of pc.tokens matches the length of the mappingTokens array, and they correspond one-to-one. You can use each mappingToken to get the corresponding underlying or wrapped token.

src/encoding/evm/swap_encoder/swap_encoders.rs
zach030: Yes! The main cases are `wrap_in` and `unwrap_out`. I think `unwrap_in` and `wrap_out` could also be handled here without any extra cost. If they’re not really needed, I can just remove them.

dianacarvalho1: yes I think you can just remove them!

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: aahh I see! Thank you for explaining! 

> // Example 2: USDC → multiple wrapped tokens

However, does this happen inside of the same pool? this info is pool specific no? I'm just wondering if this data structure is not a bit of an overkill 😬

src/encoding/evm/swap_encoder/swap_encoders.rs
zach030: For both calls ([pre-swap ](https://github.com/propeller-heads/tycho-execution/blob/ba841477907bd00f3d6bca68b1a06ee8688bb6a8/foundry/src/executors/BalancerV3Executor.sol#L82-L115)& [post-swap](https://github.com/propeller-heads/tycho-execution/blob/ba841477907bd00f3d6bca68b1a06ee8688bb6a8/foundry/src/executors/BalancerV3Executor.sol#L147-L176)) to `erc4626BufferWrapOrUnwrap`, we need to pass the corresponding `wrappedToken`.
- For `unwrap_in`, the `wrappedToken` is `swap.token_in`.
- For `wrap_in`, we need to know the corresponding `wrappedToken`.
Similarly, for `wrap_out` and `unwrap_out`, we also need to know the corresponding `wrappedToken`.

For example, if the pool has  `(wa-USDT, wa-USDC)`, and the user specifies `swap_in` as `USDT` and `swap_out` as `USDC`, we can determine from the pool’s `mappingTokens` attribute that we need `wrap_in` and `unwrap_out`. When calling `erc4626BufferWrapOrUnwrap` in the Executor contract, we have to specify the wrapped tokens (`wa-USDT` and `wa-USDC`) for each call.

These wrapped tokens can only be passed as params off-chain, the Executor contract itself cannot know which wrapped token corresponds to which underlying token.

src/encoding/evm/swap_encoder/swap_encoders.rs
zach030: Yes, this one’s optional.

foundry/src/executors/BalancerV3Executor.sol
zach030: I’ll check again, I don’t think it’s needed.

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: but as it is, if it's not present it will crash the program

src/encoding/evm/swap_encoder/swap_encoders.rs
dianacarvalho1: yes, but we don't need to know the correspondence do we? We just need to know the wrapped token addresses. So we could always force the token in and token out to be the wrapped tokens

dianacarvalho1:
PR closed due to lack of activity

PR #254, opened 2025-08-20
feat: Use Arc instead of & for protocol_state in Swap

tamaralipows:
- dealing with lifetimes makes refactoring painful for our users.


dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 can you update the quickstarts as well please?

PR #268, opened 2025-09-22
feat: Add historical_trade option to encoding 

dianacarvalho1:
Adding it is necessary because sometimes we use the encoding crate for historical trades for testing. This is relevant for protocols that need token approvals like Balancer v2 and Curve. For this case, we prefer to set the approval flag to always be true than checking if it's necessary using and RPC. This RPC check would be wrong because it always uses the latest block


tamaralipows (approved):
thx!

PR #274, opened 2025-10-07
fix: Use PLE in EkuboExecutor

tamaralipows:
Our encoding uses PLE for subsequent swaps in a swap group. The EkuboExecutor was assuming hard-coded hop length, which did not match the encoding side - leading to corrupted calldata in any swap after the first swap.


tamaralipows:
We have decided to go with the purely encoding-side solution for now. #275

PR #275, opened 2025-10-07
fix: Don't PLE-encode for Ekubo

tamaralipows:
Our encoding uses PLE for subsequent swaps in a swap group. The EkuboExecutor assumes hard-coded hop length, which did not match the encoding side - leading to corrupted calldata in any swap after the first swap.


src/encoding/evm/constants.rs
```diff
@@ -53,3 +53,11 @@ pub static CALLBACK_CONSTRAINED_PROTOCOLS: LazyLock<HashSet<&'static str>> = Laz
     set.insert("vm:balancer_v3");
     set
 });
+
+/// These groupable protocols use simple concatenation when forming swap groups instead of PLE
+/// encoding for grouped protocol data.
+pub static NON_PLE_ENCODED_PROTOCOLS: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
```
dianacarvalho1, line 58: 🫠

tests/protocol_integration_tests.rs
```diff
@@ -406,6 +406,96 @@ fn test_single_encoding_strategy_usv4_grouped_swap() {
...
+        &solution,
+        &UserTransferType::TransferFrom,
+        &eth(),
+        None,
+    )
+    .unwrap()
+    .data;
+
+    let hex_calldata = encode(&calldata);
+    write_calldata_to_file("test_single_ekubo_multi_hop", hex_calldata.as_str());
```
dianacarvalho1: can we say grouped swap instead of multi hop please?

dianacarvalho1 (approved):
Thank you so much @tamaralipows 🙏🏼 smol smol comment

PR #277, opened 2025-10-08
refactor: use Bytes instead of String for addresses in encoders

adrianbenavides:
Store addresses used in encoders as Bytes instead of as Strings.


dianacarvalho1 (approved):
Very nice! Thank you! 🙏🏼

PR #279, opened 2025-10-09
feat: Support Euler low balance swaps (univ4) 

dianacarvalho1:
Also fixed a biiiig 🐛 
Hook and hook data are per pool! not per swap. This was a big bug that was blocking a grouped swap in uniswap v4 with hooks. Updated execution and encoding to match the fix.

(review by commit for a tidier experience)


foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -163,42 +161,71 @@ contract UniswapV4Executor is
...
 
-        bytes calldata remaining = data[82:];
+        bytes calldata remaining = data[62:];
+
+        // Decode first pool with hook data
+        if (remaining.length < 48) {
+            // 20 + 3 + 3 + 20 + 2 = 48 minimum
+            revert UniswapV4Executor__InvalidDataLength();
+        }
+
```
tamaralipows, line 172: Shouldn't we do this check higher up?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -189,23 +189,24 @@ impl SwapEncoder for UniswapV4SwapEncoder {
...
-            hook_data = AlloyBytes::from(
-                swap.user_data
-                    .clone()
-                    .unwrap_or_default()
-                    .to_vec(),
-            );
-        }
+
+        let hook_data = swap
+            .user_data
```
tamaralipows, line 194: We are going to put hook info in user data? Weren't we just using this for the rfqs as a special case? Why not static attributes like the hook address?  Or are we repurposing this user data field now lol

dianacarvalho1: yes, this has been there since this was added (I think you were on vacation). We can't use the static attributes because the hook data might be different at each swap

foundry/src/executors/UniswapV4Executor.sol
dianacarvalho1: there still is a higher check like before. right in the beginning of this method

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: ok ok

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: in that case this is totally redundant no?

dianacarvalho1: not necessarily. The higher one just checks if the length is bigger than 1 swap. this check is checking if then each subsequent swap is of the right size too

tamaralipows: Oh I missed that you also added it below. AG

tamaralipows (approved):
tenk yu

PR #282, opened 2025-10-21
test: Sequential Euler swap with low balance

tamaralipows:
- We want to check pools using RLUSD because the pool manager doesn't hold much of this token, so we can't rely on pre-existing balance before the user makes a swap. We want to test a sequential swap here to check our grouping logic still works on Euler pools which store the balances on a proxy contract.


dianacarvalho1 (approved):
Amazing! Thank you so much for looking into this 🙏🏼

PR #285, opened 2025-11-07
feat: Implement FluidV1 encoder & executor

kayibal:


foundry/src/executors/FluidV1Executor.sol
```diff
@@ -0,0 +1,127 @@
...
+        address to_
+    ) external payable returns (uint256 amountOut_);
+}
+
+error FluidV1Executor__InvalidDataLength();
+error FluidV1Executor__InvalidCallback();
+
+contract FluidV1Executor is IExecutor, ICallback, RestrictTransferFrom {
+    // keccak(FluidV1Executor#CURRENT_SWAP_PARAMS)
+    // stores current dex address [0:20] and requested transfer type [31]
```
dianacarvalho1, line 30:
```suggestion
    // stores current dex address [0:20] and requested transfer type [21]
```

foundry/src/executors/FluidV1Executor.sol
```diff
@@ -0,0 +1,127 @@
...
+        }
+        dex = address(bytes20(value));
+    }
+
+    function _getTransferType() internal view returns (TransferType) {
+        uint256 value;
+        assembly {
+            value := tload(_CURRENT_SWAP_PARAMS_SLOT)
+        }
+        return TransferType(uint8(value & uint256(type(uint8).max)));
```
dianacarvalho1, line 101: this reads the lowest byte right? it's a bit obscure 😬

foundry/src/executors/FluidV1Executor.sol
```diff
@@ -0,0 +1,127 @@
...
+        (dex, zero2one, receiver, transferType) = _decodeData(data);
+
+        _setSwapParams(dex, transferType);
+
+        calculatedAmount =
+            dex.swapInWithCallback(zero2one, givenAmount, 0, receiver);
+    }
+
+    // Stores swap parameter packed into transient storage
+    function _setSwapParams(IFluidV1Dex dex, TransferType transferType)
```
dianacarvalho1, line 75: very nice! They don't support passing this in the calldata? Or did you decide to do it like this to be more gas efficient?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1008,6 +1008,55 @@ impl SwapEncoder for HashflowSwapEncoder {
...
+        &self,
+        swap: &Swap,
+        encoding_context: &EncodingContext,
+    ) -> Result<Vec<u8>, EncodingError> {
+        let component_id = AlloyBytes::from_str(&swap.component.id).map_err(|_| {
+            EncodingError::FatalError(format!(
+                "Failed parsing FluidV1 component id as bytes: {}",
+                &swap.component.id
+            ))
+        })?;
```
dianacarvalho1: you can convert from str to Address right away, no need to go through Bytes

dianacarvalho1 (commented):
Thank you @kayibal ! 🙏🏼 Overall it looks good! Left some comments (most for my understanding). Can you add a sequential test as well please? Particularly when Fluid is in the first hop. I have a feeling this will fail and that this is a [callback constrained ](https://github.com/propeller-heads/tycho-execution/blob/main/src/encoding/evm/constants.rs#L46)protocol..

foundry/src/executors/FluidV1Executor.sol
kayibal: What I wrote here is actually correct

foundry/src/executors/FluidV1Executor.sol
kayibal: You mean the fact that transferType is stored in the lowest byte or the way of extracting it?

foundry/src/executors/FluidV1Executor.sol
kayibal: Their examples don't do any callback verification at all, they simple transfer the requested amount to the liquidity contract.

foundry/src/executors/FluidV1Executor.sol
dianacarvalho1: can you explain why please?

foundry/src/executors/FluidV1Executor.sol
dianacarvalho1: the way of extracting

dianacarvalho1 (approved):
Niceeeee ✨ Thank you @kayibal ! 🙏🏼 This looks good to me!

foundry/test/Constants.sol
```diff
@@ -13,7 +13,8 @@ contract BaseConstants {
 
 contract Constants is Test, BaseConstants {
     address ADMIN = makeAddr("admin"); //admin=us
-    address BOB = makeAddr("bob"); //bob=someone!=us
+    // Changed this since someone deployed a drainer contract to makeAddr("bob")
+    address BOB = makeAddr("bob*"); //bob=someone!=us
```
dianacarvalho1, line 24: oh I think this is breaking a lot of the integration tests now 😬

src/encoding/evm/constants.rs
```diff
@@ -1,10 +1,13 @@
 use std::{collections::HashSet, sync::LazyLock};
 
+use alloy::hex;
+
 pub const DEFAULT_EXECUTORS_JSON: &str = include_str!("../../../config/executor_addresses.json");
 pub const DEFAULT_ROUTERS_JSON: &str = include_str!("../../../config/router_addresses.json");
 pub const PROTOCOL_SPECIFIC_CONFIG: &str =
     include_str!("../../../config/protocol_specific_addresses.json");
 
+pub const ETH_ADDRESS: &[u8] = &hex!("0x0000000000000000000000000000000000000000");
```
dianacarvalho1: you don't need to do this.. the native address is available through `Chain`. And you have this when creating your `SwapEncoder`

dianacarvalho1 (commented):
Thank you @kayibal ! Only have a small comment on Fluid. But the Bob address change broke a lot of tests 😬 Can you please fix this?

PR #286, opened 2025-11-11
feat: add Slipstreams executor and Base chain support

zach030:


config/test_executor_addresses.json
```diff
@@ -15,5 +15,8 @@
     "rfq:bebop": "0xD6BbDE9174b1CdAa358d2Cf4D57D1a9F7178FBfF",
     "rfq:hashflow": "0x15cF58144EF33af1e14b5208015d11F9143E27b9",
     "fluid_v1": "0x212224D2F2d262cd093eE13240ca4873fcCBbA3C"
+  },
+  "base": {
+    "aerodrome_slipstreams": "0x212224D2F2d262cd093eE13240ca4873fcCBbA3C"
```
zizou0x: I think we need to deploy before merging this, just waiting for @dianacarvalho1 input here and I'll deploy and update this

dianacarvalho1: yes, @zizou0x is part right! For the test addresses you should use the one that you get from the tycho router test..

tests/common/mod.rs
```diff
@@ -81,6 +81,19 @@ pub fn get_tycho_router_encoder(user_transfer_type: UserTransferType) -> Box<dyn
         .expect("Failed to build encoder")
 }
 
+pub fn get_base_tycho_router_encoder(
```
dianacarvalho1, line 84: 🙏🏼

foundry/src/executors/SlipstreamsExecutor.sol
```diff
@@ -0,0 +1,206 @@
...
+
+    function uniswapV3SwapCallback(
+        int256, /* amount0Delta */
+        int256, /* amount1Delta */
+        bytes calldata /* data */
+    )
+        external
+    {
+        handleCallback(msg.data);
+    }
```
dianacarvalho1, line 133: is this needed?

foundry/test/protocols/Slipstreams.t.sol
```diff
@@ -0,0 +1,257 @@
...
+        assertEq(receiver, address(2));
+        assertEq(target, address(3));
+        assertEq(zeroForOne, false);
+        assertEq(
+            uint8(transferType),
+            uint8(RestrictTransferFrom.TransferType.Transfer)
+        );
+    }
+
+    function testSwapIntegration() public {
```
dianacarvalho1: this is just testSwap. not integration

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -232,4 +242,24 @@ contract TychoRouterTestSetup is Constants, Permit2TestHelper, TestUtils {
             transferType
         );
     }
+
+    function encodeSlipstreamsSwap(
```
dianacarvalho1, line 249: I think this is unused..

dianacarvalho1 (commented):
Thank you @zach030 ! It looks good to me! I have just one suggestion.. in the executor, we are decoding the calldata and then encoding it again for the callback calldata. Do we need to _verifyPairAddress twice? once in the swap method and once in the callback? I would say that only in the callback this would be enough and then we don't need to decode to encode to decode..  
Also thank you for making the setup for Base 🙏🏼 much appreciated!

zach030:
> Thank you @zach030 ! It looks good to me! I have just one suggestion.. in the executor, we are decoding the calldata and then encoding it again for the callback calldata. Do we need to _verifyPairAddress twice? once in the swap method and once in the callback? I would say that only in the callback this would be enough and then we don't need to decode to encode to decode.. Also thank you for making the setup for Base 🙏🏼 much appreciated!

Hi @dianacarvalho1 ! I think `_verifyPairAddress` before the swap still makes sense, otherwise we call `pool.swap` directly and might hit unexpected reverts.
Also, this code is mostly from the `UniswapV3Executor` — I just changed a few swapData fields and the verifyPairAddress part. Should we stay aligned with UniswapV3, or is it fine to simplify it?

foundry/src/executors/SlipstreamsExecutor.sol
zach030: Yes, it’s needed. It follows the same callback as Uniswap V3.

foundry/test/TychoRouterTestSetup.sol
zach030: It gets called by [testSwap](https://github.com/propeller-heads/tycho-execution/blob/fb8817b92d93ab743ae6235880797658f96abc65/foundry/test/protocols/Slipstreams.t.sol#L104-L111).

foundry/src/executors/SlipstreamsExecutor.sol
dianacarvalho1: aaah. ok this is a uniswap v3 copycat?? I didn't know that

foundry/test/TychoRouterTestSetup.sol
dianacarvalho1: but this exact same function is defined there as well? 🤔

dianacarvalho1:
> > Thank you @zach030 ! It looks good to me! I have just one suggestion.. in the executor, we are decoding the calldata and then encoding it again for the callback calldata. Do we need to _verifyPairAddress twice? once in the swap method and once in the callback? I would say that only in the callback this would be enough and then we don't need to decode to encode to decode.. Also thank you for making the setup for Base 🙏🏼 much appreciated!
> 
> Hi @dianacarvalho1 ! I think `_verifyPairAddress` before the swap still makes sense, otherwise we call `pool.swap` directly and might hit unexpected reverts. Also, this code is mostly from the `UniswapV3Executor` — I just changed a few swapData fields and the verifyPairAddress part. Should we stay aligned with UniswapV3, or is it fine to simplify it?

huuumm ok. No, it's fine to simplify. But then if you want to keep the `_verifyPairAddress` before the swap can we try to get rid of the second encoding for the callback? we could just put the params we need next to each other and slice the bytes I think..

dianacarvalho1 (approved):
Thank you @zach030 ! 🙏🏼 This looks good to me! Let's deploy it!

PR #287, opened 2025-11-11
feat: add Pancakeswap V3 support for Base

zizou0x:
This PR adds support for PancakeSwapV3 on Base

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ target/
 # Compiler files
 foundry/cache/
 foundry/out/
+
+node_modules/
--- a/foundry/src/executors/UniswapV3Executor.sol
+++ b/foundry/src/executors/UniswapV3Executor.sol
@@ -121,7 +121,9 @@ contract UniswapV3Executor is IExecutor, ICallback, RestrictTransferFrom {
         int256, /* amount0Delta */
         int256, /* amount1Delta */
         bytes calldata /* data */
-    ) external {
+    )
+        external
+    {
         handleCallback(msg.data);
     }
 
@@ -129,7 +131,9 @@ contract UniswapV3Executor is IExecutor, ICallback, RestrictTransferFrom {
         int256, /* amount0Delta */
         int256, /* amount1Delta */
         bytes calldata /* data */
-    ) external {
+    )
+        external
+    {
         handleCallback(msg.data);
     }
 
--- a/foundry/src/executors/UniswapV3Executor.sol
+++ b/foundry/src/executors/UniswapV3Executor.sol
@@ -121,9 +121,15 @@ contract UniswapV3Executor is IExecutor, ICallback, RestrictTransferFrom {
         int256, /* amount0Delta */
         int256, /* amount1Delta */
         bytes calldata /* data */
-    )
-        external
-    {
+    ) external {
+        handleCallback(msg.data);
+    }
+
+    function pancakeV3SwapCallback(
+        int256, /* amount0Delta */
+        int256, /* amount1Delta */
+        bytes calldata /* data */
+    ) external {
         handleCallback(msg.data);
     }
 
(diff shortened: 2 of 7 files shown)


foundry/src/executors/UniswapV3Executor.sol
```diff
@@ -121,9 +121,15 @@ contract UniswapV3Executor is IExecutor, ICallback, RestrictTransferFrom {
...
         int256, /* amount1Delta */
         bytes calldata /* data */
-    )
-        external
-    {
+    ) external {
+        handleCallback(msg.data);
+    }
+
+    function pancakeV3SwapCallback(
```
zizou0x, line 130: Could this be a problem? AFAIK it's not strictly needed to work with the TychoRouter because it calls `handleCallBack()` directly but it's needed to make the executor work without going through the router (do we need this?)

foundry/test/protocols/UniswapV3.t.sol
```diff
@@ -266,4 +266,36 @@ contract TychoRouterForBalancerV3Test is TychoRouterTestSetup {
 
         vm.stopPrank();
     }
+
+    // Base Network Test
+    // Make sure to set the RPC_URL to base network
+    function testSwapPancakeswapBaseNetwork() public {
```
zizou0x, line 272: I'm not convinced this is the best setup for multi-chain testing but I took this from [here](https://github.com/propeller-heads/tycho-execution/blob/4b69a0b6bc88136a3012c5917d332c0a314f60df/foundry/test/protocols/UniswapV2.t.sol#L245)

dianacarvalho1: I think this is ok for now

foundry/src/executors/UniswapV3Executor.sol
dianacarvalho1: ah yeah univ3 is weird in this aspect.. we did the same for univ3 - to allow executor direct transfers (which is not really a feature we use or advertise). 
But I think for pancakeswap v3 in mainnet we didn't even add it 😅 so I would say it's fine either way.. as you prefer

.gitignore
```diff
@@ -31,3 +31,5 @@ target/
 # Compiler files
 foundry/cache/
 foundry/out/
+
+node_modules/
```
dianacarvalho1, line 35: what is this?

foundry/test/protocols/Hashflow.t.sol
```diff
@@ -31,9 +31,11 @@ contract HashflowUtils is Test {
         );
     }
 
-    function encodeRfqtQuoteWithDefaults(
-        IHashflowRouter.RFQTQuote memory quote
-    ) internal pure returns (bytes memory) {
+    function encodeRfqtQuoteWithDefaults(IHashflowRouter.RFQTQuote memory quote)
```
dianacarvalho1, line 34: damn you foundry

dianacarvalho1 (commented):
Thank you @zizou0x 🙏🏼 This looks good to me! let's deploy before merging the PR though 🚀 You are going to have to add the protocol in `config/executor_addresses.json`

.gitignore
zizou0x: It's the folder that node.js uses to store dependencies, see [here](https://www.reddit.com/r/node/comments/9o5fzk/comment/e7ribji/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

I'm surprised you never got issues with this being if the git changes. This is basically the folder that is created when you run `npm install`

foundry/src/executors/UniswapV3Executor.sol
zizou0x: I also think it's fine given that `handleCallback` is public and this is just another interface for this same function

.gitignore
dianacarvalho1: huumm weird.. never had issues 🤷🏼‍♀️

dianacarvalho1 (approved):
✨

PR #292, opened 2025-11-25
feat: Angstrom encoding + execution

tamaralipows:
I tried to make it nice for you to review by including the OG `UniswapV4Executor` in the first commit but github doesn't recognize my kindness (although my rust rover did) :( Sorry for the bad diff parsing.


foundry/src/executors/UniswapV4Executor.sol
```diff
@@ -146,9 +147,11 @@ contract UniswapV4Executor is
         return amountOut;
     }
 
+    // slither-disable-next-line dead-code
```
dianacarvalho1, line 150: why is this dead code?

foundry/src/executors/UniswapV4AngstromExecutor.sol
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        for (uint256 i = 0; i < attestations.length; i++) {
+            // slither-disable-next-line incorrect-equality
+            if (attestations[i].blockNumber == block.number) {
+                return attestations[i].attestation;
+            }
+        }
+
+        revert UniswapV4AngstromExecutor__NoValidAttestation(block.number);
```
dianacarvalho1: this is more no attestation found at all for block X. not necessarily invalid

foundry/src/executors/UniswapV4AngstromExecutor.sol
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// @notice Decodes attestation data
+    /// @dev Each attestation is exactly 93 bytes: 8 bytes blockNumber + 85 bytes attestation
+    /// @param attestationData Raw bytes containing attestations
+    /// @return attestations Array of AttestationData structs
+    function _decodeAttestations(bytes memory attestationData)
```
dianacarvalho1: I was thinking if it would be worth it to do something smarter, like decode block by block only if we don't find the correct one. But tbh I'm not even sure this would be gas worth it..

foundry/test/protocols/UniswapV4Angstrom.t.sol
```diff
@@ -0,0 +1,228 @@
...
+        deal(USDC_ADDR, address(angstromExecutor), amountIn);
+        uint256 poolManagerBalanceBefore = USDC.balanceOf(POOL_MANAGER);
+
+        // Create attestations for multiple blocks
+        // The attestation is real return data from the Angstrom attestation API
+        bytes memory attestation =
+            hex"d437f3372f3add2c2bc3245e6bd6f9c202e61bb324e10ae0affbd0fb7b3622098b00d81ee679dd1adb41c03dda9b50565bddaead205a7c383603a49b4fc576c870d6fac726a007ddc008077f5f942a177cedf3ca1c";
+
+        uint256 currentBlock = block.number;
+        console.logUint(currentBlock);
```
dianacarvalho1: upsi

foundry/test/protocols/UniswapV4Angstrom.t.sol
```diff
@@ -0,0 +1,228 @@
...
+                UniswapV4AngstromExecutor__NoValidAttestation.selector,
+                currentBlock
+            )
+        );
+        angstromExecutor.swap(amountIn, data);
+    }
+
+    function testGroupedSwapIntegration() public {
+        // Load calldata generated by test_encode_angstrom_grouped_swap
+        // This data must only be updated alongside the fork block in the setup
```
dianacarvalho1: I don't understand this sentence.. Can you explain please?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -250,6 +251,244 @@ impl SwapEncoder for UniswapV4SwapEncoder {
     }
 }
 
+/// Attestation data for Angstrom swaps
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[allow(dead_code)]
```
dianacarvalho1: why is it dead? 🤔

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -250,6 +251,244 @@ impl SwapEncoder for UniswapV4SwapEncoder {
...
+            EncodingError::FatalError(format!("Failed to parse attestation response: {}", e))
+        })?;
+
+        if !attestation_response.success {
+            return Err(EncodingError::FatalError(
+                "Angstrom API returned success=false".to_string(),
+            ));
+        }
+
+        println!("{attestation_response:?}");
```
dianacarvalho1: upsi

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -250,6 +251,244 @@ impl SwapEncoder for UniswapV4SwapEncoder {
...
+    }
+}
+
+impl SwapEncoder for UniswapV4AngstromSwapEncoder {
+    fn new(
+        executor_address: Bytes,
+        _chain: Chain,
+        config: Option<HashMap<String, String>>,
+    ) -> Result<Self, EncodingError> {
+        let cfg = config.ok_or_else(|| {
```
dianacarvalho1: shouldn't we add stuff to `protocol_specific_addresses` then?? 🤔

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -250,6 +251,244 @@ impl SwapEncoder for UniswapV4SwapEncoder {
...
+            EncodingError::FatalError("Angstrom encoder requires config".to_string())
+        })?;
+        let api_key = cfg
+            .get("angstrom_api_key")
+            .cloned()
+            .ok_or_else(|| {
+                EncodingError::FatalError(
```
dianacarvalho1: huummm not sure if this wouldn't be better read from env..

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -250,6 +251,244 @@ impl SwapEncoder for UniswapV4SwapEncoder {
...
+
+            // Encode attestation data for block (next 85 bytes)
+            encoded.extend_from_slice(&attestation_bytes);
+        }
+
+        Ok(encoded)
+    }
+}
+
+impl SwapEncoder for UniswapV4AngstromSwapEncoder {
```
dianacarvalho1: huuummm I think this is not necessary 🤔 I think we should be using the normal Univ4 Swap Encoder and if the hook address is Angstrom then do the Angstrom specific things.. because right now we don't really have a way to use this new encoder..

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 this looks good! I have some suggestions

foundry/src/executors/UniswapV4Executor.sol
tamaralipows: see commit message.. it looks like a false positive.

foundry/src/executors/UniswapV4AngstromExecutor.sol
tamaralipows: hmmm when I say valid I mean exactly that - the attestation is valid if it has been assigned block X
I can rename it to `NoAttestationForBlock`

foundry/src/executors/UniswapV4AngstromExecutor.sol
tamaralipows: I was thinking about doing that way but it's less clear :( But I can measure gas actually and see if it's worth it!

foundry/test/protocols/UniswapV4Angstrom.t.sol
tamaralipows: If you update this data i.e. by running all the rust tests, you must also update the fork block in the setup. Else the fork block will be way older than the new data. The test is skipped by default because of the API connection so it should be fine I think

tamaralipows: I rephrased. let me know if it helps

foundry/src/executors/UniswapV4AngstromExecutor.sol
tamaralipows: The current way, including test setup so only look at the gas difference not the absolute gas:

[PASS] testGroupedSwapIntegration() (gas: 1103414)

The "smart" way:

[PASS] testGroupedSwapIntegration() (gas: 1097059)

So for 3 attestations we save 6.3k gas if the attestation is the second one.

Actually, that's quite a lot... maybe it's worth it... I'm gonna make a commit

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -152,21 +154,115 @@ impl SwapEncoder for UniswapV3SwapEncoder {
...
+    angstrom_hook_address: Bytes,
 }
 
 impl UniswapV4SwapEncoder {
     fn get_zero_to_one(sell_token_address: Address, buy_token_address: Address) -> bool {
         sell_token_address < buy_token_address
     }
+
+    /// Fetches attestations from the Angstrom API (blocking)
+    fn fetch_attestations() -> Result<AttestationResponse, EncodingError> {
```
dianacarvalho1:
```suggestion
    fn fetch_angstrom_attestations() -> Result<AttestationResponse, EncodingError> {
```

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -190,11 +286,18 @@ impl SwapEncoder for UniswapV4SwapEncoder {
             Err(_) => Address::ZERO,
         };
 
-        let hook_data = swap
-            .user_data
-            .clone()
-            .unwrap_or_default()
-            .to_vec();
+        let is_angstrom_hook = **hook_address == *self.angstrom_hook_address;
```
dianacarvalho1, line 291: removing the * in both doesn't work?

dianacarvalho1 (approved):
Awesomeeee ✨ This looks really good! 🙏🏼 thank you!

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: no I think because one is bytes and one is address :((  so we get
```
can't compare `alloy::alloy_primitives::FixedBytes<20>` with `tycho_common::Bytes`

PR #293, opened 2025-11-28
fix: Angstrom execution fixes

tamaralipows:


src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -123,7 +124,7 @@ impl SwapEncoder for UniswapV3SwapEncoder {
             .map_err(|_| EncodingError::FatalError("Invalid USV3 component id".to_string()))?;
         let pool_fee_bytes = get_static_attribute(swap, "fee")?;
 
-        let pool_fee_u24 = pad_to_fixed_size::<3>(&pool_fee_bytes)
+        let pool_fee_u24 = pad_or_truncate_to_size::<3>(&pool_fee_bytes)
```
dianacarvalho1, line 127: this it true for univ3 too?? 🤔

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1229,9 +1230,10 @@ impl SwapEncoder for SlipstreamsSwapEncoder {
         })?;
         let tick_spacing_bytes = get_static_attribute(swap, "tick_spacing")?;
 
-        let tick_spacing_bytes_u24 = pad_to_fixed_size::<3>(&tick_spacing_bytes).map_err(|_| {
-            EncodingError::FatalError("Failed to extract tick_spacing bytes".to_string())
-        })?;
+        let tick_spacing_bytes_u24 =
```
dianacarvalho1, line 1233: and slipstreams??

dianacarvalho1 (approved):
Niceeeee! Great job at figuring this out 💪🏼 I have only questions abou the scope of these changes

src/encoding/evm/swap_encoder/swap_encoders.rs
tamaralipows: Not sure what the question is. This doesn't break anything that already worked. It would just prevent errors

PR #296, opened 2025-12-02
feat: add ERC4626 executor

zach030:


tamaralipows (approved):
Thank you - looks good! Couldn't find any issues here 🙏

foundry/src/executors/ERC4626Executor.sol
```diff
@@ -0,0 +1,63 @@
...
+        (tokenIn, target, receiver, transferType) = _decodeData(data);
+        _transfer(address(this), transferType, address(tokenIn), givenAmount);
+
+        if (address(tokenIn) == target) {
+            // shares --> asset
+            calculatedAmount =
+                IERC4626(target).redeem(givenAmount, receiver, address(this));
+        } else if (address(tokenIn) == IERC4626(target).asset()) {
+            // asset --> shares
+            tokenIn.forceApprove(target, type(uint256).max);
```
dianacarvalho1: do we need to do this every time?

dianacarvalho1: Have a look at balancer v2 to see how we are approving only when it's needed!

dianacarvalho1 (commented):
Thank you @zach030 ! 🙏🏼 This looks very good! I only have one suggestion about the approval!

foundry/src/executors/ERC4626Executor.sol
zach030: Thanks! I’ve added another flag for the approval.

dianacarvalho1 (approved):
Thank you 🙏🏼 Looks really good now!

PR #298, opened 2025-12-04
fix(angstrom): Block in place getting attestations

dianacarvalho1:
The encode_swap function is synchronous, but it calls fetch_angstrom_attestations() which uses reqwest::blocking::Client. The blocking client creates a Tokio runtime internally, and when this function is called from within an async context (as it is in the integration test), the runtime gets dropped causing a panic.


tamaralipows (approved):
Thank you!

PR #299, opened 2025-12-04
fix: remove ``--all-features` flag from nextest command in check script to mirror the CI

ElusAegis:


dianacarvalho1 (approved):
Thank you!

PR #301, opened 2025-12-04
feat: add Lido execution

Troshchk:


foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+error LidoExecutor__InvalidTarget();
+error LidoExecutor__InvalidFactory();
+error LidoExecutor__InvalidInitCode();
+error LidoExecutor__InvalidFee();
+error LidoExecutor__InvalidSwapDirection();
+error LidoExecutor__ZeroAddress();
+
+interface LidoPool {
+    // slither-disable-next-line naming-convention
+    function submit(address _referral) external payable returns (uint256);
```
dianacarvalho1: is this because of the _? I think we can just remove it

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+        {
+            // stETH staking: ETH -> stETH
+            // stETH is a rebasing token where balances are calculated from shares
+            // Measure actual balance changes to account for rounding in share conversions
+            uint256 balanceBefore = IERC20(stETH).balanceOf(address(this));
+
+            // slither-disable-next-line arbitrary-send-eth
+            uint256 _shares =
+                LidoPool(stETH).submit{value: givenAmount}(receiver);
+            // _shares;
```
dianacarvalho1: unnecessary comment

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+
+                uint256 receiverBalanceAfter = IERC20(stETH).balanceOf(receiver);
+                // Update calculatedAmount to reflect actual tokens received after transfer
+                // (accounts for additional rounding during transfer)
+                calculatedAmount = receiverBalanceAfter - receiverBalanceBefore;
+            }
+        } else if (
+            pool == LidoPoolType.wstETH && direction == LidoPoolDirection.Wrap
+        ) {
+            //wsteth, steth -> wsteth
```
dianacarvalho1: can we keep these cases comments consistent between them please? form and capitalised tokens or not

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+        returns (
+            address receiver,
+            TransferType transferType,
+            LidoPoolType pool,
+            LidoPoolDirection direction,
+            bool approvalNeeded
+        )
+    {
+        if (data.length != 24) {
+            // TODO: double check the length
```
dianacarvalho1: todo left here

foundry/test/protocols/Lido.t.sol
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+contract LidoExecutorTest is Constants, Permit2TestHelper, TestUtils {
+    using SafeERC20 for IERC20;
+
+    LidoExecutorExposed LidoExposed;
+
+    function setUp() public {
+        uint256 forkBlock = 23934489; //change for a newer block
```
dianacarvalho1: todo comment left

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+        if (pool == LidoPoolType.stETH && direction == LidoPoolDirection.Stake)
+        {
+            // stETH staking: ETH -> stETH
+            // stETH is a rebasing token where balances are calculated from shares
+            // Measure actual balance changes to account for rounding in share conversions
+            uint256 balanceBefore = IERC20(stETH).balanceOf(address(this));
+
+            // slither-disable-next-line arbitrary-send-eth
+            uint256 _shares =
+                LidoPool(stETH).submit{value: givenAmount}(receiver);
```
dianacarvalho1: here the (receiver) is misleading. can we not pass anything here? or an empty address? this is actual for the referral and we don't have any

foundry/test/protocols/Lido.t.sol
```diff
@@ -0,0 +1,239 @@
...
+        bytes memory callData = loadCallDataFromFile(
+            "test_single_encoding_strategy_unwrap_wsteth_lido"
+        );
+        (bool success,) = tychoRouterAddr.call(callData);
+
+        assertTrue(success, "Call Failed");
+        assertEq(IERC20(WSTETH_ADDR).balanceOf(ALICE), 0);
+    }
+
+    function testSingleUsv4LidoIntegrationGroupedSwap() public {
```
dianacarvalho1: this is not single nor a grouped swap. this is a sequential swap

foundry/test/protocols/Lido.t.sol
```diff
@@ -0,0 +1,239 @@
...
+        );
+
+        (bool success,) = tychoRouterAddr.call(callData);
+
+        vm.stopPrank();
+
+        assertTrue(success, "Call Failed");
+    }
+
+    function testSingleCurveLidoIntegrationGroupedSwap() public {
```
dianacarvalho1: same here

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -133,6 +135,8 @@ contract TychoRouterTestSetup is Constants, Permit2TestHelper, TestUtils {
         address poolManagerAddress = 0x000000000004444c5dc75cB358380D2e3dE08A90;
         address ekuboCore = 0xe0e0e08A6A4b9Dc7bD67BCB7aadE5cF48157d444;
         address ekuboMevResist = 0x553a2EFc570c9e104942cEC6aC1c18118e54C091;
+        address steth = 0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84;
```
dianacarvalho1: we already have both of these in constants no?

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1365,6 +1365,102 @@ impl SwapEncoder for ERC4626SwapEncoder {
...
+}
+
+#[repr(u8)]
+enum LidoPoolDirection {
+    Stake = 0,
+    Wrap = 1,
+    Unwrap = 2,
+}
+
+impl LidoSwapEncoder {}
```
dianacarvalho1: unnecessary

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1365,6 +1365,102 @@ impl SwapEncoder for ERC4626SwapEncoder {
...
+impl SwapEncoder for LidoSwapEncoder {
+    fn new(
+        executor_address: Bytes,
+        chain: Chain,
+        _config: Option<HashMap<String, String>>,
+    ) -> Result<Self, EncodingError> {
+        Ok(Self {
+            executor_address,
+            st_eth_address: Bytes::from("0xae7ab96520de3a18e5e111b5eaab095312d7fe84"),
+            wst_eth_address: Bytes::from("0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0"),
```
dianacarvalho1: why are we hardcoding this? we can read this from the config

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -1365,6 +1365,102 @@ impl SwapEncoder for ERC4626SwapEncoder {
...
+
+            (LidoPool::WStETH, LidoPoolDirection::Wrap, approval_needed)
+        } else if swap.token_in == self.wst_eth_address && swap.token_out == self.st_eth_address {
+            (LidoPool::WStETH, LidoPoolDirection::Unwrap, false)
+        } else {
+            return Err(EncodingError::InvalidInput("Combination not allowed".to_owned()))
+        };
+
+        // Token in address is always needed to perform a manual transfer from the router,
+        // since no optimizations are performed that send from one pool to the next
```
dianacarvalho1: unnecessary comment

src/encoding/evm/swap_encoder/swap_encoders.rs
```diff
@@ -3018,4 +3114,181 @@ mod tests {
...
+                    "02",
+                    // pool
+                    "01",
+                    // direction
+                    "02",
+                    // approval_needed
+                    "00",
+                ))
+            );
+            write_calldata_to_file("test_encode_lido", hex_swap.as_str());
```
dianacarvalho1: this isn't being used anywhere I think

tests/protocol_integration_tests.rs
```diff
@@ -1481,3 +1481,346 @@ fn test_sequential_encoding_strategy_erc4626() {
...
+
+    let hex_calldata = encode(&calldata);
+    write_calldata_to_file(
+        "test_single_encoding_strategy_unwrap_wsteth_lido",
+        hex_calldata.as_str(),
+    );
+}
+
+#[test]
+fn test_single_encoding_strategy_usv4_lido_grouped_swap() {
```
dianacarvalho1: same thing here. this is sequential not grouped swap nor single

tests/protocol_integration_tests.rs
```diff
@@ -1481,3 +1481,346 @@ fn test_sequential_encoding_strategy_erc4626() {
...
+    let hex_calldata = encode(&calldata);
+
+    write_calldata_to_file(
+        "test_single_encoding_strategy_usv4_lido_grouped_swap",
+        hex_calldata.as_str(),
+    );
+}
+
+#[test]
+fn test_single_encoding_strategy_curve_lido_grouped_swap() {
```
dianacarvalho1: same here

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 Logic wise I think we are there ✨ Form wise I have some comments and suggestions 🧹

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+            revert LidoExecutor__ZeroAddress();
+        }
+        stETH = _stETH_address;
+        wstETH = _wstETH_address;
+    }
+
+    // slither-disable-next-line locked-ether
+    function swap(
+        uint256 givenAmount,
+        bytes calldata data //abi packed encoded
```
tamaralipows: Comment unnecessary

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+
+            uint256 balanceAfter = IERC20(stETH).balanceOf(address(this));
+            calculatedAmount = balanceAfter - balanceBefore;
+
+            // submit() sends stETH to this contract, transfer to receiver if needed
+            if (receiver != address(this)) {
+                uint256 receiverBalanceBefore =
+                    IERC20(stETH).balanceOf(receiver);
+
+                IERC20(stETH).safeTransfer(receiver, calculatedAmount);
```
tamaralipows: Since you're constantly casting this to `IERC20` I'm curious if it would be better to just initialize `stETH` as `IERC20` instead of `address`

foundry/src/executors/LidoExecutor.sol
tamaralipows: What does the `wsteth` before the comma mean? Isn't `steth -> wsteth` sufficient to explain the situation?

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,158 @@
...
+                calculatedAmount = receiverBalanceAfter - receiverBalanceBefore;
+            }
+        } else if (
+            pool == LidoPoolType.wstETH && direction == LidoPoolDirection.Wrap
+        ) {
+            //wsteth, steth -> wsteth
+            _transfer(address(this), transferType, stETH, givenAmount);
+
+            if (approvalNeeded) {
+                IERC20(stETH).forceApprove(wstETH, type(uint256).max - 1);
```
tamaralipows: I'm curious why we do max - 1? I probably missed it somewhere

foundry/test/protocols/Lido.t.sol
```diff
@@ -0,0 +1,239 @@
...
+
+        vm.startPrank(ALICE);
+
+        bytes memory callData =
+            loadCallDataFromFile("test_single_encoding_strategy_steth_lido");
+
+        (bool success,) = tychoRouterAddr.call{value: 1 ether}(callData);
+
+        assertTrue(success, "Call Failed");
+        assertEq(ALICE.balance, 0);
```
tamaralipows, line 169: Should we also check the token that we are receiving here?

foundry/test/protocols/Lido.t.sol
```diff
@@ -0,0 +1,239 @@
...
+        LidoPool(STETH_ADDR).submit{value: 1 ether}(address(ALICE));
+
+        IERC20(STETH_ADDR).approve(tychoRouterAddr, type(uint256).max - 1);
+        bytes memory callData = loadCallDataFromFile(
+            "test_single_encoding_strategy_wrap_wsteth_lido"
+        );
+        (bool success,) = tychoRouterAddr.call(callData);
+
+        assertTrue(success, "Call Failed");
+        assertEq(IERC20(STETH_ADDR).balanceOf(ALICE), 0);
```
tamaralipows, line 186: And here?

foundry/test/protocols/Lido.t.sol
tamaralipows: FYI the grouped swaps only refer to `UniswapV4` and `Ekubo` (protocols that do flash accounting). More info here: https://docs.propellerheads.xyz/tycho/for-solvers/execution/encoding#swap-group

tamaralipows (commented):
Thank youuu! A few more comments

foundry/src/executors/LidoExecutor.sol
Troshchk: [Lido does not allow to for allowances to be set to max](https://www.codeslaw.app/contracts/ethereum/0x17144556fd3424edc8fc8a4c940b2d04936d17eb?file=contracts%2F0.4.24%2FStETH.sol&start=411), so we have decided to the maximal allowed number

foundry/test/protocols/Lido.t.sol
Troshchk: I assumed that since `checked_amount` already exists in the `Solution` and we are validating against it, there was no need to add an additional check for the token we are receiving. Do you think it would be better to add that check here anyway?

foundry/src/executors/LidoExecutor.sol
Troshchk: Good point, I still need stETH address in several places, so I have added one more variable instead!

foundry/test/protocols/Lido.t.sol
tamaralipows: I think it would be better to add it twice - since it would be a double-confirmation that our check in the contract is working properly with our encoding (checking the right token, properly encoded, sending to the right address)

Troshchk: Makes sense—I just added the tests.
In fact, I found a value I needed to change using that test, so it was already helpful!

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,152 @@
...
+error LidoExecutor__InvalidFee();
+error LidoExecutor__InvalidSwapDirection();
+error LidoExecutor__ZeroAddress();
+
+interface LidoPool {
+    function submit(address referral) external payable returns (uint256);
+}
+
+interface LidoWrappedPool {
+    // slither-disable-next-line naming-convention
```
dianacarvalho1: is this still needed? I don't really know tbh

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    constructor(
+        address _stETH_address,
+        address _wstETH_address,
+        address _permit2
+    ) RestrictTransferFrom(_permit2) {
+        if (_stETH_address == address(0) || _wstETH_address == address(0)) {
+            revert LidoExecutor__ZeroAddress();
+        }
+        stETHIERC20 = IERC20(_stETH_address);
```
dianacarvalho1: I would prefer to call this `stETH` and the other `stETHAddress`

foundry/src/executors/LidoExecutor.sol
```diff
@@ -0,0 +1,152 @@
...
+    IERC20 public immutable stETHIERC20;
+    address public immutable stETH;
+    address public immutable wstETH;
+
+    constructor(
+        address _stETH_address,
+        address _wstETH_address,
+        address _permit2
+    ) RestrictTransferFrom(_permit2) {
+        if (_stETH_address == address(0) || _wstETH_address == address(0)) {
```
dianacarvalho1: note the [solidity conventions](https://docs.soliditylang.org/en/latest/style-guide.html#naming-styles) on naming variables. the should be mixedCase and not snake_case

dianacarvalho1 (approved):
Thank you @Troshchk ! 🙏🏼 only have tiny tiny tiny comments but otherwise it looks goooddd ✅

PR #302, opened 2025-12-04
feat: (Angstrom) Don't revert in executor if attestation not found

tamaralipows:
- Just send empty data
- The revert will now happen in the Angstrom contract instead of on our side
- This makes it easier to overwrite the Angstrom behaviour for simulation purposes


dianacarvalho1 (approved):
amazing ✨

PR #304, opened 2025-12-05
feat: rocketpool integration

ElusAegis:
Two points I still have:
1. I currently test the full integration path only for the `deposit` case, leaving the `withdraw` untested. This is because I gathered we should only have a single calldata test case in `calldata.txt`. Is this a right assumption? Or should I add several calldatas to test to for a single protocol to `calldata.txt`.
2. I am still not sure how the native ETH exchange works. Notice that if the transfer type is `TransferFrom`, we assume that the router/executor should withdraw the target token from the usre balance for the exchange. However, for ETH, this is impossible, and it should be transfered with the function call, breaking the assumption of the  `TransferFrom`. Thus it feels like `TransferFrom` should not be supported for the deposit. However, I have also noticed that `FluidV1` also works with native ETH and does none of such checks, so I am a bit unclear about if this is needed or not.


foundry/src/executors/RocketpoolExecutor.sol
```diff
@@ -0,0 +1,86 @@
...
+
+error RocketpoolExecutor__InvalidDataLength();
+
+contract RocketpoolExecutor is IExecutor, RestrictTransferFrom {
+    using SafeERC20 for IRocketTokenRETH;
+
+    IRocketTokenRETH public constant RETH =
+        IRocketTokenRETH(0xae78736Cd615f374D3085123A210448E74Fc6393);
+    IRocketDepositPool public constant ROCKET_DEPOSIT_POOL =
+        IRocketDepositPool(0xDD3f50F8A6CafbE9b31a427582963f465E745AF8);
```
tamaralipows, line 24: Do you know if we have plans to deploy to other chains other than ETH? If not, then it's alright to hardcode here

foundry/src/executors/RocketpoolExecutor.sol
```diff
@@ -0,0 +1,86 @@
...
+        returns (uint256 calculatedAmount)
+    {
+        (bool isDeposit, TransferType transferType, address receiver) =
+            _decodeData(data);
+
+        if (isDeposit) {
+            // ETH -> rETH: Deposit ETH to Rocketpool to receive rETH
+            // We don't need to _transfer ETH into this contract since it must be sent along with the call
+            uint256 rethBefore = RETH.balanceOf(address(this));
+            ROCKET_DEPOSIT_POOL.deposit{value: givenAmount}();
```
tamaralipows, line 41: Interesting - I see fluid has the same restriction, that we must send it along with the function call instead of using our own `_transfer` method. I think you did this correctly!

tamaralipows (commented):
Hi @ElusAegis, thanks for your work, this looks great! I couldn't find any issues. Regarding your questions:

> I currently test the full integration path only for the deposit case, leaving the withdraw untested. This is because I gathered we should only have a single calldata test case in calldata.txt. Is this a right assumption? Or should I add several calldatas to test to for a single protocol to calldata.txt.

Please add the withdraw method test if you find it important! You can add as many cases as you want in `calldata.txt` :) All you need to do is give them different names.

> I am still not sure how the native ETH exchange works. Notice that if the transfer type is TransferFrom, we assume that the router/executor should withdraw the target token from the usre balance for the exchange. However, for ETH, this is impossible, and it should be transfered with the function call, breaking the assumption of the TransferFrom. Thus it feels like TransferFrom should not be supported for the deposit. However, I have also noticed that FluidV1 also works with native ETH and does none of such checks, so I am a bit unclear about if this is needed or not.

That's a good question. Let me explain to you:
On the encoding side, if you have a look at `TransferOptimization.get_transfers`, you will see this case:

```
        if swap.token_in == self.native_token {
            // Funds are already in router. All protocols currently take care of native transfers.
            TransferType::None
 ```
 
 Note that the `TransferType` will always be `None` if the token in is the native token. The user is then responsible for sending the correct value in their transaction when calling the `TychoRouter`. Then, the pool usually takes it from our contract, or, in the case of **fluid** and **rocketpool**, we need to call a specific method when sending the value so we can't use our `_transfer` method anyway (so it's fine that the type is set to `None`).
 
 Please let me know if that clears it up for you.
 
 Let me know if you add the second test, and this should be good to deploy!

File: foundry/src/executors/RocketpoolExecutor.sol (hunk 1 above)
Comment (ElusAegis) [reply foundry/src/executors/RocketpoolExecutor.sol:24]: No, Rocketpool is only on Ethereum, with no plans to move to other chains

File: foundry/src/executors/RocketpoolExecutor.sol (hunk 2 above)
Comment (ElusAegis) [reply foundry/src/executors/RocketpoolExecutor.sol:41]: Perfect!

File: src/encoding/evm/constants.rs [hunk 3]
```diff
@@ -43,6 +43,7 @@ pub static IN_TRANSFER_REQUIRED_PROTOCOLS: LazyLock<HashSet<&'static str>> = Laz
     set.insert("vm:balancer_v3");
     set.insert("fluid_v1");
     set.insert("aerodrome_slipstreams");
+    set.insert("rocketpool");
```
dianacarvalho1: I think it shouldn't be here. For both cases (minting and burning) we assume that the funds need to be in the tycho router. The funds don't need to be sent to the "pool". This set is for protocols where we need to send funds to the pool. In the rETH case it assumes that msg.sender has the funds and that's it

tests/protocol_integration_tests.rs
```diff
@@ -1028,6 +1028,52 @@ fn test_sequential_encoding_strategy_fluid() {
     write_calldata_to_file("test_sequential_encoding_strategy_fluid_v1", hex_calldata.as_str());
 }
 
+#[test]
+fn test_sequential_encoding_strategy_rocketpool() {
```
dianacarvalho1: it's single no?

src/encoding/evm/constants.rs
dianacarvalho1: If you make a sequential swap like DAI -(univ4)-> ETH -(rocket)-> rETH I believe it should fail if the protocol is in this set

dianacarvalho1 (commented):
Thank you for this @ElusAegis ! 🙏🏼 This looks very good, I only have one concern!

src/encoding/evm/constants.rs
ElusAegis: Thank you, you are absolutely right, I misunderstood the purpose of this list. For rocketpool, we indeed expect that we do all the token work from the Executor , not doing any direct transfer to the rocketpool per say

dianacarvalho1 (approved):
Thank you @ElusAegis ! 🙏🏼  It looks good to me now!

PR #305, opened 2025-12-10
feat: Group USV4 hooks + non-hooks

tamaralipows:


src/encoding/evm/constants.rs
```diff
@@ -71,3 +71,15 @@ pub static NON_PLE_ENCODED_PROTOCOLS: LazyLock<HashSet<&'static str>> = LazyLock
...
+/// Protocols in the same family can be grouped together for gas optimization.
+///
+/// For example, `uniswap_v4` and `uniswap_v4_hooks` are part of the same family
+/// since they use the same PoolManager and flash accounting system.
+pub fn get_protocol_grouping_family(protocol_system: &str) -> &str {
+    match protocol_system {
+        "uniswap_v4" | "uniswap_v4_hooks" => "uniswap_v4",
+        _ => protocol_system,
+    }
+}
```
tamaralipows: This is such a specific case I also wouldn't mind just hardcoding it into the `group_swaps.rs` file... to not overwhelm the user with too many options in `constants.rs` ... let me know what you thnk

foundry/test/protocols/UniswapV4.t.sol
```diff
@@ -624,3 +624,35 @@ contract TychoRouterForBalancerV3Test is TychoRouterTestSetup {
...
 }
+
+contract TychoRouterForUniswapV4AndEulerTest is TychoRouterTestSetup {
+    function getForkBlock() public view virtual override returns (uint256) {
+        return 22689128;
+    }
+
+    function testSingleUSV4AndHooksIntegrationGroupedSwap() public {
+        // Test created with calldata from our router encoder.
+        // Tests that uniswap_v4 and uniswap_v4_hooks can be grouped together
```
dianacarvalho1, line 635: did you inspect the traceback to see that they were indeed grouped?

src/encoding/evm/constants.rs
dianacarvalho1: aaff yeah I don't mind hardcoding this either. the less concepts we introduce because of this the better

dianacarvalho1 (commented):
Thank you!

foundry/test/protocols/UniswapV4.t.sol
tamaralipows: I've checked the calldata which shows that they were indeed grouped.
I haven't checked the internal working of the token transfers - do you think this is necessary to dive into?

dianacarvalho1: no, if the calldata is grouped then the transfers should be too

dianacarvalho1 (approved):
perfection ✨

PR #307, opened 2025-12-16
feat: -

tamaralipows:


foundry/src/TychoFees.sol
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        amountAfterFee = amount - fee;
+
+        // Credit fee to msg.sender's vault
+        _creditUserVault(msg.sender, token, fee);
+
+        // Transfer remainder to receiver
+        _transferToken(token, receiver, amountAfterFee);
+    }
```
tamaralipows: This draft does not yet include the Tycho fee (that will be taken as a percentage of the solver fee or a percentage of the output amount)

foundry/src/TychoRouter.sol
```diff
@@ -809,33 +840,13 @@ contract TychoRouter is
...
         return token == address(0)
             ? owner.balance
             : IERC20(token).balanceOf(owner);
     }
 
-    /**
-     * @dev Verifies that the expected amount of output tokens was received by the receiver.
-     * It also handles the case of arbitrage swaps where the input and output tokens are the same.
-     */
-    function _verifyAmountOutWasReceived(
```
dianacarvalho1, line 823: I think we should keep this but accounting the fee as well

foundry/src/TychoVault.sol
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        if (token == address(0)) {
+            // Native ETH deposit
+            require(msg.value == amount, "Value mismatch");
+            _vaultBalances[msg.sender][token] += amount;
+        } else {
+            // ERC20 deposit - transfer to this contract (router)
+            _vaultBalances[msg.sender][token] += amount;
+            IERC20(token).safeTransferFrom(msg.sender, address(this), amount);
```
dianacarvalho1: this means the msg.sender needs to approve the router before depositing, ok ok

foundry/src/TychoVault.sol
```diff
@@ -0,0 +1,212 @@
...
+     */
+    function _creditUserVault(address user, address token, uint256 amount)
+        internal
+        virtual
+    {
+        if (amount == 0) return;
+        _vaultBalances[user][token] += amount;
+        emit Transfer(
+            address(this), address(0), user, uint256(uint160(token)), amount
+        );
```
dianacarvalho1: emitting an event is expensive 😕 not sure if we would like to do this here

foundry/src/TychoVault.sol
```diff
@@ -0,0 +1,212 @@
...
+        internal
+    {
+        uint256 balance = _vaultBalances[user][token];
+        if (balance < amount) {
+            revert TychoVault__InsufficientBalance(user, token, amount, balance);
+        }
+        _vaultBalances[user][token] = balance - amount;
+        emit Transfer(
+            address(this), user, address(0), uint256(uint160(token)), amount
+        );
```
dianacarvalho1: same here

foundry/src/TychoVault.sol
dianacarvalho1: I don't even understand why we would emit a Transfer event tbh

foundry/src/RestrictTransferFrom.sol
```diff
@@ -128,6 +142,20 @@ contract RestrictTransferFrom {
             } else {
                 IERC20(tokenIn).safeTransfer(receiver, amount);
             }
+        } else if (transferType == TransferType.TransferFromVault) {
+            address sender;
```
dianacarvalho1: can we do the amount allowed and token in storage checks as well here please?

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 small small comments only (I didn't look at the fee taking and encoding too closely yet!)

foundry/src/TychoVault.sol
tamaralipows: woops not left here on purpose lol.

foundry/src/TychoRouter.sol
tamaralipows: In that version that you are commenting on - I had moved it to the Fees contract. Moving it back now!

foundry/src/executors/FeeExecutor.sol
```diff
@@ -2,41 +2,74 @@
...
-     * @return amountOut The output amount (after fee deduction)
+     * @dev Deducts fees from the input amount and credits them to fee receivers' vaults
+     * @dev Verifies msg.sender has sufficient balance in vault before deducting
+     * @param amountIn The input amount (before fees)
+     * @param data Encoded fee parameters:
+     *        solutionFeeBps (uint16) | solutionFeeReceiver (address) |
+     *        routerFeeBps (uint16) | routerFeeReceiver (address) | token (address)
+     * @return amountOut The output amount (after fee deductions)
      */
     function swap(uint256 amountIn, bytes calldata data)
```
tamaralipows: Let's call this method something different.

PR #310, opened 2025-12-22
feat: track wei remainder from stETH share-to-ETH cast

Troshchk:


foundry/src/executors/CurveExecutor.sol
```diff
@@ -9,6 +9,9 @@ import {RestrictTransferFrom} from "../RestrictTransferFrom.sol";
 error CurveExecutor__AddressZero();
 error CurveExecutor__InvalidDataLength();
 
+address constant STETH_ADDR =
+    address(0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84);
```
dianacarvalho1: Can we have this as a constructor argument please? It might differ for different chains

foundry/src/executors/CurveExecutor.sol
```diff
@@ -49,6 +52,19 @@ contract CurveExecutor is IExecutor, RestrictTransferFrom {
...
 
+    function _getBalanceWithStEthHandling(address tokenOut)
+        internal
+        returns (uint256)
+    {
+        uint256 balance;
+        if (tokenOut == STETH_ADDR) {
+            balance = IERC20(STETH_ADDR).balanceOf(address(this));
+        } else {
+            balance = _balanceOf(tokenOut);
```
dianacarvalho1: how is this different? the `_balanceOf` literally does `IERC20(token).balanceOf(address(this))` as well 😕

foundry/src/executors/CurveExecutor.sol
```diff
@@ -104,17 +121,26 @@ contract CurveExecutor is IExecutor, RestrictTransferFrom {
             }
         }
 
-        uint256 balanceAfter = _balanceOf(tokenOut);
+        uint256 balanceAfter = _getBalanceWithStEthHandling(tokenOut);
```
dianacarvalho1: the balance checks are unnecessary.. they are already done in the original contract

foundry/src/executors/CurveExecutor.sol
```diff
@@ -104,17 +121,26 @@ contract CurveExecutor is IExecutor, RestrictTransferFrom {
...
+
         if (receiver != address(this)) {
             if (tokenOut == nativeToken) {
                 Address.sendValue(payable(receiver), amountOut);
             } else {
                 IERC20(tokenOut).safeTransfer(receiver, amountOut);
             }
+            if (tokenOut == STETH_ADDR) {
+                castRemainderWei = IERC20(STETH_ADDR).balanceOf(address(this));
+            }
```
dianacarvalho1: I think only this part is relevant actually

dianacarvalho1 (commented):
Thank you @Troshchk ! I have some comments..

foundry/src/executors/CurveExecutor.sol
Troshchk: You are right, removed it

foundry/src/executors/CurveExecutor.sol
Troshchk: Yes, done

foundry/src/executors/CurveExecutor.sol
Troshchk: Reverted to the original method

foundry/src/executors/CurveExecutor.sol
```diff
@@ -105,16 +116,26 @@ contract CurveExecutor is IExecutor, RestrictTransferFrom {
...
                 IERC20(tokenOut).safeTransfer(receiver, amountOut);
             }
+            // stETH is not deployed on Unichain, so stEthAddress may be address(0)
+            if (hasStETH && tokenOut == stEthAddress) {
+                castRemainderWei = IERC20(stEthAddress).balanceOf(address(this));
+            }
         }
-        return amountOut;
+
+        uint256 amountOutFinal = amountOut - castRemainderWei;
```
dianacarvalho1: you can do `amountOut =- castRemainderWei` inside of the if statement above, no need to do it here and having to define castRemainderWei outside

dianacarvalho1: what if this is the second stETH transfer and `address(this)` already has moneys accumulated there from before? we would be wrongly updating the `amountOut`

dianacarvalho1 (commented):
Thank you @Troshchk ! I believe this not correct yet though 😕

foundry/src/executors/CurveExecutor.sol
Troshchk: Good point! I’ve updated the logic to account for any existing stETH balance already held by the contract and added a test that covers this scenario. Does this make sense?

foundry/src/executors/CurveExecutor.sol
```diff
@@ -39,14 +39,24 @@ contract CurveExecutor is IExecutor, RestrictTransferFrom {
...
-    constructor(address _nativeToken, address _permit2)
+    constructor(address _nativeToken, address _permit2, address _stEthAddress)
         RestrictTransferFrom(_permit2)
     {
         if (_nativeToken == address(0)) {
             revert CurveExecutor__AddressZero();
         }
         nativeToken = _nativeToken;
+
+        // Optional: stETH is not deployed on Unichain, so zero address is allowed
```
tamaralipows: For my context - why would this not be allowed on Unichain? Things can't be address(0) on Unichain?

foundry/test/protocols/Curve.t.sol
```diff
@@ -134,11 +135,42 @@ contract CurveExecutorTest is Test, TestUtils, Constants {
...
 
-        assertEq(amountOut, 1001072414418410897);
-        assertEq(
-            IERC20(STETH_ADDR).balanceOf(ALICE),
-            amountOut - 1 // there is something weird in this pool, but won't investigate for now because we don't currently support it in the simulation
+        assertEq(amountOut, 1001072414418410896);
+        assertEq(IERC20(STETH_ADDR).balanceOf(ALICE), amountOut);
+    }
+
+    function testStEthPoolWithInitialstETH() public {
```
tamaralipows, line 144: What exactly is being tested here that wasn't in the previous tests? That the initial stETH is there in the router from the previous swap? Could we add a small comment at the top?

foundry/src/executors/CurveExecutor.sol
```diff
@@ -105,16 +116,27 @@ contract CurveExecutor is IExecutor, RestrictTransferFrom {
...
         uint256 amountOut = balanceAfter - balanceBefore;
 
+        uint256 castRemainderWei = 0;
+
         if (receiver != address(this)) {
             if (tokenOut == nativeToken) {
                 Address.sendValue(payable(receiver), amountOut);
             } else {
                 IERC20(tokenOut).safeTransfer(receiver, amountOut);
             }
```
tamaralipows, line 129: Can we add a comment here that we might possibly lose 1wei on the transfer due to rounding errors

foundry/src/executors/CurveExecutor.sol
tamaralipows: I agree with diana's first comment that we should do `amountOut -= castRemainderWei` instead of adding an extra line here

tamaralipows (commented):
Thank you! Just minor readability comments! Logic looks good

foundry/src/executors/CurveExecutor.sol
Troshchk: Changed it

foundry/test/protocols/Curve.t.sol
Troshchk: Yes, exactly. I have added this test to also cover the case Diana mentioned here:
https://github.com/propeller-heads/tycho-execution/pull/310#discussion_r2686565443
Also added a comment

foundry/src/executors/CurveExecutor.sol
Troshchk: Good catch, in that case the comment is incorrect, I will remove it.

stETH is not deployed on Unichain, so there is no need to update amountOut using castRemainderWei on that chain.
The constructor still expects an stETH address. If address(0) is passed, it means stETH doesn’t exist on that chain and we set the hasStETH flag to false. This is needed because the stETH address can be different on each chain.
Later on, the hasStETH flag is used to decide whether amountOut should be adjusted with castRemainderWei.

dianacarvalho1 (commented):
Thank @Troshchk ! Logic wise this looks good now! _However_ do we know if this is still a problem with Lido v3?? 🫠  can we try to make a new test in a block after the update to confirm this?

foundry/test/protocols/LidoV3.t.sol
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: BUSL-1.1
```
dianacarvalho1: What was the motivation to make an entire new file with repeated tests? 🤔 I thought you meant to add something like `TychoRouterForLidoTest` at the end of the current Lido test file. Making this entire file might be extremely misleading, like we are already implementing v3 or something

Troshchk: I put it in a separate file because I thought it’d be clearer to keep the one file per protocol tests consistent with how the other protocols are handled. But I see how that could be misleading, so I’ll move it into the same file as the other Lido tests.

foundry/test/protocols/Lido.t.sol
```diff
@@ -240,3 +240,214 @@ contract TychoRouterForLidoTest is TychoRouterTestSetup {
...
+    LidoExecutorExposed LidoExposed;
+
+    function setUp() public {
+        uint256 forkBlock = 24238735;
+        vm.createSelectFork(vm.rpcUrl("mainnet"), forkBlock);
+        LidoExposed =
+            new LidoExecutorExposed(STETH_ADDR, WSTETH_ADDR, PERMIT2_ADDRESS);
+    }
+
+    function testDecodeParams() public view {
```
dianacarvalho1: I still don't understand why we need all the tests though.. Can you explain please? My expectation is that only one test would be relevant.
Can you also please add a comment explaining why the test is necessary and that lidov3 is not actually implemented yet?

Troshchk: We discussed it in a call and I have made the changes

dianacarvalho1 (commented):
ok perfect now ✨ Thank you! Now let's deploy on monday (there is one addition we need to do here still and then we can merge this!)

dianacarvalho1 (approved):
Thank you 🙏🏼

PR #315, opened 2026-01-06
feat: Do the transfers in the Dispatcher and not in the Executors

dianacarvalho1:
Implemented it only for uniswap v2, v3 and ekubo and ran the Router tests ✅

### DIFF
--- a/foundry/interfaces/ICallback.sol
+++ b/foundry/interfaces/ICallback.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.26;
 
+import "../src/RestrictTransferFrom.sol";
+
 interface ICallback {
     /**
      * @notice Handles callback data from a protocol or contract interaction.
@@ -10,9 +12,9 @@ interface ICallback {
      * @param data The encoded callback data to be processed.
      * @return result The encoded result of the callback processing.
      */
-    function handleCallback(
-        bytes calldata data
-    ) external returns (bytes memory result);
+    function handleCallback(bytes calldata data)
+        external
+        returns (bytes memory result);
 
     /**
      * @notice Verifies the validity of callback data.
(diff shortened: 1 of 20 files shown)


foundry/interfaces/ICallback.sol
```diff
@@ -22,4 +24,15 @@ interface ICallback {
...
+    function getCallbackTransferData(bytes calldata data)
+    external
+    payable
+    returns (
+        RestrictTransferFrom.TransferType transferType,
+        address receiver,
+        address tokenIn,
+        bool approvalNeeded,
+        uint256 amount
+    );
```
dianacarvalho1, line 37: this could be a special type if we want

foundry/interfaces/IExecutor.sol
```diff
@@ -20,10 +21,21 @@ interface IExecutor {
...
+    function getTransferData(bytes calldata data)
+        external
+        payable
+        returns (
+            RestrictTransferFrom.TransferType transferType,
+            address receiver,
+            address tokenIn,
+            bool approvalNeeded,
+            address tokenOut
+        );
```
dianacarvalho1, line 38: same here. and I was thinking of instead of returning the tokenOut here we could change the swap method signature to return the calculatedAmount and the tokenOut. 
The tokenOut is not needed for v2 but it will be used for v3

tamaralipows:
We still need to perform the crediting after the executor swap - so the executors will need to return (`output_token`, `receiver`) so that we can update the delta accounting (credit the user) if the `receiver == address(this)` in the dispatcher.

foundry/src/Dispatcher.sol
```diff
@@ -69,6 +76,32 @@ contract Dispatcher {
...
+                        : abi.encodePacked("Getting transfer data failed")
+                )
+            );
+        }
+
+        (
+            RestrictTransferFrom.TransferType transferType,
+            address receiver,
+            address tokenIn,
+            bool approvalNeeded,
```
tamaralipows, line 97: This bool is no longer needed here

PR #317, opened 2026-01-07
feat: Get rid of SwapBuider and improve Swap

dianacarvalho1:
Swap now has private attributes with a default new() method and setters and getters for the attributes Updated entire codebase


tamaralipows (approved):
Thank you!

PR #319, opened 2026-01-08
feat: Add ability to register external encoders to SwapEncoderRegistry

dianacarvalho1:
This means changing the interface of the EncoderBuilders to take the SwapEncoderRegistry directly instead of building it on the fly


src/encoding/evm/swap_encoder/swap_encoder_registry.rs
```diff
@@ -6,63 +6,154 @@ use crate::encoding::{
...
+        Self { chain, encoders: HashMap::new() }
+    }
+
     /// Populates the registry with the `SwapEncoders` for the given blockchain by parsing the
     /// executors' addresses in the file at the given path.
-    pub fn new(executors_addresses: Option<String>, chain: Chain) -> Result<Self, EncodingError> {
+    pub fn with_default_encoders(
+        mut self,
+        executors_addresses: Option<String>,
+    ) -> Result<Self, EncodingError> {
```
tamaralipows: Perhaps we should call this `add_default_encoders` to emphasize that the user can call this **and also** register their own encoders? With this naming I feel like it's one or the other

tamaralipows (commented):
This looks good, I only have one suggestion - what do you think?

src/encoding/evm/swap_encoder/swap_encoder_registry.rs
dianacarvalho1: addressed it!

tamaralipows (approved):
Thank you!

PR #320, opened 2026-01-08
feat: Use CREATE2 in deployment scripts

tamaralipows:
CREATE doesn’t include the code hash in the address computation, so now CurveExecutor on Unichain has the same address as UniswapV3Executor on Base which can be confusing. 

Furthermore, our scripts were broken since the `to` address was not properly found after deployment (failing in the dungeon of hardhat - out of our control), making it look like our transaction failed when it was really passing. Now, the scripts are running fine without this same problem.

We now use an existing well-known create2 factory contract mentioned in the foundry docs. The source code for that factory contract is here: https://github.com/Arachnid/deterministic-deployment-proxy

Why this factory is necessary: When you send a transaction with bytecode but no to address, the EVM automatically uses CREATE, whereas CREATE2 requires an explicit opcode call with a salt parameter that can only be executed from within contract code as opposed to by using Javascript/Hardhat.

I've tested that the script works fine by deploying a UniswapV2Executor [here](https://etherscan.io/address/0x2B9faB7B44f42B90ED5BA963A0Da2C20bB4311B7#code). It successfully went through the whole deployment/verification pipeline without manual intervention - very smooth.

### DIFF
--- a/foundry/scripts/deploy-executors.js
+++ b/foundry/scripts/deploy-executors.js
@@ -220,18 +220,41 @@ async function main() {
     console.log(`Deploying with account: ${deployer.address}`);
     console.log(`Account balance: ${ethers.utils.formatEther(await deployer.getBalance())} ETH`);
 
+    // Deterministic Deployment Proxy
+    // More info: https://getfoundry.sh/guides/deterministic-deployments-using-create2/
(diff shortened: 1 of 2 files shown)


dianacarvalho1 (approved):
ai amazing 🥹 thank you 🙏🏼

PR #321, opened 2026-01-09
feat: Add a script to export run time bytecode

dianacarvalho1:
This was previously done with a test (hacky)

I compared the results between both versions and all good!


tamaralipows (approved):
Yay thank you!

PR #322, opened 2026-01-13
feat: Etherfi executor

zach030:


src/encoding/evm/swap_encoder/etherfi.rs
```diff
@@ -0,0 +1,275 @@
...
+    evm::{
+        approvals::protocol_approvals_manager::ProtocolApprovalsManager, utils::bytes_to_address,
+    },
+    models::{EncodingContext, Swap},
+    swap_encoder::SwapEncoder,
+};
+
+const EETH_ADDRESS: &str = "0x35fA164735182de50811E8e2E824cFb9B6118ac2";
+const WEETH_ADDRESS: &str = "0xCd5fE23C85820F7B72D0926FC9b05b43E359b7ee";
+const REDEMPTION_MANAGER_ADDRESS: &str = "0xDadEf1fFBFeaAB4f68A9fD181395F68b4e4E7Ae0";
```
dianacarvalho1: is this the same for all chains?? We have this file `config/protocol_specific_addresses.json` to save addresses like this!

foundry/src/executors/EtherfiExecutor.sol
```diff
@@ -0,0 +1,138 @@
...
+    address public constant ETH_ADDRESS =
+        0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+    address public constant EETH_ADDRESS =
+        0x35fA164735182de50811E8e2E824cFb9B6118ac2;
+    address public constant LIQUIDITY_POOL_ADDRESS =
+        0x308861A430be4cce5502d0A12724771Fc6DaF216;
+    address public constant WEETH_ADDRESS =
+        0xCd5fE23C85820F7B72D0926FC9b05b43E359b7ee;
+    address public constant REDEMPTION_MANAGER_ADDRESS =
+        0xDadEf1fFBFeaAB4f68A9fD181395F68b4e4E7Ae0;
```
dianacarvalho1: again.. are these the same for all chains? maybe they should be constructor args instead?

foundry/test/protocols/Etherfi.t.sol
```diff
@@ -0,0 +1,196 @@
...
+            EtherfiDirection.EethToWeeth,
+            true
+        );
+
+        uint256 balanceBefore = IERC20(WEETH_ADDR).balanceOf(BOB);
+        uint256 amountOut = etherfiExposed.swap(minted, protocolData);
+        uint256 balanceAfter = IERC20(WEETH_ADDR).balanceOf(BOB);
+
+        assertGt(balanceAfter, balanceBefore);
+        assertApproxEqAbs(balanceAfter - balanceBefore, amountOut, 1);
```
dianacarvalho1, line 138: the balance is not exactly the same?? This might cause problems in the router.. Is there something we can do to address this in the executor?

dianacarvalho1 (commented):
Thank you @zach030 ! I didn't review the Etherfi logic (I'm not familiar with the protocol) but I left some comments on how this is integrated in tycho-execution

src/encoding/evm/swap_encoder/etherfi.rs
zach030: Done

foundry/src/executors/EtherfiExecutor.sol
zach030: Done

foundry/test/protocols/Etherfi.t.sol
zach030: I made a fix here. eETH is a share token, so the transfer amount can slightly differ from the balance due to rounding. I’ve aligned the swap return value to use the balance instead of shares.

dianacarvalho1 (approved):
Nice! Thank you @zach030 ! 🙏🏼 this looks good to me!

PR #324, opened 2026-01-22
fix: Admin wallet used in solidity tests

dianacarvalho1:
The previous wallet had some txs recently which means that the addresses deployed with it would vary depending on the block. We don't really want this.
[issue](https://propeller-heads.atlassian.net/browse/ENG-5317)


tamaralipows (approved):
Thank you!

foundry/test/protocols/Bebop.t.sol
```diff
@@ -409,14 +409,14 @@ contract BebopExecutorTest is Constants, Permit2TestHelper, TestUtils {
 contract TychoRouterForBebopTest is TychoRouterTestSetup {
     // Override the fork block for Bebop tests
     function getForkBlock() public pure override returns (uint256) {
-        return 22667986;
+        return 24290852;
```
Troshchk, line 412: Why the blocks chosen for `Bebop` and `Hashflow ` are different?

Troshchk (commented):
Thank you @dianacarvalho1 !
It looks good to me, just one question about the block numbers!

foundry/test/protocols/Bebop.t.sol
dianacarvalho1: it depends on the quote's validities. I did the quotes independently at slightly different times

PR #326, opened 2026-01-27
fix: Remove option for uniswap_v4_hooks encoder entirely

dianacarvalho1:
This was introducing a bug in Angstrom. Right now we are merging all the uniswap_v4 or uniswap_v4_hooks swaps (so we can take advantage of univ4 flash accounting for swaps with or without hooks). However, the config necessary for the Angstrom hook was only defined for uniswap_v4_hooks which led to the Angstrom extra steps being entirely ignored


tamaralipows (approved):
Thank you! Please also copy these changes to v3

PR #329, opened 2026-02-02
feat: add timelock for new executors

Troshchk:
In this PR:
- a timelock for newly deployed contracts introduced
- _setExecutors updated with an additional parameter to bypass the timelock in test scenarios
- gas measured using `forge test --match-test testSetValidExecutor --gas-report`

Gas change
exposedSetExecutor: 47,661 → 47,890 gas (+229)
Executor read: 2461 → 2526 gas (+65)

### DIFF
--- a/foundry/src/Dispatcher.sol
+++ b/foundry/src/Dispatcher.sol
@@ -22,7 +22,7 @@ error Dispatcher__InvalidDataLength();
  */
 contract Dispatcher {
     struct ExecutorData {
-        uint256 activationBlock;
+        uint64 activationBlock;
         bool approved;
     }
 
@@ -32,8 +32,6 @@ contract Dispatcher {
     uint256 private constant _CURRENTLY_SWAPPING_EXECUTOR_SLOT =
         0x098a7a3b47801589e8cdf9ec791b93ad44273246946c32ef1fc4dbe45390c80e;
 
-    uint256 private constant _BLOCKS_TO_DELAY_FOR_NEW_EXECUTOR = 50400; // ~1 week
-
     event ExecutorSet(address indexed executor);
     event ExecutorRemoved(address indexed executor);
 
@@ -42,15 +40,15 @@ contract Dispatcher {
      *  contract.
      * @param target address of the executor contract
      */
-    function _setExecutor(address target) internal {
+    function _setExecutor(address target, uint64 block_to_delay_activation)
+        internal
+    {
         if (target.code.length == 0) {
             revert Dispatcher__NonContractExecutor();
         }
 
         executorData[target] = ExecutorData({
-            activationBlock: uint64(
-                block.number + _BLOCKS_TO_DELAY_FOR_NEW_EXECUTOR
-            ),
+            activationBlock: uint64(block.number + block_to_delay_activation),
             approved: true
         });
         emit ExecutorSet(target);
--- a/foundry/src/Dispatcher.sol
+++ b/foundry/src/Dispatcher.sol
@@ -73,13 +73,13 @@ contract Dispatcher {
         uint256 amount,
         bytes calldata data
     ) internal returns (uint256 calculatedAmount) {
-        ExecutorData memory d = executorData[executor];
+        ExecutorData memory executorInfo = executorData[executor];
 
-        if (!d.approved) {
+        if (!executorInfo.approved) {
             revert Dispatcher__UnapprovedExecutor(executor);
         }
 
-        if (block.number < d.activationBlock) {
+        if (block.number < executorInfo.activationBlock) {
             revert Dispatcher__ExecutorIsTimelocked(executor);
         }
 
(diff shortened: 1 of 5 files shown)


foundry/src/Dispatcher.sol
```diff
@@ -102,7 +120,9 @@ contract Dispatcher {
             executor := tload(_CURRENTLY_SWAPPING_EXECUTOR_SLOT)
         }
 
-        if (!executors[executor]) {
+        ExecutorData memory executorInfo = executorData[executor];
```
dianacarvalho1, line 123: why don't we need to check for the block here?

foundry/src/TychoRouter.sol
```diff
@@ -89,6 +89,15 @@ contract TychoRouter is
...
 
+    function _blocksToDelayForNewExecutor()
+        internal
+        view
+        virtual
+        returns (uint64)
+    {
+        return 50400; // ~1 week
+    }
+
```
dianacarvalho1, line 100: if we want to define it like this, this could be just a constant (no need to be a method). However I'm wondering if there couldn't be value with us being able to modify this value. 🤔 what do you think? Also I think 1 week is too much. maybe just 2 or 3 days?

foundry/src/Dispatcher.sol
```diff
@@ -20,7 +21,12 @@ error Dispatcher__InvalidDataLength();
  *  an alternate selector is specified.
  */
 contract Dispatcher {
-    mapping(address => bool) public executors;
+    struct ExecutorData {
+        uint64 activationBlock;
+        bool approved;
```
dianacarvalho1, line 26: I'm wondering if we need the `approved` to begin with.. 🤔 we could just check if activationBlock>0 for that no?

foundry/test/Dispatcher.t.sol
```diff
@@ -14,7 +14,11 @@ contract DispatcherExposed is Dispatcher {
     }
 
     function exposedSetExecutor(address target) external {
-        _setExecutor(target);
+        _setExecutor(target, 50400);
+    }
+
+    function exposedSetExecutorImmediately(address target) external {
+        _setExecutor(target, 0);
```
dianacarvalho1, line 21: I don't like this much 😕

foundry/test/Dispatcher.t.sol
```diff
@@ -65,6 +81,34 @@ contract DispatcherTest is Constants {
         dispatcherExposed.exposedSetExecutor(BOB);
     }
 
+    function testCallTimelockedExecutor() public {
```
dianacarvalho1, line 84: Can you test this functionality with a more expressive (not so obscure) test please?
Using a real tx makes it confusing imo. We could create a very simple setup to do this exact same thing. Also using a real tx assumes that the TychoRouter code/interfaces are constant

foundry/test/Dispatcher.t.sol
dianacarvalho1: We could just roll the fork number to bypass the waiting time. I mean:
- Force the block to be X when setting the executor
- Then change the block to be X + waiting time when actually running the test

foundry/test/TychoRouterTestSetup.sol
```diff
@@ -32,6 +32,15 @@ import "@src/TychoRouter.sol";
...
 
 contract TychoRouterExposed is TychoRouter {
+    function _blocksToDelayForNewExecutor()
+        internal
+        view
+        override
+        returns (uint64)
+    {
+        return 0;
+    }
```
dianacarvalho1, line 42: also here, we could play with the block numbers instead of forcing this

dianacarvalho1: It is never a good sign when we need to change the production code to accommodate testing 😕

dianacarvalho1 (commented):
Thank you @Troshchk ! 🙏🏼 Overall this looks good! I have a few comments

foundry/src/TychoRouter.sol
Troshchk: Sounds good -- I’ll move it to 2–3 days 👍
I added a `_blocksToDelayForNewExecutor` method instead of a constant here so it can be redefined for tests in `TychoRouterExposed`
I like your suggestion of forcing the block to be different for tests, so I’ll try that approach for testing and use a constant instead of `_blocksToDelayForNewExecutor`. I think it is safer to have a a hardcoded constant here rather than a method

foundry/test/TychoRouterTestSetup.sol
Troshchk: Agree! I could not find a different method to enable testing, but I like your suggestion about forcing the block to be different in the tests, so I will try it

foundry/src/Dispatcher.sol
Troshchk: We do, updated it

PR #333, opened 2026-02-09
feat(sdk notify): Wait for cargo release, not just github publish

tamaralipows:
- We were sending the event before it was published to crates, which was problematic because tycho-protocol-sdk then attempted to pull from crates.
- About getting the latest tag: when you trigger on workflow_run, you DON'T have access to the release event context anymore. The workflow_run event only tells you "the Release workflow completed" - it doesn't include the release details. For this reason, we need to add another API call here to get the latest tag.

### DIFF
--- a/.github/workflows/notify-sdk.yaml
+++ b/.github/workflows/notify-sdk.yaml
@@ -13,7 +13,7 @@ jobs:
     if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'release' }}
     steps:
       - name: Wait for crates.io propagation
-        run: sleep 30
+        run: sleep 60
 
       - name: Get latest release tag
         id: get_tag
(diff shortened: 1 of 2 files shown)


.github/workflows/notify-sdk.yaml
```diff
@@ -1,22 +1,37 @@
...
+      - completed
 
 jobs:
   notify-sdk:
     name: Trigger SDK update workflow
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'release' }}
     steps:
+      - name: Wait for crates.io propagation
+        run: sleep 30
```
dianacarvalho1: is 30 seconds enough?

dianacarvalho1 (approved):
Thank you 🙏🏼

.github/workflows/notify-sdk.yaml
tamaralipows: doubling in case

tamaralipows:
Marking this PR as feat to trigger CI for testing

PR #340, opened 2026-02-26
fix: rocketpool v1.4 update

ElusAegis:


dianacarvalho1 (commented):
It looks good, thank you @ElusAegis ! 
I won't approve yet because we miss the redeployment and assigning the new address in the config

dianacarvalho1 (approved):

